PR #54, opened 2020-03-30
build: populate version git vars when building using Makefile.

jrasell:
The Makefile build target was not populating the version git vars
which meant the version command was missing some information. This
change ensures the makefile build target populates to the ldflags
to reflect those used during the release process.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -1,6 +1,11 @@
 SHELL = bash
 default: lint test build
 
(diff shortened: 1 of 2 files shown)


cgbaker:
why does this remove the `build-docker` target?

jrasell:
build-docker target was obsolete so removed it as part of the makefile work.

PR #57, opened 2020-03-30
version: update version to reflect current 0.0.1 dev phase.

jrasell:


version/version.go
```diff
@@ -20,7 +20,7 @@ var (
 	// A pre-release marker for the version. If this is "" (empty string)
 	// then it means that it is a final release. Otherwise, this is a pre-release
 	// such as "dev" (in development), "beta", "rc1", etc.
-	VersionPrerelease = "techpreview1"
+	VersionPrerelease = "dev"
```
cgbaker: question: should we call it `0.0.1-dev`, if that's what we're targeting next?
i think `dev` might be a little generic.

jrasell: The version output will be `Nomad Autoscaler v0.0.1-dev (commit)` as `VersionPrerelease` is suffixed onto the `Version` variable when creating the full version string.

cgbaker: marvelous, i didn't read the whole thing. sorry!

PR #67, opened 2020-03-31
Allow `nil` value for Count

### LINKED ISSUE #41: strategy: allow `Count` to be `nil`
If a strategy returns a `nil` value for `Count` it means that the existing count value in the target should not change.

lgfa29:
If a Strategy passes a `nil` value in an Action, this indicates to the Autoscaler and Targets that the current count value shouldn't change.

This can be used to register events against the Target without executing any action.

Closes #41.

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -452,22 +452,15 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
 
 		// Check if this action will not violate the min and max limits set by
 		// the policy.
-		withinLimts, err := action.IsWithinLimits(p.Min, p.Max)
-		if err != nil {
-			logger.Error(fmt.Sprintf("failed to execute action: %s", err))
-			continue
-		}
-
+		// We can ignore the returned error becase we check for nil Count above.
+		withinLimts, _ := action.IsWithinLimits(p.Min, p.Max)
 		if !withinLimts {
 			logger.Info("next count outside limits",
 				"from", currentCount, "to", *action.Count, "min", p.Min, "max", p.Max)
 
 			// Make sure new count value is within [min, max] limits
-			err := action.CapCount(p.Min, p.Max)
-			if err != nil {
-				logger.Error(fmt.Sprintf("failed to execute action: %s", err))
-				continue
-			}
+			// We can ignore the returned error becase we check for nil Count above.
+			action.CapCount(p.Min, p.Max)
 
 			logger.Info("updated count to be within limits",
 				"from", currentCount, "to", *action.Count, "min", p.Min, "max", p.Max)
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -424,62 +424,35 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
 
 	// scale target
 	for _, action := range results.Actions {
+		actionLogger := logger.With("target_config", p.Target.Config)
+
 		// Make sure returned action has sane defaults instead of relying on
 		// plugins doing this.
 		action.Canonicalize()
+
+		// Make sure new count value is within [min, max] limits
+		action.CapCount(p.Min, p.Max)
 
 		// If the policy is configured with dry-run:true then we set the
 		// action count to nil so its no-nop. This allows us to still
 		// submit the job, but not alter its state.
 		if val, ok := p.Target.Config["dry-run"]; ok && val == "true" {
-			logger.Info("scaling dry-run is enabled, using no-op task group count",
-				"target_config", p.Target.Config)
-			action.Meta["nomad_autoscaler.dry_run"] = "true"
-			action.Meta["nomad_autoscaler.dry_run_count"] = action.Count
-			action.Count = nil
+			actionLogger.Info("scaling dry-run is enabled, using no-op task group count")
+			action.SetDryRun(true)
 		}
 
-		// A nil count indicates that the current value shouldn't change,
-		// so just notify the policy's target of this event.
 		if action.Count == nil {
-			logger.Info("no change in count")
-
-			if err = (*targetPlugin).Scale(action, p.Target.Config); err != nil {
-				logger.Error("failed to scale target", "error", err)
-			}
-			continue
+			actionLogger.Info("registering scaling event",
+				"count", currentCount, "reason", action.Reason, "meta", action.Meta)
+		} else {
+			actionLogger.Info("scaling target",
+				"from", currentCount, "to", *action.Count,
+				"reason", action.Reason, "meta", action.Meta)
 		}
-
-		// Check if this action will not violate the min and max limits set by
-		// the policy.
-		// We can ignore the returned error becase we check for nil Count above.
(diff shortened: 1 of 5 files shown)


agent/agent.go
```diff
@@ -427,34 +428,59 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
 		// plugins doing this.
 		action.Canonicalize()
 
-		if !action.IsWithinLimits(p.Min, p.Max) {
+		// If the policy is configured with dry-run:true then we set the
+		// action count to nil so its no-nop. This allows us to still
+		// submit the job, but not alter its state.
+		if val, ok := p.Target.Config["dry-run"]; ok && val == "true" {
```
cgbaker, line 434: is the .Meta API exposed in the autoscaler yet? this would be a good candidate for some additional metadata.

agent/agent.go
```diff
@@ -427,34 +428,59 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
...
+			logger.Info("scaling dry-run is enabled, using no-op task group count",
+				"target_config", p.Target.Config)
+			action.Count = nil
```
cgbaker, line 437: ^^^ like, for example, logging the count that we would have set if dry-run hadn't been enabled.

agent/agent.go
```diff
@@ -427,34 +428,59 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
...
+				logger.Error("failed to scale target", "error", err)
+			}
+			continue
+		}
+
+		// Check if this action will not violate the min and max limits set by
+		// the policy.
+		withinLimts, err := action.IsWithinLimits(p.Min, p.Max)
+		if err != nil {
+			logger.Error(fmt.Sprintf("failed to execute action: %s", err))
```
cgbaker, line 455:
```suggestion
			logger.Error(fmt.Sprintf("failed to check limits: %s", err))
```

cgbaker: is this error message intentional or a copy-paste consequence?

agent/agent.go
```diff
@@ -427,34 +428,59 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
...
+		if !withinLimts {
 			logger.Info("next count outside limits",
-				"from", currentCount, "to", action.Count, "min", p.Min, "max", p.Max)
+				"from", currentCount, "to", *action.Count, "min", p.Min, "max", p.Max)
 
 			// Make sure new count value is within [min, max] limits
-			action.CapCount(p.Min, p.Max)
+			err := action.CapCount(p.Min, p.Max)
+			if err != nil {
+				logger.Error(fmt.Sprintf("failed to execute action: %s", err))
```
cgbaker, line 466: ^^^ same for this error message

agent/agent.go
```diff
@@ -427,34 +428,59 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
...
+			}
 
 			logger.Info("updated count to be within limits",
-				"from", currentCount, "to", action.Count, "min", p.Min, "max", p.Max)
+				"from", currentCount, "to", *action.Count, "min", p.Min, "max", p.Max)
 		}
 
-		if action.Count == currentCount {
+		// If count is not nil, but its value doesn't change, assume that
+		// no action should happen, not even an empty scale event.
```
cgbaker, line 475: so, i think later on once there's support for .Meta, it may be worth logging an event here, so we could at least see the count that would have been set.

plugins/nomad/target/tgcount.go
```diff
@@ -43,8 +43,12 @@ func (t *NomadGroupCount) Count(config map[string]string) (int64, error) {
 }
 
 func (t *NomadGroupCount) Scale(action strategy.Action, config map[string]string) error {
-	countInt := int(action.Count)
-	_, _, err := t.client.Jobs().Scale(config["job_id"], config["group"], &countInt, &action.Reason, nil, nil, nil)
+	var countIntPtr *int
+	if action.Count != nil {
+		countInt := int(*action.Count)
```
cgbaker, line 48: ugh, i'm starting to regret using int64, because i'm tired of having to check the safety of these casts. maybe i should change the API in Jobs().Scale() so, at least, you don't have to worry about it.

strategy/strategy.go
```diff
@@ -24,18 +24,28 @@ func (a *Action) Canonicalize() {
 	}
 }
 
-func (a *Action) IsWithinLimits(min, max int64) bool {
-	return a.Count >= min && a.Count <= max
+func (a *Action) IsWithinLimits(min, max int64) (bool, error) {
```
cgbaker, line 27: it may be cleaner to just have `IsWithinLimits` and `CapCount` be no-ops when `count == nil`, since the only error case is nil. it will eliminate the error handling, at least.

lgfa29: I thought of this, but I couldn't find a good no-op for this method. Returning either `true` or `false` are both misleading and could cause problems.

This is not the case for the current only place that I call this, since I check for a `nil` value before calling `IsWithingLimits`, but I didn't want to tie the method's signature with how it will be used. 

What I can do is to just ignore the error in the agent since, as I mentioned, I will be checking for `nil` beforehand.

plugins/nomad/target/tgcount.go
lgfa29: Yeah, it's ugly but it makes sense I guess, since `Count` is an `int` in Nomad.

agent/agent.go
lgfa29: I think this would be the most common scenario, which is nothing is happening. Adding a call to the scaling API just to register that nothing happened would be too verbose I think.

strategy/action.go
```diff
@@ -0,0 +1,80 @@
...
+	if newCount != oldCount {
+		a.Meta[metaKeyCountCapped] = true
+		a.Meta[metaKeyCountOriginal] = oldCount
+		a.PushReason(fmt.Sprintf("capped count from %d to %d to stay withing limits", oldCount, newCount))
+		a.Count = &newCount
+	}
+}
+
+// PushReason updates the Reason value and stores previous Reason into Meta.
+func (a *Action) PushReason(r string) {
```
cgbaker, line 67: 🔥🔥🔥

cgbaker: ¡hawt!

strategy/action.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// Canonicalize ensures Action has proper default values.
+func (a *Action) Canonicalize() {
+	if a.Meta == nil {
+		a.Meta = make(map[string]interface{})
+	}
+
+	if a.Count == nil {
+		a.Meta[metaKeyCountNil] = true
```
cgbaker, line 29: fyi, the submitted count is already part of the scaling event, even if nil.

lgfa29: Ah good to know. So I think we can remove this meta tag, seems redundant.

PR #75, opened 2020-04-03
plugins: refactor plugin architecture to reduce code repetition.

### LINKED ISSUE #29: plugins: ensure plugins log to the same level and format as the agent
Currently the plugins setup their own logger, which can result in a situation where plugins log in a different format and at a different level to that of the agent. In order to provide logging consistency, the autoscaler should ensure any plugins are run with the same logging configuration as the agent.

### LINKED ISSUE #58: Refactor plugin architecture
The current implementation of how the agent manages plugins is heavily extrapolated from the initial prototype for APM plugins. As more plugin types were defined and implemented, the code path became brittle and repetitive, with code blocks being copy and pasted around and lacking basic testing.

jrasell:
This work aims to reduce the code repetition in relation to the
Autoscalers plugin, but also make navigating the repo easier with
an eye to the future of versioning plugin interfaces.

The largest addition is the plugin manager which is responsible
for launching, dispensing and providing access to the plugins
from the agent. This manager is not feature complete, but
provides a good basic form of central management to move forward
with. The manager now uses a common base plugin interface which
all plugins should implement. This is not currently embedded in
the plugin interfaces (future) but allows for an easier work
flow. Using the manager now also passes the agent logger to the
plugins, meaning the plugins now log to the level and format set
by the operator.

The default config has been updated to include the nomad-apm and
nomad-target plugins. This means these plugins are loaded by
default without the requirement for the user to specify. It is
not possible to turn these off because of the config merging, but
is something to keep in mind for the future.

The plugin interface declarations have been moved within the
plugin directory to their own sub-dir. From a navigation sense I
believe this makes sense, but also keeps the option of individual
sub go-modules if we wish in the future. The split has also meant
changing the default names of the nomad-local plugin, to be more
descriptive and also remove duplication. This leads to nomad-apm
and nomad-target.

The bultin plugins have been moved within a special directory and
if previously the plugin implemented two interfaces, these have
been split apart. This is due to the problem with implementing a
common base interface across all the plugins which includes a
function like the set function. It is impossible/very hard to
accommodate this setup alongside the base interface. In relation
the Makefile has been updated to reflect both the additional
plugin to build, and also the change in paths to build.

Fully tested with internal and external plugin builds.

closes #58
closes #29 

Future items to address:
- embed the base plugin interface
- handle external plugins which would have crashed by re-launching and dispensing
- add more comments to the plugins interfaces
- add additional tests
- improve the UX of developing external plugins


plugins/manager/manager.go
```diff
@@ -0,0 +1,242 @@
...
+		if err := inst.Plugin().(base.Plugin).SetConfig(pInfo.config); err != nil {
+			inst.Kill()
+			_ = multierror.Append(&mErr, fmt.Errorf("failed to set config on plugin %s: %v", pID.Name, err))
+			continue
+		}
+
+		// Store our plugin instance.
+		pm.pluginInstancesLock.Lock()
+		pm.pluginInstances[pID] = inst
+		pm.pluginInstancesLock.Unlock()
```
cgbaker, line 155: we only call this method once, so it's not necessary, but.... should we explicitly clear out the instance map at the top of the method?

also, not sure whether the repeated locking is necessary (i don't think there's any concurrent access to this struct right now); is this in preparation for concurrent loading?

agent/agent.go
```diff
@@ -341,43 +176,43 @@ func (a *Agent) handlePolicy(p *policystorage.Policy) {
...
+	var apmInst apmpkg.APM
+	var strategyInst strategypkg.Strategy
 
 	// dispense plugins
-	targetPlugin, err := a.targetManager.Dispense(p.Target.Name)
+	targetPlugin, err := a.pluginManager.Dispense(p.Target.Name, plugins.PluginTypeTarget)
 	if err != nil {
 		logger.Error("target plugin not initialized", "error", err, "plugin", p.Target.Name)
 		return
 	}
```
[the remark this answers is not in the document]
jrasell: Yes 100% in similar lines to how I expect we will handle policies which are not valid. Surfacing these errors in an easier to find way is something I am keen to tackle soon; otherwise as you know operators need to watch the logs to understand the health.

plugins/manager/external.go
```diff
@@ -0,0 +1,37 @@
...
+
+	info := &pluginInfo{
+		args:    cfg.Args,
+		config:  cfg.Config,
+		exePath: filepath.Join(pm.pluginDir, cleanPluginExecutable(cfg.Name)),
+	}
+
+	// Add the plugin.
+	pm.pluginsLock.Lock()
+	pm.plugins[plugins.PluginID{Name: cfg.Name, PluginType: pluginType}] = info
```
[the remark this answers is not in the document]
jrasell: If the operator configures two plugins with the same name but they are different plugin types I don't believe there would be an overlap. Within the config they are held as separate items, partition by their type and here the map key is a composite which would allow them to coexist. If they have the same name and plugin type, then this wouldn't make it this far and be caught up in the config loading/merging, although I am unsure exactly how this would play out.

plugins/manager/manager.go
```diff
@@ -0,0 +1,242 @@
...
+
+// PluginManager is the brains of the plugin operation and should be used to
+// manage plugin lifecycle as well as provide access to the underlying plugin
+// interfaces.
+type PluginManager struct {
+
+	// cfg is our stored configuration of plugins to dispense. The format is
+	// not ideal, but it works well and is easy to understand. The maps are
+	// keyed by plugin type and plugin driver respectively.
+	cfg map[string]map[string]*config.Plugin
```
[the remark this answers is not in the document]
jrasell: you can never have too much nesting.

plugins/manager/manager.go
```diff
@@ -0,0 +1,242 @@
...
+	if !ok {
+		return nil, fmt.Errorf("failed to dispense plugin: %s of type %s is not stored", name, pluginType)
+	}
+	return inst, nil
+}
+
+// dispensePlugins launches all configured plugins. It is responsible for
+// executing external binaries as well as setting the config on all plugins so
+// they are in a ready state. Any errors from this process will result in the
+// agent failing on startup.
```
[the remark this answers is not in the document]
jrasell: Yeah I agree. I went with this route for the PR as its very clear. If we want to just log these errors and continue the startup it is a single line change. It would be worth noting for the future to present this errors in a nicer way to the operator.

plugins/manager/external.go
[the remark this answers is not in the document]
cgbaker: 👍

plugins/manager/manager.go
jrasell: Yeah I we can clear it out. What is the advantage of doing this seeing as the method is only called once? To protect in the future if it gets called > 1 time?

Currently its locking for the sake of locking; but in the future this will be needed when the manager makes attempts to re-launch and dispense plugins which have failed after normal startup.

plugins/manager/manager.go
```diff
@@ -0,0 +1,242 @@
...
+func (pm *PluginManager) dispensePlugins() error {
+
+	var mErr multierror.Error
+
+	pm.pluginsLock.RLock()
+	p := pm.plugins
+	pm.pluginsLock.RUnlock()
```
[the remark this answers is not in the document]
jrasell: Ah yeah good catch.

I'll look at a better way of handling this and push the change(s). I think as you mention returning the information to this function and using a defer unlock should make this less horrible.

jrasell:
PR should be squashed before merge!

plugins/manager/manager.go
```diff
@@ -0,0 +1,252 @@
...
+		if err != nil {
+			_ = multierror.Append(&mErr, fmt.Errorf("failed to dispense plugin %s: %v", pID.Name, err))
+			continue
+		}
+
+		// Update our tracking to detail the plugin base information returned
+		// from the plugin itself.
+		pm.pluginsLock.Lock()
+		pm.plugins[pID].baseInfo = info
+		pm.pluginsLock.Unlock()
```
cgbaker: there's already a lock above:
https://github.com/hashicorp/nomad-autoscaler/commit/600f917bbb596a89bceaf049492ea159c27082ed#diff-b126bb70dc68df115ac86dea111afd42R128

agent/config/config.go
```diff
@@ -162,6 +169,8 @@ func Default() (*Agent, error) {
 			Address: defaultNomadAddress,
 			Region:  defaultNomadRegion,
 		},
+		APMs:    []*Plugin{{Name: defaultNomadAPM, Driver: defaultNomadAPM}},
+		Targets: []*Plugin{{Name: defaultNomadTarget, Driver: defaultNomadTarget}},
```
lgfa29, line 173: Should we use something more unique for `Name` to avoid accidentally overriding the default plugins?

agent/plugins.go
```diff
@@ -0,0 +1,51 @@
...
+		cfg[plugins.PluginTypeTarget] = a.config.Targets
+	}
+
+	// Iterate the configurations to match any nomad plugins. If these plugins
+	// have an empty config map, update this with the agents top level Nomad
+	// config. This helps stop the need for repeating the same params over and
+	// over.
+	for _, cfgs := range cfg {
+		for _, c := range cfgs {
+			if (c.Name == "nomad-apm" || c.Name == "nomad-target") && len(c.Config) == 0 {
```
lgfa29, line 44: Same as above, maybe reference the values in `plugins/manager/internal.go`?

agent/config/config.go
lgfa29: Also maybe reference the values in `plugins/manager/internal.go` to avoid a mismatch if they change?

plugins/manager/manager.go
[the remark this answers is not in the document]
lgfa29: I would say in this case it's OK since it's part of the agent startup flow, so it might be better to fail early because the fix here is most likely updating a config and restarting the agent, meaning all policies would be disrupted anyway. If we fail early small errors would be easier to detect.

A similar example is for nginx config files, if you have even one upstream wrong the server won't start (which I have a love-hate relationship with, but it's usually my fault for not checking with `-t`).

We could reevaluate this once we have signal config reload, better error visibility, more detailed health check etc.

plugins/manager/util_test.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+func Test_generateConfigMap(t *testing.T) {
+	testCases := []struct {
+		inputCfgs      map[string][]*config.Plugin
+		expectedOutput map[string]map[string]*config.Plugin
+	}{
+		{
+			inputCfgs: map[string][]*config.Plugin{
+				plugins.PluginTypeAPM:      {{Name: "nomad-apm", Driver: "nomad-apm"}},
```
lgfa29, line 22: `Driver` is a bad name....I am bad at naming things 😅

policystorage/nomad.go
```diff
@@ -63,7 +63,7 @@ func canonicalize(from *api.ScalingPolicy, to *Policy) {
 	}
 
 	if to.Target.Name == "" {
-		to.Target.Name = "local-nomad"
+		to.Target.Name = "nomad-target"
```
lgfa29, line 66: Same as before. Have this value in a single place and just reference it in other places.

policystorage/nomad.go
```diff
@@ -74,7 +74,7 @@ func canonicalize(from *api.ScalingPolicy, to *Policy) {
 	to.Target.Config["group"] = from.Target["Group"]
 
 	if to.Source == "" {
-		to.Source = "local-nomad"
+		to.Source = "nomad-apm"
```
lgfa29, line 77: Same as before. Have this value in a single place and just reference it in other places.

PR #77, opened 2020-04-06
plugins: allow target-value to handle zero count and target params

### LINKED ISSUE #76: [bug] target-value plugin doesn't handle count=0 task groups
The internal target-value plugin does not currently have the ability to handle and therefore scale task groups which have a count 0. This is because the plugin currently calculates the new count using the old count without zero checks.

The plugin should allow for the ability to scale to/from 0 if the scaling policy allows. The fix should include basic tests.

jrasell:
It is possible that the count of a task group we wish to scale is
0. It is also possible that the target the operator wishes the
metric maintain is zero. An example here would be a build system
queue, where the number of agents can scale to 0 if there is no
work to process. If work is added to the queue, the strategy is
to scale the build agents so that the queue metric is reduced to
the target of 0 items.

closes #76

### DIFF
--- a/plugins/builtin/strategy/target-value/plugin/plugin.go
+++ b/plugins/builtin/strategy/target-value/plugin/plugin.go
@@ -11,7 +11,8 @@ import (
 )
 
 const (
-	// pluginName is the name of the plugin
+	// pluginName is the unique name of the this plugin amongst strategy
+	// plugins.
 	pluginName = "target-value"
 )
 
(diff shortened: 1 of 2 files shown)


eveld:
<3

plugins/builtin/strategy/target-value/plugin/plugin.go
```diff
@@ -64,30 +75,78 @@ func (s *StrategyPlugin) Run(req strategy.RunRequest) (strategy.RunResponse, err
...
 	resp.Actions = append(resp.Actions, action)
 	return resp, nil
 }
+
+// calculateDirection is used to calculate the direction of scaling that should
+// occur, if any at all. It takes into account the current task group count in
+// order to correctly account for 0 counts.
+//
+// TODO(jrasell) the direction should probably be a type, rather than a plain
+// 	string so we don't have to return an empty string for no direction.
```
lgfa29, line 134: I think `string` here is fine, specially since this is a private method.

plugins/builtin/strategy/target-value/plugin/plugin_test.go
```diff
@@ -0,0 +1,170 @@
...
+			expectedError: fmt.Errorf("missing required field `target`"),
+			name:          "incorrect input config",
+		},
+		{
+			inputReq: strategy.RunRequest{
+				PolicyID: "test-policy",
+				Config:   map[string]string{"target": "not-the-float-you're-looking-for"},
+			},
+			expectedResp:  strategy.RunResponse{Actions: []strategy.Action{}},
+			expectedError: fmt.Errorf("invalid value for `target`: not-the-float-you're-looking-for (string)"),
```
cgbaker, line 51: 👍

PR #80, opened 2020-04-08
plugins: ensure named loggers use underscore partition.

jrasell:


cgbaker:
https://github.com/hashicorp/nomad-autoscaler/blob/ad81f841b4d92f42db201cd2ecc075e131e1fa81/agent/health.go#L46

i think you missed one.

PR #82, opened 2020-04-08
plugins: improve external plugin development UX and mimic Nomad.

jrasell:
The UX for developing external plugins was OK, but required some
knowledge of the go-plugin library in order to correctly setup
and serve the plugin. This change aims to lower the barrier to
entry on writing plugins by making the required code simpler and
also mimicking the Nomad plugin interface.

PluginInfo has been moved into the base package as part of this
work. This seems a better logic location for the definition as it
is the response object of a defined interface function inside this
package.

### DIFF
--- a/plugins/plugin.go
+++ b/plugins/plugin.go
@@ -82,8 +82,8 @@ func Serve(f PluginFactory) {
 	// Generate the plugin.
 	p := f(logger)
 	if p == nil {
-	  logger.Error("plugin factory returned nil")
-	  return
+		logger.Error("plugin factory returned nil")
+		return
 	}
 
 	// Build the base plugin configuration which is independent of the plugin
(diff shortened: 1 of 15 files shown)


jrasell:
please feel free to leave this before reviewing; the last couple I have raised have been hefty.

plugins/plugin.go
```diff
@@ -69,10 +72,35 @@ func (p PluginID) String() string {
...
-	PluginType string
+// Serve is used to serve a Nomad Autoscaler Plugin.
+func Serve(f PluginFactory) {
+	logger := hclog.New(&hclog.LoggerOptions{
+		Level:      hclog.Trace,
+		JSONFormat: true,
+	})
+
+	// Generate the plugin.
+	p := f(logger)
```
cgbaker, line 83:
```suggestion
	p := f(logger)
	if p == nil {
	  logger.Error("plugin factory returned nil")
	  return
	}
```

cgbaker: explicit error is nicer than NPE

PR #84, opened 2020-04-08
Add reference to nightly builds in README

lgfa29:
I was wondering if I should add more details about the nightly builds, but I thought it was not necessary. Any thoughts?


attachmentgenie:
the PR seems to revert the changes made in #85 perhaps you can rebase on master

jrasell:
@attachmentgenie I am happy with the overwrite (I actually fixed the conflict due to my PR) as this more accurately displays that the nightly builds work with beta2+ which encompasses 0.11.0. New releases will update this information to provide more accurate detail also. That being said i'll let @lgfa29 have the final input into this.

lgfa29:
Thanks @attachmentgenie. We decided to release a new version targeting Nomad 0.11.0 since it contains breaking changes in the scaling API from 0.11.0-beta1.

PR #95, opened 2020-04-19
policy validation: don't try so hard

### LINKED ISSUE #94: panic when `policy` does not include `strategy`
for empty scaling policy in nomad job: 
```
goroutine 85 [running]:
github.com/hashicorp/nomad-autoscaler/policystorage.validate(0xc0003c21e0, 0xc00071c060, 0x24, 0x0)
	/Users/cgbaker/go/src/github.com/hashicorp/nomad-autoscaler/policystorage/nomad.go:103 +0x334
github.com/hashicorp/nomad-autoscaler/policystorage.(*Nomad).Get(0xc00000e028, 0xc00071c060, 0x24, 0x0, 0x0, 0x0)
	/Users/cgbaker/go/src/github.com/hashicorp/nomad-autoscaler/policystorage/nomad.go:38 +0xa8
github.com/hashicorp/nomad-autoscaler/agent.(*Agent).Run.func1(0xc00078b730, 0x195b840, 0xc00020c500, 0xc00000e028, 0x1965c60, 0xc000717f80, 0xc00010e480, 0xc00071c060, 0x24)
	/Users/cgbaker/go/src/github.com/hashicorp/nomad-autoscaler/agent/agent.go:84 +0xa6
created by github.com/hashicorp/nomad-autoscaler/agent.(*Agent).Run
	/Users/cgbaker/go/src/github.com/hashicorp/nomad-autoscaler/agent/agent.go:78 +0x60f
```

cgbaker:
resolves #94 

policy was nobly trying to validate additional errors after discovery that "strategy" was of the wrong type (or, in the case of #94, did not exist). early return from this case.

### DIFF
--- a/policystorage/nomad.go
+++ b/policystorage/nomad.go
@@ -98,6 +98,7 @@ func validate(policy *api.ScalingPolicy) []error {
 	strategyList, ok := policy.Policy["strategy"].([]interface{})
 	if !ok {
 		errs = append(errs, fmt.Errorf("Policy.strategy (%T) is not a []interface{}", policy.Policy["strategy"]))
+		return errs
 	}
 
 	_, ok = strategyList[0].(map[string]interface{})
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ pkg/
 # ignore Vagrant related files
 *.log
 .vagrant
+
+# ignore GoLand IDE
+.idea
--- a/go.mod
+++ b/go.mod
@@ -17,6 +17,8 @@ require (
 	github.com/mitchellh/cli v1.0.0
 	github.com/mitchellh/copystructure v1.0.0
 	github.com/mitchellh/go-wordwrap v1.0.0 // indirect
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
+	github.com/modern-go/reflect2 v1.0.1 // indirect
 	github.com/prometheus/client_golang v1.5.1
 	github.com/prometheus/common v0.9.1
 	github.com/stretchr/testify v1.4.0
--- a/policystorage/nomad_test.go
+++ b/policystorage/nomad_test.go
@@ -1,0 +1,32 @@
+package policystorage
+
+import (
+	"fmt"
+	"reflect"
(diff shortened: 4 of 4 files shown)


policystorage/nomad.go
```diff
@@ -98,6 +98,7 @@ func validate(policy *api.ScalingPolicy) []error {
 	strategyList, ok := policy.Policy["strategy"].([]interface{})
```
cgbaker, line 98: question: why is this `[]interface{}` (i.e., []map[string]interface{} below)? why does this work?

cgbaker:
> As a future item, I am curious if we can just remove the `[]error` from the validate function. I don't believe it will ever return > 1 error with this change.

I was thinking the same thing. I imagine it's probably cribbed from Nomad, where plan/validate can tell you everything that's wrong with your job. But the function as written doesn't have that depth of capability.

policystorage/nomad.go
cgbaker: @jrasell , @lgfa29 : what about this? ^

cgbaker: i'm guessing it's some weird Go thing I didn't know about?

jrasell: I am honestly not sure. Unless we plan/planned on having scaling policies support a list of strategies, I believe we could just perform the `map[string]interface{}` cast. This is all conjecture without testing though.

lgfa29: `hcl.DecodeObject` [here](https://github.com/hashicorp/nomad/blob/master/jobspec/parse_group.go#L350) turns `strategy` into a list.

lgfa29:
> > As a future item, I am curious if we can just remove the `[]error` from the validate function. I don't believe it will ever return > 1 error with this change.
> 
> I was thinking the same thing. I imagine it's probably cribbed from Nomad, where plan/validate can tell you everything that's wrong with your job. But the function as written doesn't have that depth of capability.

Validations often return multiple errors, that's why I used a list here. We currently don't have a case where multiple errors would be returned, but it could happen in the future. Using [go-multierror](https://github.com/hashicorp/go-multierror) would probably be better.

PR #103, opened 2020-04-24
plugin: add the plugin name to the logger when launching internal.

jrasell:
When launching internal plugins, the named logger was being set
to the same value on all plugins. This makes it hard to understand
the logs of internal plugins. Adding the plugin name to the logger
helps differentiate.

### DIFF
--- a/plugins/manager/manager.go
+++ b/plugins/manager/manager.go
@@ -176,7 +176,7 @@ func (pm *PluginManager) dispensePlugins() error {
 // launchInternalPlugin is used to dispense internal plugins.
 func (pm *PluginManager) launchInternalPlugin(id plugins.PluginID, info *pluginInfo) (PluginInstance, *base.PluginInfo, error) {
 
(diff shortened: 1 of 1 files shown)


plugins/manager/manager.go
```diff
@@ -176,7 +176,7 @@ func (pm *PluginManager) dispensePlugins() error {
 // launchInternalPlugin is used to dispense internal plugins.
 func (pm *PluginManager) launchInternalPlugin(id plugins.PluginID, info *pluginInfo) (PluginInstance, *base.PluginInfo, error) {
 
-	raw := info.factory(pm.logger.ResetNamed("internal_plugin"))
+	raw := info.factory(pm.logger.ResetNamed("internal_plugin" + id.Name))
```
lgfa29, line 179: Adding a separator here, otherwise it looks like this:

```
2020-04-24T11:21:15.875-0400 [INFO]  internal_pluginprometheus: hi
```

```suggestion
	raw := info.factory(pm.logger.ResetNamed("internal_plugin." + id.Name))
```

PR #108, opened 2020-04-28
skip action when count doesn't change

lgfa29:
This PR prevents the Autoscaler from taking an action when count value doesn't change to avoid unnecessary work.


lgfa29:
> we trust `currentCount`, right?

`currentCount` is the values used to calculate the next count, so if it go stale the next count would be wrong as well.

We could re-check if it's still valid, but there's no guarantee it will always be right without some kind of global lock. 

If `currentCount` does get stale, the Autoscaler will notice this in the next evaluation cycle and correct it, so I think it's fine to trust it at this point.

PR #111, opened 2020-04-29
plugin: add tests to strategy action.

jrasell:
Adds tests to the strategy Action set of functions. The tests
uncovered a bug and lint enhancement which are fixed within this
commit.

The lint enhancement changes the PushReason function to be
un-exported as it is currently only used within the strategy pkg.

The bug occured where a reason was added to the action in
situations where no reason previously existed. The meta reason
history was appended to from the empty reason string, before the
reason string was subsequently set.

### DIFF
--- a/plugins/strategy/action.go
+++ b/plugins/strategy/action.go
@@ -54,13 +54,13 @@ func (a *Action) CapCount(min, max int64) {
 	if newCount != oldCount {
 		a.Meta[metaKeyCountCapped] = true
 		a.Meta[metaKeyCountOriginal] = oldCount
-		a.PushReason(fmt.Sprintf("capped count from %d to %d to stay within limits", oldCount, newCount))
+		a.pushReason(fmt.Sprintf("capped count from %d to %d to stay within limits", oldCount, newCount))
 		a.Count = &newCount
 	}
 }
 
 // PushReason updates the Reason value and stores previous Reason into Meta.
-func (a *Action) PushReason(r string) {
+func (a *Action) pushReason(r string) {
 	history := []string{}
 
 	// Check if we already have a reason stack in Meta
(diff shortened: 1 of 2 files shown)


plugins/strategy/action.go
```diff
@@ -70,7 +70,7 @@ func (a *Action) PushReason(r string) {
 		}
 	}
 
-	// Append current reason to history and update action
-	a.Meta[metaKeyReasonHistory] = append(history, a.Reason)
+	// Append current reason to history and update action.
```
cgbaker, line 73: my only question would be whether we should bother duplicating `a.Reason` in the reason history?

plugins/strategy/action.go
```diff
@@ -70,7 +70,7 @@ func (a *Action) PushReason(r string) {
...
 	a.Reason = r
+	a.Meta[metaKeyReasonHistory] = append(history, a.Reason)
```
cgbaker, line 75:
```suggestion
	if (a.Reason) {
		history = append(history, a.Reason)
	}
	a.Reason = r
	a.Meta[metaKeyReasonHistory] = history
```

cgbaker: my only question would be whether we should bother duplicating a.Reason in the reason history?

jrasell: Good shout. I don't have any objection to removing the duplicate. @lgfa29 thoughts?

lgfa29: So to get the full reason history you would concatenate `a.Reason + a.Meta[metaKeyReasonHistory]`?

I am ok with that.

jrasell: OK. I will follow up with an additional PR as this changes behaviour rather than fixing bugs/lint uncovered by adding tests.

PR #112, opened 2020-04-29
Move nomad query parsing

lgfa29:
Moving the query parsing logic to the Nomad APM plugin keeps it closer to where it actually is used.

### DIFF
--- a/plugins/builtin/apm/nomad/plugin/metrics.go
+++ b/plugins/builtin/apm/nomad/plugin/metrics.go
@@ -42,10 +42,12 @@ type Query struct {
 }
 
 func (a *APMPlugin) Query(q string) (float64, error) {
-	query, err := a.parseQuery(q)
+	query, err := parseQuery(q)
 	if err != nil {
 		return 0, fmt.Errorf("failed to parse query: %v", err)
 	}
+
+	a.logger.Debug("expanded query", "from", q, "to", fmt.Sprintf("%# v", query))
 
 	var resp Metrics
 	_, err = a.client.Raw().Query("/v1/metrics", &resp, nil)
@@ -94,7 +96,7 @@ func (a *APMPlugin) Query(q string) (float64, error) {
 	return result, nil
 }
 
-func (a *APMPlugin) parseQuery(q string) (*Query, error) {
+func parseQuery(q string) (*Query, error) {
 	mainParts := strings.Split(q, "/")
 	if len(mainParts) != 3 {
 		return nil, fmt.Errorf("expected <query>/<job>/group>, received %s", q)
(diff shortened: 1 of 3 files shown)


plugins/builtin/apm/nomad/plugin/metrics.go
```diff
@@ -91,3 +95,47 @@ func (a *APMPlugin) Query(q string) (float64, error) {
...
+		Group: mainParts[2],
+	}
+
+	opMetricParts := strings.Split(mainParts[0], "_")
+	if len(opMetricParts) < 2 {
+		return nil, fmt.Errorf(`expected <operation>_<metric>, received "%s"`, mainParts[0])
+	}
+
+	op := opMetricParts[0]
+	metric := strings.Join(opMetricParts[1:], "_")
```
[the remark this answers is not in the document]
lgfa29: TIL. Very useful function.

plugins/builtin/apm/nomad/plugin/metrics.go
```diff
@@ -91,3 +95,47 @@ func (a *APMPlugin) Query(q string) (float64, error) {
 
 	return result, nil
 }
+
+func parseQuery(q string) (*Query, error) {
+	mainParts := strings.Split(q, "/")
+	if len(mainParts) != 3 {
```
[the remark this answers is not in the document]
lgfa29: Since only jobs can have forward slashes I just switched the order of group and job and used `SplitN`. It's an internal representational, so changing it shouldn't be an issue.

plugins/builtin/apm/nomad/plugin/metrics.go
```diff
@@ -91,3 +95,47 @@ func (a *APMPlugin) Query(q string) (float64, error) {
...
+
+	switch op {
+	case "sum":
+		fallthrough
+	case "avg":
+		fallthrough
+	case "min":
+		fallthrough
+	case "max":
+		query.Operation = op
```
jrasell, line 135: for me this is a bit easier to read and more concise

```suggestion
	case "sum", "avg", "min", "max":
		query.Operation = op
```

policy/nomad/parser.go
```diff
@@ -103,19 +102,9 @@ func canonicalizePolicy(from *api.ScalingPolicy, to *policy.Policy) {
...
-		switch metric {
-		case "cpu":
-			metric = "nomad.client.allocs.cpu.total_percent"
-		case "memory":
-			metric = "nomad.client.allocs.memory.usage"
-		}
-
-		to.Query = fmt.Sprintf("%s/%s/%s/%s", metric, from.Target["Job"], from.Target["Group"], op)
+	if to.Source == plugins.InternalAPMNomad {
+		to.Query = fmt.Sprintf("%s/%s/%s", to.Query, from.Target["Group"], from.Target["Job"])
```
jrasell, line 108: curious if this needs to be its own conditional or could it be moved above?

lgfa29: This needs to be a separate conditional for the case when `to.Source` is defined by the operator as already being `plugins.InternalAPMNomad`, for example:

```hcl
...
scaling {
  ...
  policy {
    source = "nomad-apm"
    query  = "avg_cpu"
    ...
  }
}
```

It's a bit of an edge case, but we should cover it regardless. I am writing tests for the parser in a separate PR, so I will make sure to create one to cover this.

jrasell: ohhh yeah!

PR #113, opened 2020-04-30
Improve policy validation

lgfa29:
This PR improves policy validation in a few ways:
1. **validate the right target component**: the initial implementation only validated `ScalingPolicy.Target` which is a `map[string]string` and doesn't actually need validation.
2. **encapsulate the logic to validate an HCL block in a function**: this allows us to re-use the validation logic in different places, such as `strategy` and `target`.
3. **return error messages related to the jobspec when possible**: previously the error messages would be related to the specific internal structure of the policy, which is not significant to end users. New error messages use the same `block1->block2` notation from the Nomad docs.
4. **only test `validateScalingPolicy`**: this is the main function that calls all the others. Testing individual sub-functions failed to capture some cases.
5. **expand test cases**: test cases now have a full policy as input, as opposed to selectively remove fields from a known valid input. This makes it more clear to understand what is being tested.
6. **show validation error message in tests**: it's useful to see how validation messages will be displayed to users. You can use the `-show-validation-error` flag when running tests to seem them:
```sh
go test -v -run Test_validate ./policy/nomad/ -show-validation-error
```

### DIFF
--- a/policy/nomad/validate.go
+++ b/policy/nomad/validate.go
@@ -67,7 +67,7 @@ func validateScalingPolicy(policy *api.ScalingPolicy) error {
 //   +----------+
 //  }
 func validatePolicy(p map[string]interface{}) error {
-	const path = "ScalingPolicy.Policy"
+	const path = "scaling->policy"
 
 	var result *multierror.Error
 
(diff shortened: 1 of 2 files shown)


policy/nomad/validate.go
```diff
@@ -172,17 +170,107 @@ func validateStrategy(s map[string]interface{}) error {
...
+	}
+
+	return result.ErrorOrNil()
+}
+
+// validateHCLBlock validates the kind of unusual structure we receive when the policy HCL block is parsed.
+func validateHCLBlock(in interface{}, path, key string, validator func(in map[string]interface{}) error) error {
+	var result *multierror.Error
+
+	list, ok := in.([]interface{})
```
cgbaker, line 255: is it worthwhile allowing this to short-circuit if it's already `map[string]interface{}`?

lgfa29: Good idea, this also future-proof this code in case we change the Nomad API to return just the `map`.

PR #116, opened 2020-05-05
policy: fix Nomad target parsing when target block is empty.

jrasell:
When parsing a scaling policy from Nomad which did not contain a
target block within the policy map, the outer target map which
contained both the job and group information was ignored. This
resulted in evaluation failures and rendered the policy incapable
of being used by the autoscaler.

The fix updates the target parsing to correctly handle the above
described situation. Whilst also figuring this out, it made sense
to remove the unnecessary additional target mapping within
canonicalizePolicy() as pointed out in comments by @lgfa29. This
requires a small change to the Nomad target plugin, but results in
a simpler workflow.

### DIFF
--- a/plugins/builtin/target/nomad/plugin/plugin.go
+++ b/plugins/builtin/target/nomad/plugin/plugin.go
@@ -21,8 +21,8 @@ const (
 
 	// configKeys are the accepted configuration map keys which can be
 	// processed when performing SetConfig().
-	configKeyJobID = "job_id"
-	configKeyGroup = "group"
+	configKeyJobID = "Job"
+	configKeyGroup = "Group"
 
 	// garbageCollectionNanoSecondThreshold is the nanosecond threshold used
 	// when performing garbage collection of job status handlers.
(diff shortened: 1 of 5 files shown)


lgfa29:
Doh, that's like the most common case 🤦 

But I think the parser logic doesn't need to change that much, this should do the trick:
```diff
diff --git a/policy/nomad/parser.go b/policy/nomad/parser.go
index a2c17df..8e8ec3b 100644
--- a/policy/nomad/parser.go
+++ b/policy/nomad/parser.go
@@ -117,22 +117,22 @@ func parseTarget(targetBlock interface{}, targetAttr map[string]string) *policy.
 	}
 
 	targetMap := parseBlock(targetBlock)
-	if targetMap == nil {
+	if targetMap == nil && targetAttr == nil {
 		return nil
 	}
 
-	var configMapString map[string]string
-	configMap := parseBlock(targetMap["config"])
-
-	if configMap != nil {
-		configMapString = make(map[string]string)
+	configMapString := make(map[string]string)
+	for k, v := range targetAttr {
+		configMapString[k] = v
+	}
 
-		for k, v := range targetAttr {
-			configMapString[k] = v
-		}
+	if targetMap != nil {
+		configMap := parseBlock(targetMap["config"])
 
-		for k, v := range configMap {
-			configMapString[k] = fmt.Sprintf("%v", v)
+		if configMap != nil {
+			for k, v := range configMap {
+				configMapString[k] = fmt.Sprintf("%v", v)
+			}
 		}
 	}
```

Basically only return `nil` if the target map **and** attribute are `nil` and move the logic to copy files from the target attribute outside the map parsing logic.

jrasell:
@lgfa29 updated based on your idea.

PR #117, opened 2020-05-05
agent: support scaling cooldowns to aid oscillation migration.

### LINKED ISSUE #12: agent: add cooldown period
One of the strategies to avoid resource count instability is to add a cooldown period in which changes in metric values do not affect scaling decisions.

The cooldown is defined as a time period (e.g., '5s') and its value is derived out of the following locations:

* default global value (least precedence)
* default Autoscaler agent configuration
* policy specific (highest precedence)

jrasell:
Cooldown is now supported within a scaling policy as a time
duration. If the scaling policy omits the cooldown field, a
default value will be used. The default value can be controlled
by the CLI and has a default value of 5mins.

In the event the agent successfully registers a scaling action,
the policy manager will be instructed to place the policy eval
handler in cooldown by updating the ticker to the cooldown time
duration. The ticker is reset to its eval time duration on the
first tick out of cooldown.

Target plugins can optionally return a meta entry to detail the
last scaling event from their prespective. This allows the
autoscaler to enforce cooldowns, even when scaling actions have
been triggered manually by operators. The flow is similar to that
mentioned above, the only difference being that the ticker update
is based on the calculated remaning cooldown time from the last
event timestamp.

closes #12

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -123,7 +123,7 @@ type Policy struct {
 	// DefaultCooldown is the default cooldown parameter added to all policies
 	// which do not explicitly configure the parameter.
 	DefaultCooldown    time.Duration
-	DefaultCooldownHCL string `hcl:"default_cooldown,optional" json:"-"`
+	DefaultCooldownHCL string `hcl:"default_cooldown,optional"`
 }
 
 const (
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -46,6 +46,9 @@ type Agent struct {
 	// Nomad is the configuration used to setup the Nomad client.
 	Nomad *Nomad `hcl:"nomad,block"`
 
+	// Policy is the configuration used to setup the policy manager.
+	Policy *Policy `hcl:"policy,block"`
+
 	APMs       []*Plugin `hcl:"apm,block"`
 	Targets    []*Plugin `hcl:"target,block"`
 	Strategies []*Plugin `hcl:"strategy,block"`
@@ -113,6 +116,16 @@ type Plugin struct {
 	Config map[string]string `hcl:"config,optional"`
 }
 
+// Policy holds the configuration information specific to the policy manager
+// and resulting policy parsing.
+type Policy struct {
+
+	// DefaultCooldown is the default cooldown parameter added to all policies
+	// which do not explicitly configure the parameter.
+	DefaultCooldown    time.Duration
+	DefaultCooldownHCL string `hcl:"default_cooldown,optional" json:"-"`
+}
+
 const (
 	// defaultLogLevel is the default log level used for the Autoscaler agent.
 	defaultLogLevel = "info"
@@ -139,6 +152,10 @@ const (
 	// defaultNomadRegion is the default Nomad region to use when performing
 	// Nomad API calls.
 	defaultNomadRegion = "global"
+
+	// defaultPolicyCooldown is the default time duration applied to policies
+	// which do not explicitly configure a cooldown.
+	defaultPolicyCooldown = 5 * time.Minute
 )
 
 // Default is used to generate a new default agent configuration.
@@ -162,6 +179,9 @@ func Default() (*Agent, error) {
 		Nomad: &Nomad{
 			Address: defaultNomadAddress,
 			Region:  defaultNomadRegion,
+		},
+		Policy: &Policy{
+			DefaultCooldown: defaultPolicyCooldown,
 		},
 		APMs:    []*Plugin{{Name: plugins.InternalAPMNomad, Driver: plugins.InternalAPMNomad}},
 		Targets: []*Plugin{{Name: plugins.InternalTargetNomad, Driver: plugins.InternalTargetNomad}},
@@ -191,6 +211,10 @@ func (a *Agent) Merge(b *Agent) *Agent {
 
 	if b.Nomad != nil {
 		result.Nomad = result.Nomad.merge(b.Nomad)
+	}
+
+	if b.Policy != nil {
+		result.Policy = result.Policy.merge(b.Policy)
 	}
 
 	if len(result.APMs) == 0 && len(b.APMs) != 0 {
@@ -305,6 +329,15 @@ func (p *Plugin) copy() *Plugin {
 	return &c
 }
 
+func (p *Policy) merge(b *Policy) *Policy {
+	result := *p
+
+	if b.DefaultCooldown != 0 {
+		result.DefaultCooldown = b.DefaultCooldown
+	}
+	return &result
+}
+
 // pluginConfigSetMerge merges two sets of plugin configs. For plugins with the
 // same name, the configs are merged.
 func pluginConfigSetMerge(first, second []*Plugin) []*Plugin {
@@ -356,6 +389,15 @@ func parseFile(file string, cfg *Agent) error {
 		}
 		cfg.DefaultEvaluationInterval = d
 	}
+
+	if cfg.Policy != nil && cfg.Policy.DefaultCooldownHCL != "" {
+		d, err := time.ParseDuration(cfg.Policy.DefaultCooldownHCL)
+		if err != nil {
+			return err
+		}
+		cfg.Policy.DefaultCooldown = d
+	}
+
 	return nil
 }
 
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -22,7 +22,7 @@ func Test_Default(t *testing.T) {
 	assert.Equal(t, def.Nomad.Address, "http://127.0.0.1:4646")
 	assert.Equal(t, "127.0.0.1", def.HTTP.BindAddress)
 	assert.Equal(t, 8080, def.HTTP.BindPort)
-	assert.Equal(t, def.Policy.DefaultCooldown, time.Duration(300000000000))
+	assert.Equal(t, def.Policy.DefaultCooldown, 5 * time.Minute)
 	assert.Len(t, def.APMs, 1)
 	assert.Len(t, def.Targets, 1)
 }
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -22,6 +22,7 @@ func Test_Default(t *testing.T) {
 	assert.Equal(t, def.Nomad.Address, "http://127.0.0.1:4646")
 	assert.Equal(t, "127.0.0.1", def.HTTP.BindAddress)
 	assert.Equal(t, 8080, def.HTTP.BindPort)
+	assert.Equal(t, def.Policy.DefaultCooldown, time.Duration(300000000000))
 	assert.Len(t, def.APMs, 1)
 	assert.Len(t, def.Targets, 1)
 }
@@ -69,6 +70,9 @@ func TestAgent_Merge(t *testing.T) {
 			TLSServerName: "cows-or-pets",
 			SkipVerify:    true,
 		},
+		Policy: &Policy{
+			DefaultCooldown: 20 * time.Minute,
+		},
 		APMs: []*Plugin{
 			{
 				Name:   "influx-db",
@@ -111,6 +115,9 @@ func TestAgent_Merge(t *testing.T) {
 			TLSServerName: "cows-or-pets",
 			SkipVerify:    true,
 		},
+		Policy: &Policy{
+			DefaultCooldown: 20 * time.Minute,
+		},
 		APMs: []*Plugin{
 			{
 				Name:   "nomad-apm",
@@ -150,6 +157,7 @@ func TestAgent_Merge(t *testing.T) {
 	assert.Equal(t, expectedResult.Nomad, actualResult.Nomad)
 	assert.Equal(t, expectedResult.PluginDir, actualResult.PluginDir)
 	assert.Equal(t, expectedResult.DefaultEvaluationInterval, actualResult.DefaultEvaluationInterval)
+	assert.Equal(t, expectedResult.Policy, actualResult.Policy)
 	assert.ElementsMatch(t, expectedResult.APMs, actualResult.APMs)
 	assert.ElementsMatch(t, expectedResult.Targets, actualResult.Targets)
 	assert.ElementsMatch(t, expectedResult.Strategies, actualResult.Strategies)
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -124,10 +124,9 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 
 			eval, err := h.handleTick(ctx, currentPolicy)
 			if err != nil {
-				if err == context.Canceled {
-					return
+				if err != context.Canceled {
+					h.log.Error(err.Error())
 				}
-				h.log.Error(err.Error())
 				return
 			}
 
@@ -194,13 +193,12 @@ func (h *Handler) handleTick(ctx context.Context, policy *Policy) (*Evaluation, 
 		return eval, nil
 	}
 
-	// Check whether we need to enter cooldown or not.
-	if !h.isInCooldown(policy.Cooldown, curTime, lastTS) {
+	// Calculate the remaining time period left on the cooldown. If this is 0
+	// or below, we do not need to enter cooldown.
+	cdPeriod := h.calculateRemainingCooldown(policy.Cooldown, curTime, int64(lastTS))
+	if cdPeriod <= 0 {
 		return eval, nil
 	}
-
-	// Calculate the remaining time period left on the cooldown.
-	cdPeriod := h.calculateRemainingCooldown(policy.Cooldown, curTime, int64(lastTS))
 
 	// Enforce the cooldown which will block until complete. A false response
 	// means we did not reach the end of cooldown due to a request to shutdown.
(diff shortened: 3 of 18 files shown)


policy/handler_test.go
```diff
@@ -0,0 +1,73 @@
...
+		inputCooldown  time.Duration
+		inputTimestamp int64
+		inputLastEvent uint64
+		expectedOutput bool
+		name           string
+	}{
+		{
+			inputCooldown:  100 * time.Hour,
+			inputTimestamp: time.Now().UTC().UnixNano(),
+			inputLastEvent: uint64(time.Now().UTC().UnixNano()),
```
lgfa29, line 22: I don't know much about the details of testing in Go, but it might be better to use [`time.Date()`](https://golang.org/pkg/time/#Date) to have deterministic inputs for tests.

policy/manager.go
```diff
@@ -127,3 +128,20 @@ func (m *Manager) stopHandler(h *Handler) {
...
+	// Attempt to grab the handler and pass the enforcement onto the
+	// implementation. Its possible cooldown is requested on a policy which
+	// gets removed, its not a problem but log to aid debugging.
+	if handler, ok := m.handlers[PolicyID(id)]; ok && handler.cooldownCh != nil {
+		handler.cooldownCh <- t
+	} else {
+		m.log.Debug("attempted to set cooldown on on-existent handler", "policy_id", id)
+	}
+
+	m.lock.RUnlock()
```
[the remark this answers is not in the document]
jrasell: No. Seeing as the call chain is small and there is only one possible exit point from the function I didn't see the need for a defer.

policy/handler.go
```diff
@@ -193,6 +218,25 @@ func (h *Handler) generateEvaluation(policy *Policy) (*Evaluation, error) {
...
 
+	// If the target status includes a last event meta key, check for cooldown
+	// due to out-of-band events.
+	if ts, ok := status.Meta[targetpkg.MetaKeyLastEvent]; ok {
+
+		// Convert the last event string.
+		lastTS, err := strconv.ParseUint(ts, 10, 64)
+		if err != nil {
+			h.log.Error("failed to parse last event timestamp as uint64", "error", err)
+			return nil, nil
```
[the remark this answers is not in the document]
jrasell: Fair point; I wasn't too sure about this one and could have gone either way.
I'll fix it up so that we make an eval if this occurs.

policy/handler_test.go
jrasell: Makes sense to be explicit although I would never expect this test to fail. I'll fix it up to use date.

policy/manager.go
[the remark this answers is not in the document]
lgfa29: It's more about future-proofing I guess. If a new exit point gets added later we would have to remember to unlock.

agent/config/config.go
```diff
@@ -139,6 +152,11 @@ const (
 	// defaultNomadRegion is the default Nomad region to use when performing
 	// Nomad API calls.
 	defaultNomadRegion = "global"
+
+	// defaultPolicyCooldown is the default time duration applied to policies
+	// which do not explicitly configure a cooldown. 300000000000 nanoseconds
+	// represents 5 minutes.
+	defaultPolicyCooldown = time.Duration(300000000000)
```
jippi, line 159: why do you hate

```suggestion
	defaultPolicyCooldown = 5 * time.Minute
```

agent/config/config_test.go
```diff
@@ -69,6 +70,9 @@ func TestAgent_Merge(t *testing.T) {
 			TLSServerName: "cows-or-pets",
 			SkipVerify:    true,
 		},
+		Policy: &Policy{
+			DefaultCooldown: 1800000000000,
```
jippi, line 74: feels like something that should be a constant or a var somewhere since these values are repeated so much :)

policy/handler.go
```diff
@@ -143,7 +203,7 @@ func (h *Handler) Stop() {
 
 // generateEvaluation returns an evaluation if the policy needs to be evaluated.
 // Returning an error will stop the handler.
-func (h *Handler) generateEvaluation(policy *Policy) (*Evaluation, error) {
+func (h *Handler) generateEvaluation(policy *Policy, evalTimestamp int64) (*Evaluation, error) {
```
lgfa29, line 206: Were you planning on storing this value in the evaluation or is it a leftover from the previous implementation?

policy/handler.go
```diff
@@ -218,3 +278,56 @@ func (h *Handler) updateHandler(current, next *Policy) {
...
+	}
+}
+
+// isInCooldown is used to check whether the handler should be in cooldown or
+// not. This is used when performing a policy evaluation and allows the
+// autoscaler to enforce a cooldown on objects that have been scaled outside of
+// the autoscaler.
+func (h *Handler) isInCooldown(cd time.Duration, ts int64, lastEvent uint64) bool {
+	if threshold := int64(lastEvent) + cd.Nanoseconds(); threshold >= ts {
+		h.log.Info("policy is in cooldown due to out-of-band change")
```
lgfa29, line 323: This log message would be better outside of this function since it's implying some external condition on when it's called, reducing re-usability.

After it's moved, what's left is just:

```go
func (h *Handler) isInCooldown(cd time.Duration, ts int64, lastEvent uint64) bool {
	return int64(lastEvent)+cd.Nanoseconds() >= ts
}
```

policy/handler.go
```diff
@@ -218,3 +278,56 @@ func (h *Handler) updateHandler(current, next *Policy) {
...
 }
+
+// enforceCooldown blocks until the cooldown period has been reached, or the
+// handler has been instructed to exit. When this function returns the caller
+// should perform a check to identify if the handler should exit or not. This
+// can be done:
+//
+// if ctx.Err() != nil || !h.running {
+//   return
+// }
```
lgfa29, line 289: Maybe we can avoid having to check this by returning a `bool`. Something like: 

```go
chilled := enforceCooldown(ctx, t)
if !chilled {
  // cooldown was cancelled, we must return
  return
}
```

Or maybe an `error`...but it's not really an error, so maybe a `bool` is probably fine ¯\\_(ツ)_/¯

policy/manager.go
```diff
@@ -127,3 +128,24 @@ func (m *Manager) stopHandler(h *Handler) {
...
+	// gets removed, its not a problem but log to aid debugging.
+	//
+	// Thinking(jrasell): when enforcing cooldown, should we calculate the
+	// duration based on the remaining time between calling this function and
+	// it actually running. Obtaining the lock could cause a delay which may
+	// skew the cooldown period, but this is likely very small.
+	if handler, ok := m.handlers[PolicyID(id)]; ok && handler.cooldownCh != nil {
+		handler.cooldownCh <- t
+	} else {
+		m.log.Debug("attempted to set cooldown on on-existent handler", "policy_id", id)
```
lgfa29, line 149:
```suggestion
		m.log.Debug("attempted to set cooldown on non-existent handler", "policy_id", id)
```

policy/manager.go
```diff
@@ -127,3 +128,24 @@ func (m *Manager) stopHandler(h *Handler) {
...
+	defer m.lock.RUnlock()
+
+	// Attempt to grab the handler and pass the enforcement onto the
+	// implementation. Its possible cooldown is requested on a policy which
+	// gets removed, its not a problem but log to aid debugging.
+	//
+	// Thinking(jrasell): when enforcing cooldown, should we calculate the
+	// duration based on the remaining time between calling this function and
+	// it actually running. Obtaining the lock could cause a delay which may
+	// skew the cooldown period, but this is likely very small.
```
lgfa29, line 145: Yeah, I would agree that the time to acquire a lock is negligible. Specially considering that the cooldown period is in the order of magnitude of minutes.

policy/handler.go
```diff
@@ -218,3 +278,56 @@ func (h *Handler) updateHandler(current, next *Policy) {
...
+
+	// Using a timer directly is mentioned to be more efficient than
+	// time.After() as long as we ensure to call Stop(). So setup a timer for
+	// use and defer the stop.
+	timer := time.NewTimer(t)
+	defer timer.Stop()
+
+	// Cooldown should not mean we miss other handler control signals. So wait
+	// on all the channels desired here.
+	for {
```
lgfa29, line 305: This loop doesn't seem necessary. Once a channel is triggered we exit right away.

agent/config/config_test.go
```diff
@@ -22,6 +22,7 @@ func Test_Default(t *testing.T) {
 	assert.Equal(t, def.Nomad.Address, "http://127.0.0.1:4646")
 	assert.Equal(t, "127.0.0.1", def.HTTP.BindAddress)
 	assert.Equal(t, 8080, def.HTTP.BindPort)
+	assert.Equal(t, def.Policy.DefaultCooldown, time.Duration(300000000000))
```
jippi, line 25:
```suggestion
	assert.Equal(t, def.Policy.DefaultCooldown, 5 * time.Minute)
```

policy/handler.go
```diff
@@ -115,13 +121,65 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
...
+			// been re-deployed.
+			if ts, ok := eval.TargetStatus.Meta[targetpkg.MetaKeyLastEvent]; ok {
+
+				// Convert the last event string. If an error occurs, just log and
+				// continue with the evaluation. A malformed timestamp shouldn't mean
+				// we skip scaling.
+				lastTS, err := strconv.ParseUint(ts, 10, 64)
+				if err != nil {
+					h.log.Error("failed to parse last event timestamp as uint64", "error", err)
+				} else {
```
jippi, line 153: completely **personal** preference, but `} else {` should be avoided.

In this case, it can be avoided (and along with it, the indentation) by either using a `label` to skip down to ´evalCh`  or have `isInCooldown` return false if `lastTS.IsZero()` is `true` - or spin out the entire logic block here into a function - it's doing quite a lot of branching and indentation already :)

policy/handler.go
```diff
@@ -141,6 +156,61 @@ func (h *Handler) Stop() {
...
+		return eval, nil
+	}
+
+	// Calculate the remaining time period left on the cooldown.
+	cdPeriod := h.calculateRemainingCooldown(policy.Cooldown, curTime, int64(lastTS))
+
+	// Enforce the cooldown which will block until complete. A false response
+	// means we did not reach the end of cooldown due to a request to shutdown.
+	if !h.enforceCooldown(ctx, cdPeriod) {
+		return nil, fmt.Errorf("cooldown period interupted by shutdown request")
```
lgfa29, line 205: This will cause an error message to be log, which might be confusing and distract operators during a debug session, but we need to return an error here so the main loop terminates without risking reading from any of the other channels.

We could use a custom error type here (or maybe `context.Canceled`) and check for it in the main loop to avoid logging.

policy/handler.go
```diff
@@ -218,3 +291,46 @@ func (h *Handler) updateHandler(current, next *Policy) {
...
+	// on all the channels desired here.
+	select {
+	case <-timer.C:
+		return true
+	case <-ctx.Done():
+		return false
+	case <-h.doneCh:
+		return false
+
+	}
```
cgbaker, line 321:
```suggestion
	}
```

policy/handler.go
```diff
@@ -218,3 +291,46 @@ func (h *Handler) updateHandler(current, next *Policy) {
...
+// not. This is used when performing a policy evaluation and allows the
+// autoscaler to enforce a cooldown on objects that have been scaled outside of
+// the autoscaler.
+func (h *Handler) isInCooldown(cd time.Duration, ts int64, lastEvent uint64) bool {
+	return int64(lastEvent)+cd.Nanoseconds() >= ts
+}
+
+// calculateRemainingCooldown calculates the remaining cooldown based on the
+// time since the last event.
+func (h *Handler) calculateRemainingCooldown(cd time.Duration, ts, lastEvent int64) time.Duration {
```
cgbaker, line 334: combining `calculateRemainingCooldown` and `isInCooldown` into one method will help make sure that we don't accidentally mis-call the latter with a different `currentTime` than the former and a negative number. 

in fact, `isInCooldown == (calculateRemainingCooldown > 0)` (check my math), so they're redundant

policy/handler.go
```diff
@@ -218,3 +291,46 @@ func (h *Handler) updateHandler(current, next *Policy) {
 		h.ticker = time.NewTicker(next.EvaluationInterval)
 	}
 }
+
+// enforceCooldown blocks until the cooldown period has been reached, or the
```
cgbaker, line 295: we should probably document it somewhere, but one of the really nice things about the way that we block here and the fact that we use `NewTicker` for `h.ticker` is the following: `h.ticker` will "drop items on the floor" as the ticks come in. our approach doesn't stop or alter the ticker, so it continues to tick while we are blocked in the cooldown. this means that if `EvaluationInterval < Cooldown`, we will immediately trigger an evaluation after exiting cooldown (as in the previous PR). however, if `Cooldown < EvaluationInterval`, then cooldown doesn't have any effect; but that seems semantically sound.

policy/handler.go
```diff
@@ -218,3 +291,46 @@ func (h *Handler) updateHandler(current, next *Policy) {
 		h.ticker = time.NewTicker(next.EvaluationInterval)
```
cgbaker, line 291: btw, this discards the current evaluation interval by resetting the ticker to zero. that's fine (and probably moot) for "short" evaluation intervals. for "longer" ones, maybe that would be undesirable? in which case, we end up with a conundrum where we need to have a short (potentially, zero) first interval and then transition to the new interval (probably would want to use a custom Ticker implementation).

agent/config/config.go
```diff
@@ -113,6 +116,16 @@ type Plugin struct {
 	Config map[string]string `hcl:"config,optional"`
 }
 
+// Policy holds the configuration information specific to the policy manager
+// and resulting policy parsing.
+type Policy struct {
```
cgbaker, line 121: maybe this is just me, but... I know that this is `config.Policy`, which is a distinct object from `Policy.Policy`. but it is a little confusing (at least, in GitHub PRs) that they have the same name. just my two cents. by the same token, `config.PolicyConfig` would be a little redundant.

policy/handler.go
```diff
@@ -115,14 +121,26 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 			h.updateHandler(currentPolicy, &p)
 			currentPolicy = &p
 		case <-h.ticker.C:
-			eval, err := h.generateEvaluation(currentPolicy)
+
+			eval, err := h.handleTick(ctx, currentPolicy)
 			if err != nil {
+				if err == context.Canceled {
+					return
+				}
```
cgbaker, line 129: probably could have reversed this... 
if err != context { log }
return

policy/handler.go
cgbaker: also, because `true/false` are ambiguous returns, i've seen (mostly in other languages) where folks recommend "named" values (sometimes even protected by type safety, apparently not in golang) for readability. like so: 
https://play.golang.org/p/WRfAPGDAJCF
is this something that folks do in golang? is there a more idiomatic way to do this?

policy/handler.go
```diff
@@ -115,13 +121,67 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
...
 			currentPolicy = &p
 		case <-h.ticker.C:
-			eval, err := h.generateEvaluation(currentPolicy)
+
+			// Timestamp the invocation of this evaluation run. This can be
+			// used when checking cooldown or emitting metrics to ensure some
+			// consistency.
+			curTime := time.Now().UTC().UnixNano()
+
+			eval, err := h.generateEvaluation(currentPolicy, curTime)
```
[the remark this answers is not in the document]
jrasell: It is indeed too much and something that hasn't changed as part of this work. We have #119 to track fixing this.

agent/config/config.go
jrasell: how does `PolicyManager` sound?

agent/config/config.go
```diff
@@ -113,6 +116,16 @@ type Plugin struct {
...
+
+	// DefaultCooldown is the default cooldown parameter added to all policies
+	// which do not explicitly configure the parameter.
+	DefaultCooldown    time.Duration
+	DefaultCooldownHCL string `hcl:"default_cooldown,optional" json:"-"`
```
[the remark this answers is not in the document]
jrasell: So the silly answer, its how Nomad does it. The spec states it is a way to set the field as private the therefore gets skipped when marshalling.

I have tested this locally without the json tag on CLI flag, JSON and HCL config methods and all get picked up. I'll remove it and then raise a follow up to remove it from the eval period. It is potentially something required with HCL1, and not HCL2.

policy/handler.go
jrasell: I am not the biggest fan of bool returns like this. I had considered using an error to identify if the cooldown completed without interruption, but this is just checked and wouldn't be logged so seems incorrect.

The function could use a named return to aid readers:
```
func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) (completed bool) {
...

	select {
	case <-timer.C:
                completed = true
		return
	case <-ctx.Done():
		return
	case <-h.doneCh:
		return
	}
}
```

agent/config/config.go
cgbaker: i think I'm going to retract this comment.

agent/config/config.go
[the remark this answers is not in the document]
cgbaker: oh yeah, I remember that now 👍

policy/handler.go
cgbaker: that's definitely better. mainly just a lazy gripe on my part that one has to go look at the function docs/code to see whether caller should look for `true` or `false`

policy/handler.go
lgfa29: I think the key here is the order of magnitude of "short" and "long". Horizontal autoscalers are usually targeted for long (multi-day) running services. This condition would only be triggered by a change in `evaluation_interval`, which would probably be set in the order of minutes at most, which sounds long when doing something like a quick test, but it's actually short when compared to the lifecycle of the service.

PR #120, opened 2020-05-09
Improve demo

### LINKED ISSUE #70: Improve demo
The current demo relies on a under-powered container to accumulate connections for triggering the Autoscaling policy.

A better approach would be to use a system that can keep connections open in a more reliable way, for example using the count dash or something like [toxiproxy](https://github.com/Shopify/toxiproxy)

lgfa29:
This PR improves the current demo in the following ways:

* add Grafana back to help visualize the Autoscaler actions
* Grafana dashboard is deployed automatically and doesn't require authorization for access
* add [toxiproxy](https://github.com/shopify/toxiproxy) as a sidecar to `webapp` to simulate a slow response without limiting resources on `webapp`. This allows `webapp` to start faster, consequently making the demo faster as well. 
* resource limits were fine tuned to allow more room for scaling `webapp`
* minor improvements on jobspecs, such as pinning Docker image tags, using interpolation instead of hard coding values and using the new `lifecycle` stanza
* fixed a bug where Prometheus would stop scrapping HAProxy metrics if HAproxy went down

closes #70


demo/vagrant/jobs/grafana.nomad
```diff
@@ -0,0 +1,94 @@
...
+  allowUiUpdates: false
+  options:
+    path: /var/lib/grafana/dashboards
+EOH
+
+        destination = "local/dashboards/nomad-autoscaler.yaml"
+      }
+
+      artifact {
+        source      = "https://raw.githubusercontent.com/hashicorp/nomad-autoscaler/improve-demo/demo/vagrant/files/dashboard.json"
```
lgfa29, line 70: TODO: update this link to `master` once PR is merge

demo/vagrant/jobs/webapp.nomad
```diff
@@ -50,14 +46,82 @@ job "webapp" {
...
+      template {
+        data = <<EOH
+#!/bin/sh
+
+set -ex
+
+/go/bin/toxiproxy -host 0.0.0.0  &
+/go/bin/toxiproxy-cli create webapp -l 0.0.0.0:{{ env "NOMAD_PORT_webapp"  }} -u {{ env "NOMAD_ADDR_webapp_http" }}
+/go/bin/toxiproxy-cli toxic add -n latency -t latency -a latency=1000 -a jitter=500 webapp
+tail -f /dev/null
```
lgfa29, line 91: This is awful but I couldn't find a way to bring back the `toxiproxy` process (`fg` didn't work). Any ideas?

jrasell: The inability to configure the toxics via a config file makes me sad.
I can't think of a better way currently; i'll keep thinking about it though.

cgbaker: one alternative is an init task that performs the toxiproxy-cli commands. it looks like this: 
```
    task "toxiproxy-config" {
      driver = "docker"

      lifecycle {
        hook    = "prestart"
        sidecar = false
      }

      config {
        image      = "shopify/toxiproxy:2.1.4"
        entrypoint = ["/entrypoint.sh"]

        volumes = [
          "local/entrypoint.sh:/entrypoint.sh",
        ]
      }

      template {
        data = <<EOH
#!/bin/sh

set -ex

while ! nc -z $NOMAD_IP_toxiproxy_api $NOMAD_PORT_toxiproxy_api; do
  echo "Waiting toxiproxy to launch on $NOMAD_PORT_toxiproxy_api..."
  sleep 2
done

echo "toxiproxy is listening"

/go/bin/toxiproxy-cli -h $NOMAD_ADDR_toxiproxy_api create webapp -l 0.0.0.0:{{ env "NOMAD_PORT_webapp"  }} -u {{ env "NOMAD_ADDR_webapp_http" }}
/go/bin/toxiproxy-cli -h $NOMAD_ADDR_toxiproxy_api toxic add -n latency -t latency -a latency=1000 -a jitter=500 webapp
        EOH

        destination = "local/entrypoint.sh"
        perms       = "755"
      }

      resources {
        cpu    = 100
        memory = 32
      }
    }
```

then toxiproxy sidecar can run like usual.

cgbaker: in my testing this task sometimes fails, because toxiproxy will apparently start listening before it's ready to run, but then the init is restarted. the nomad restart interval for this task may need some tuning.

demo/vagrant/README.md
```diff
@@ -42,7 +42,16 @@ Prometheus can now be started, again using the `nomad run` command. The virtual
...
+To visualize the state of the system and better understand what actions are being taken by the Autoscaler, you can run the Grafana service which comes with a sample dashboard that can be accessed from your browser at [http://localhost:3000/d/8QlvShyZz/nomad-autoscaler-demo](http://localhost:3000/d/8QlvShyZz/nomad-autoscaler-demo?orgId=1&refresh=5s).
+
+```
+$ nomad run grafana.nomad
+```
+
+We will explore the details of each panel shortly.
+
 ## Start the Autoscaler Component Jobs
+
```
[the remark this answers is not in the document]
lgfa29: Ah I didn't notice the difference. I think the editor I use for Markdown added them automatically. I make sure there's no space from now on.

demo/vagrant/jobs/webapp.nomad
lgfa29: Hum...I might need to sprinkle some magic `sleep` here and there. I will try to reproduce it.

lgfa29: I've used `wget` instead of `nc` to make sure `toxiproxy` is ready. 

I also found a potential bug in Nomad where allocs in a deployment could be marked as `unhealthy` even though they passed the checks. This could be problem you were experiencing.

PR #134, opened 2020-05-20
Re-connect to Nomad if connection is lost

### LINKED ISSUE #119: Autoscaler fails to re-connect to Nomad cluster
When the Autoscaler loses connection to the Nomad cluster, all policy handlers are stopped and are never re-evaluated, even when the Nomad cluster comes back up.

lgfa29:
This PR fixes an issue where the Autoscaler would fail to re-connect to the Nomad server if the connection was dropped.

The main cause for this issue is that the `policyIDsErrCh` was not being consumed, and so the goroutine monitoring the list of policy IDs would block in case of an error.

I also uncovered some more subtle ways that the Autoscaler could fail to re-connect properly:
* policy handlers would exit in case of connection error and never be re-created.
* the policy list ID monitor would silently `continue` if the Autoscler re-connected to a Nomad cluster where the list of policies were different but the modify index had the same value.

To fix these problems I made a few assumptions:
* Policy handlers only exit when their context is closed. They will keep looping even if an error occurs.
* The policy manager is responsible for stopping handlers in case of an [unrecoverable errors](https://github.com/hashicorp/nomad-autoscaler/pull/134/files#diff-bbdb50489b1a80cb45d42cf486e5870bR184-R193).
* The policy manager will re-run if it exists until its context is closed.

Closes #119

### DIFF
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -102,7 +102,14 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 			return
 		case <-h.doneCh:
 			return
+
 		case err := <-h.errCh:
+			// In case of error, log the error message and loop around.
+			// Handlers never stop running unless ctx.Done() or doneCh is
+			// closed.
+
+			// multierror.Error objects are logged differently to allow for a
+			// more structured output.
 			merr, ok := err.(*multierror.Error)
 			if ok && len(merr.Errors) > 1 {
 				// Transform Errors into a slice of strings to avoid logging
@@ -112,22 +119,24 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 					errors[i] = e.Error()
 				}
 				h.log.Error(errors[0], "errors", errors[1:])
-				return
+			} else {
+				h.log.Error(err.Error())
 			}
+			continue
 
-			h.log.Error(err.Error())
-			return
 		case p := <-h.ch:
 			h.updateHandler(currentPolicy, &p)
 			currentPolicy = &p
-		case <-h.ticker.C:
 
+		case <-h.ticker.C:
 			eval, err := h.handleTick(ctx, currentPolicy)
 			if err != nil {
-				if err != context.Canceled {
-					h.log.Error(err.Error())
+				if err == context.Canceled {
+					// Context was canceled, return to stop the handler.
+					return
 				}
-				return
+				h.log.Error(err.Error())
+				continue
 			}
 
 			if eval != nil {
@@ -135,9 +144,9 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 			}
 
 		case ts := <-h.cooldownCh:
-
 			// Enforce the cooldown which will block until complete.
 			if !h.enforceCooldown(ctx, ts) {
+				// Context was canceled, return to stop the handler.
 				return
 			}
 		}
--- a/policy/manager.go
+++ b/policy/manager.go
@@ -3,6 +3,7 @@ package policy
 import (
 	"context"
 	"fmt"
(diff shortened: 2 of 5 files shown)


policy/nomad/source.go
```diff
@@ -94,12 +94,6 @@ func (s *Source) MonitorIDs(ctx context.Context, resultCh chan<- []policy.Policy
 				continue
 			}
 
-			// If the index has not changed, the query returned because the timeout
-			// was reached, therefore start the next query loop.
-			if !blocking.IndexHasChanged(meta.LastIndex, q.WaitIndex) {
```
jrasell, line 99: > the policy list ID monitor would silently continue if the Autoscler re-connected to a Nomad cluster where the list of policies were different but the modify index had the same value.

I guess this removal relates to the above PR mention. If not please ignore this comment. In what situation would we reconnect to a cluster that has an altered list output but the table index has not changed?

policy/manager.go
```diff
@@ -149,3 +180,23 @@ func (m *Manager) EnforceCooldown(id string, t time.Duration) {
...
+//
+// An error is considered unrecoverable if it has the potential to affect the
+// policy manager's internal state, and so the policy manager should re-run.
+//
+// Example of an unrecoverable error: the Nomad server becomes unreachable;
+// upon reconnecting, the state of the Nomad cluster could be different from
+// what's stored in the policy manager (e.g., a different cluster becomes
+// available in the same address).
+func isUnrecoverableError(err error) bool {
+	unrecoverableErrors := []string{"connection refused", "EOF"}
```
jrasell, line 195: The net stdlib package has [custom errors](https://github.com/golang/go/blob/4ec4a792f6e60c9c4f40a928650c66419ee0e446/src/net/net.go#L392) which have a `Temporary()` function. I am not sure this would cover the cases you have specifically found, but something worth noting at least.

policy/handler.go
```diff
@@ -102,7 +102,14 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 			return
 		case <-h.doneCh:
 			return
+
 		case err := <-h.errCh:
+			// In case of error, log the error message and loop around.
+			// Handlers never stop running unless ctx.Done() or doneCh is
+			// closed.
+
+			// multierror.Error objects are logged differently to allow for a
```
jrasell, line 111: if we have to do this somewhere else in the future, we should certainly pull the logic into a helper.

policy/nomad/source.go
lgfa29: It's an odd case that is very unlikely to happen in production, but if you run a Nomad agent in dev mode, register a job with a scaling policy, then stop the dev agent and register the same job again, you will end up with the same index (since the same modifications were executed) but a different list (the policy ID will be different).

The root of this issue is that lists don't have an ID, so you can't really match an index to a list like, for example, a policy. The only way to make sure we have the same list is by iterating over the policy IDs, which we would anyway in the policy handler.

_But_ this is a very unusual case that is extremely unlikely to happen in a real environment. Also, this was an early "fix". Re-running the policy manager on disconnect actually re-creates this monitoring goroutine, so this problem doesn't happen anymore.

I will add this check back since it's useful to avoid unnecessary work.

policy/manager.go
lgfa29: TIL

But the problem is when the error reaches the policy manager it has been transformed into a generic `error`. We could start considering using [`error.Wrap`](https://godoc.org/github.com/pkg/errors#Wrap) instead of creating new error objects

policy/handler.go
lgfa29: Yeah, it might be worth to even add this to `hclog`. I will keep it here for now since policy validation is the only place we use `multierror` so far.

PR #137, opened 2020-05-20
Limit factor precision in target-value plugin

### LINKED ISSUE #132: Reduce precision when using the target-value strategy
Currently the `target-strategy` plugin calculates a factor of how much the target count value should change and then makes a decision based on a strict comparison against `1`.

This strict comparison can cause changes to count even with a slight variation in the metric being tracked:

```
2020-05-14T12:35:27.860-0400 [TRACE] internal_plugin.target-value: calculated scaling strategy results: policy_id=b87a3b1c-3d0d-9ad4-e514-90b5421897b7 current_count=8 new_count=9 metric_value=5.000000000000001 factor=1.0000000000000002
```

A better approach would be to consider a change if the factor deviates from 1 ± ε, with ε being configurable, but defaulting to a reasonable value.

lgfa29:
This PR introduces the `precision` key for a `RunRequest.Config` for the `target-value` plugin.  `precision` limits how big of a change in the input metric should be considered significant to trigger a scaling event. It also includes small refactorings to improve the `target-value` plugin code.

I set the default precision as `0.01` as a guesstimate. It means the metric needs to change more that 1% of the target value in order for a scale up action to happen. Scaling down is a little higher because we use `math.Ceil` when calculating the next count, so `precision` doesn't affect it as much.

Closes #132

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ IMPROVEMENTS:
  * cli: always use cli library exit code when exiting main function [[GH-130](https://github.com/hashicorp/nomad-autoscaler/pull/130)]
  * core: update Nomad API dependency to 0.11.2 [[GH-128](https://github.com/hashicorp/nomad-autoscaler/pull/128)]
  * plugins/prometheus: use the logger rather than fmt.Print to output Prometheus query warnings [[GH-107](https://github.com/hashicorp/nomad-autoscaler/pull/107)]
+ * plugins/target-value: add new policy configuration `precision` [[GH-132](https://github.com/hashicorp/nomad-autoscaler/issues/132)]
 
 BUG FIXES:
  * agent: fix issue where Nomad Autoscaler would fail to re-connect to Nomad [[GH-119](https://github.com/hashicorp/nomad-autoscaler/issues/119)]
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -87,7 +87,7 @@ policy {
 
     config = {
       target    = 20
-      precision = 0.0001
+      threshold = 0.0001
     }
   }
   ...
@@ -96,4 +96,4 @@ policy {
 ##### Policy configuration
 
 * `target` `(float: <required>)` - Specifies the metric value the Autscaler should try to meet.
-* `precision` `(float: 0.01)` - Specifies how significant a change in the input metric should be considered. Small precision values can lead to output fluctuation.
+* `threshold` `(float: 0.01)` - Specifies how significant a change in the input metric should be considered. Small threshold values can lead to output fluctuation.
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -86,7 +86,7 @@ policy {
     name = "target-value"
 
     config = {
-      target = 20
+      target    = 20
       precision = 0.0001
     }
   }
--- a/docs/plugins/README.md
+++ b/docs/plugins/README.md
@@ -77,7 +77,7 @@ next_count = current_count * (metric_value / target)
 ```
 
 ##### Sample policy
-An example would be attempting to keep the number of active connections to a web server at 20 per instance of the application.  
+An example would be attempting to keep the number of active connections to a web server at 20 per instance of the application.
 
 ```hcl
 policy {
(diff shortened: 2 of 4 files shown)


docs/plugins/README.md
```diff
@@ -87,7 +87,13 @@ policy {
 
     config = {
       target = 20
```
jrasell, line 89: nit: hcl formatting aligns the equal to that of the new line below

plugins/builtin/strategy/target-value/plugin/plugin.go
```diff
@@ -43,6 +51,27 @@ type StrategyPlugin struct {
...
 }
 
+// scaleDirection is used to indicate if the resulting count should increase
+// (scale up), descrease (scale down), or stay the same (none).
+type scaleDirection int8
+
+const (
+	scaleDirectionUp scaleDirection = iota
+	scaleDirectionDown
+	scaleDirectionNone
```
cgbaker, line 61: any particular reason that "up" is the zero?
i don't think it causes any problems, but it will probably keep me up at night.
https://play.golang.org/p/Gk2fZilqlLI

plugins/builtin/strategy/target-value/plugin/plugin.go
```diff
@@ -15,6 +15,14 @@ const (
 	// pluginName is the unique name of the this plugin amongst strategy
 	// plugins.
 	pluginName = "target-value"
+
+	// These are the keys read from the RunRequest.Config map.
+	runConfigKeyTarget    = "target"
+	runConfigKeyPrecision = "precision"
```
cgbaker, line 21: bike-shedding: i'm not a gigantic fan of `precision`... i think `threshold` might be more explicit: we require the factor to exceed the threshold. `precision` suggests to me that we're just being sloppy with our arithmetic.

plugins/builtin/strategy/target-value/plugin/plugin.go
```diff
@@ -131,23 +172,22 @@ func (s *StrategyPlugin) Run(req strategy.RunRequest) (strategy.RunResponse, err
...
+func (s *StrategyPlugin) calculateDirection(count int64, factor, e float64) scaleDirection {
 	switch count {
 	case 0:
 		if factor > 0 {
-			return "up"
-		} else {
-			return "down"
+			return scaleDirectionUp
 		}
+		return scaleDirectionDown
```
cgbaker, line 183: why is this `scaleDirectionDown` and not `scaleDirectionNone`? if it is unreachable code, we should add a comment (and use an appropriate "zero" value)

plugins/builtin/strategy/target-value/plugin/plugin.go
lgfa29: Yeah, that's not a good name. I started out thinking about in terms of significant digits, but that was kind of hard to reason about and slightly more complicated to calculate, so I change to something like an error interval, but calling it `error` sounded like a bad idea. Threshold makes sense.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ IMPROVEMENTS:
  * cli: always use cli library exit code when exiting main function [[GH-130](https://github.com/hashicorp/nomad-autoscaler/pull/130)]
  * core: update Nomad API dependency to 0.11.2 [[GH-128](https://github.com/hashicorp/nomad-autoscaler/pull/128)]
  * plugin: use the logger rather than fmt.Print to output Prometheus query warnings [[GH-107](https://github.com/hashicorp/nomad-autoscaler/pull/107)]
+ * plugins/target-value: add new policy configuration `precision` [[GH-132](https://github.com/hashicorp/nomad-autoscaler/issues/132)]
```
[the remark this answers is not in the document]
lgfa29: I didn't notice there were other changelog entries for plugin related things. I will update them in `master`.

plugins/builtin/strategy/target-value/plugin/plugin.go
lgfa29: Yeah, it seems like this should return `scaleDirectionNone`. It doesn't make sense to scale down if `count` is `0`.

@jrasell any thoughts?

PR #141, opened 2020-05-26
Add the Autoscaler logs as annotations in grafana

attachmentgenie:
Adding promtail, loki and grafana annotations to be able to correlate action and reaction of the Autoscaler.

Now that loki no longer needs a full database to keep track of indexes (1.5.0 release) we can easily add the Autoscaler log entries as annotations to the grafana dashboard so that we can show that there is not magic going on but we can clearly correlate the actions taken by the Autoscaler and the reactions of the platorm

the updated dashboard artifact points into my branch to be able to easily test the changes, this would obviously needs to be changed if and when merged

### DIFF
--- a/demo/vagrant/files/dashboard.json
+++ b/demo/vagrant/files/dashboard.json
@@ -4,11 +4,15 @@
       {
         "$$hashKey": "object:103",
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": "Loki",
         "enable": true,
+        "expr": "{job=\"autoscaler\"} |= \"submitted scaling action\"",
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
-        "name": "Annotations & Alerts",
+        "limit": 100,
+        "name": "Scaling Actions",
+        "showIn": 0,
+        "target": {},
         "type": "dashboard"
       }
     ]
@@ -686,7 +690,7 @@
       "fillGradient": 0,
       "gridPos": {
         "h": 5,
-        "w": 4,
+        "w": 6,
         "x": 0,
         "y": 22
       },
@@ -784,8 +788,8 @@
       "fillGradient": 0,
       "gridPos": {
         "h": 5,
-        "w": 4,
-        "x": 4,
+        "w": 6,
+        "x": 6,
         "y": 22
       },
       "hiddenSeries": false,
@@ -881,8 +885,8 @@
       "fillGradient": 0,
       "gridPos": {
         "h": 5,
-        "w": 4,
-        "x": 8,
+        "w": 6,
+        "x": 12,
         "y": 22
       },
       "hiddenSeries": false,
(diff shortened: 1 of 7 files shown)


demo/vagrant/jobs/loki.nomad
```diff
@@ -0,0 +1,101 @@
...
+
+      resources {
+        cpu    = 100
+        memory = 256
+
+        network {
+          mbits = 10
+
+          port "loki_port" {
+            static = 3100
```
lgfa29, line 83: I am not too familiar with Loki. Does this port provide any useful information for users, like an API or CLI? 

If it does, it would be nice to have it exposed in the `Vagrantfile`, otherwise we could use a dynamic port rather than `static`.

demo/vagrant/files/dashboard.json
```diff
@@ -4,11 +4,15 @@
       {
         "$$hashKey": "object:103",
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": "Loki",
         "enable": true,
+        "expr": "{job=\"autoscaler\"} |= \"submitted scaling action\"",
```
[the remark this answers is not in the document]
attachmentgenie: nice, makes a lot of sense i have adapted the query and now shows the more verbose info

demo/vagrant/jobs/loki.nomad
```diff
@@ -0,0 +1,101 @@
...
+      driver = "docker"
+
+      config {
+        image = "grafana/loki:1.5.0"
+
+        args = [
+          "--config.file=/etc/loki/config/loki.yml",
+        ]
+
+        network_mode = "host"
```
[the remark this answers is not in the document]
attachmentgenie: nope doesnt need it, i have now changed it to a dynamic setup with host network

demo/vagrant/jobs/loki.nomad
attachmentgenie: the host network and static port was an artifact of previous work where i posed myself the same question.

Both loki and promtail do expose a http endpoint [1][2] that exposes an api and health endpoints for loki and a simple dashboard (shows what files it will scrape) and 2 health points for promtail. 

IMHO exposing those would have distracted from the story the demo wants to tell, so i opted not the expose but forgot to cleanup after myself. If you feel you want to expose those (and then promtail makes most sense) i am happy to do the work and expose them.

[1] https://github.com/grafana/loki/blob/master/docs/api.md
[2]https://github.com/grafana/loki/blob/master/docs/clients/promtail/README.md#api

attachmentgenie:
> We noticed that loading the dashboard from a GitHub URL is kind of bad because it makes it hard to work on PRs that make changes to it, so I switched to loading the dashboard config from a volume. Could you rebase your PR to the new `master`?
> 
> Sorry for this, let me know if you need any help.

not a problem i was thinking about creating that PR as well ;-)

demo/vagrant/jobs/loki.nomad
lgfa29: Thanks for the extra info. I agree that we don't need these endpoints for the demo 👍

PR #148, opened 2020-06-01
Fix reason stack to avoid overwriting previous reason

lgfa29:
The current implementation of `pushReason` updates the action before pushing the action's reason into the stack. This causes the previous reason to be lost.

image

In the screenshot above, the event at the top has the right stack, while the event at the bottom has the current, wrong stack logic.

### DIFF
--- a/plugins/strategy/action.go
+++ b/plugins/strategy/action.go
@@ -80,6 +80,9 @@ func (a *Action) pushReason(r string) {
 	}
 
 	// Append current reason to history and update action.
-	a.Meta[metaKeyReasonHistory] = append(history, a.Reason)
+	if a.Reason != "" {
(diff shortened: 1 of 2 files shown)


lgfa29:
> It's just a 1 line change, how bad could it be.

\- Developer who didn't run tests before opening the PR

lgfa29:
Fixing the tests opened a small can of worms. We've previously discussed having the `reason_history` containing all reasons ever attached to an action. 

This is currently not attainable because `Reason` is an exported field and `pushReason` is not, so there's not way for us to enforce the history to be correct since it can be modified directly. Adding methods to set and get `Reason` would fix this, but it would also make it more cumbersome.

I will keep things as they are for now, but keep thinking on how this can be improved.

PR #150, opened 2020-06-04
plugin/target: allow operators to limit the amount of change.

### LINKED ISSUE #149: plugin/strategy: allow setting change limit per evaluation
As an operator, I want to limit the amount of change the autoscaler can make in a single evaluation by configuring the scaling policy with a limit.

The strategy plugin could be updated to accept a config param through the scaling policy which limits the change a single strategy evaluation can make. This would be an integer and would work like follows:

- operator configures change limit value of 2
- strategy calculation wants to scale in/out by 3
- config parameter is adhered to so strategy decides to scale in/out by 2

jrasell:
Currently there is no method to limit the amount of change made in
a single evaluation. The only enforcements made are to ensure the
lower and upper limits are never breached. This change introduces
an optional target config parameter "max_change" which can limit
the maximum change in count actioned during a single evaluation.

Docs and CHANGELOG.md entry to come in follow up PR.

closes #149

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad-autoscaler/agent/config"
@@ -248,8 +249,14 @@ func (a *Agent) handlePolicy(p *policy.Policy) {
 		// plugins doing this.
 		action.Canonicalize()
 
-		// Make sure new count value is within [min, max] limits
-		action.CapCount(p.Min, p.Max)
+		// Run the limit enforcement on the action to keep changes within all
+		// applicable bounds. An error received here means we are unable to
+		// continue as we my make changes which are undesirable for the
+		// operator.
+		if err := a.enforceActionLimits(p, &action, currentStatus); err != nil {
+			actionLogger.Error("failed to enforce policy limits", "error", err)
+			continue
+		}
 
 		// If the policy is configured with dry-run:true then we set the
 		// action count to nil so its no-nop. This allows us to still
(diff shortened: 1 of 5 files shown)


plugins/strategy/action.go
```diff
@@ -62,13 +58,59 @@ func (a *Action) CapCount(min, max int64) {
...
+	// perform our calculations.
+	if a.Count < currentCount {
+		if change := currentCount - a.Count; change > maxChange {
+			newCount = currentCount - maxChange
+		}
+	} else if a.Count > currentCount {
+		if change := a.Count - currentCount; change > maxChange {
+			newCount = currentCount + maxChange
+		}
+	}
```
cgbaker, line 82:
```suggestion
	if a.Count < currentCount - maxChange {
		newCount = currentCount - maxChange
	} else if a.Count > currentCount + maxChange {
		newCount = currentCount + maxChange
	}
```
this is a little simpler, although the original approach (which computes and compares changes) has less risk of overflow as `abs(count)` approaches `2^31`

cgbaker: note, numerical aspects aside, this is equivalent to the previous: `a.Count < currentCount - maxChange` can only happen if `a.Count < currentCount`. and the same for the other case, mutatis mutandis.

agent/agent.go
```diff
@@ -285,3 +292,27 @@ func (a *Agent) handlePolicy(p *policy.Policy) {
...
+		return nil
+	}
+
+	// Make sure new count value is within [min, max] limits.
+	action.CapCount(p.Min, p.Max)
+
+	// If the operator has configured a max_change value, attempt to enforce
+	// this.
+	if maxChange, ok := p.Target.Config[targetpkg.ConfigKeyMaxChange]; ok {
+		maxChangeInt, err := strconv.ParseInt(maxChange, 10, 64)
```
cgbaker, line 311: expanding this parsing to recognize percentages could allow users to specify max changes as a relative value (some percentage of the current value). i think this config could only work when current count is non-zero.

for example, 
```
maxChangeInt = float(maxChangePct) / 100.0 * status.Count
```

jrasell:
This PR has become stale therefore I am closing in preference for revisiting at a later date.

PR #161, opened 2020-06-08
helper: add scaleutils to provide cluster scale in helper tasks.

### LINKED ISSUE #153: cluster-scaling: create Nomad scaleutil functionality
In order to make the development of infrastructure target providers, the Nomad Autoscaler should expose helper functions which perform the required tasks when scaling in a Nomad clusters clients. These functions should include to a basic level:
- identification of nodes to remove from the cluster
- draining of nodes
- monitoring and confirmation of node drain
- translation of Nomad nodeID to remote provider ID
- node purge from Nomad cluster

The functionality should be presented in a way that is easy to consume from within a target plugin.

Stretch goal: ideally the method which identifies nodes in pool should be flexible enough to be extended in the future to include methods other than client class. Additionally the strategy used to decide which nodes are selected for removal should also attempt to be flexible. This however does not block progress of this issue, and is fine to be pushed back for future work.

jrasell:
When performing a scale in action on a pool of Nomad nodes, there
are a number of tasks which should be conducted before the
instances are terminated within the remote infrastructure provider.

The scaleutils helper package provides a method for performing
Nomad node identification, node draining, node drain monitoring
and nodeID translation. This is exposed as a single function so
that target plugin authors can consume the functionality in a
simple manner.

closes #153

### DIFF
--- a/helper/scaleutils/filter.go
+++ b/helper/scaleutils/filter.go
@@ -1,0 +1,99 @@
+package scaleutils
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/nomad/api"
+)
+
+// NodeIDMap provides a mapping between the Nomad ID of a node and its remote
+// infrastructure provider specific ID.
+type NodeIDMap struct {
+	NomadID, RemoteID string
+}
+
+// IdentifierKey is the identifier to group nodes into a pool of resource and
+// thus forms the scalable object.
+type IdentifierKey string
+
+// IdentifierKeyClass uses the Node.Class field to identify nodes into pools of
+// resource. This is the default.
+const IdentifierKeyClass IdentifierKey = "class"
+
+// RemoteProvider is infrastructure provider which hosts and therefore manages
+// the Nomad client instances. This is used to understand how to translate the
+// Nomad NodeID to an ID that the provider understands.
+type RemoteProvider string
+
+// RemoteProviderAWS is the Amazon Web Services remote provider. This provider
+// will use the node attribute as defined by nodeAttrAWSInstanceID to perform
+// ID translation.
+const RemoteProviderAWS RemoteProvider = "aws"
+
+// NodeIDStrategy is the strategy used to identify nodes for removal as part of
+// scaling in.
+type NodeIDStrategy string
+
+// IDStrategyNewestCreateIndex uses the Nomad Nodes().List() output in the
+// order it is presented. This means we do not need additional sorting and thus
+// it is fastest. In an environment that uses bin-packing this may also be
+// preferable as nodes with older create indexes are expected to be most
+// packed.
+const IDStrategyNewestCreateIndex NodeIDStrategy = "newest_create_index"
+
+// nodeAttrAWSInstanceID is the node attribute to use when identifying the
+// AWS instanceID of a node.
+const nodeAttrAWSInstanceID = "unique.platform.aws.instance-id"
+
+// defaultClassIdentifier is the class used for nodes which have an empty class
+// parameter when using the IdentifierKeyClass.
+const defaultClassIdentifier = "autoscaler-default-pool"
+
+// filterByClass returns a filtered list of nodes which are active in the
+// cluster and where the specified class matches that of the nodes.
+func filterByClass(n []*api.NodeListStub, id string) []*api.NodeListStub {
+
+	// Create our output list object.
+	var out []*api.NodeListStub
+
+	for _, node := range n {
+
+		// Ignore nodes that are not in a ready state.
+		if node.Status != api.NodeStatusReady {
+			continue
+		}
+
+		// Ignore nodes that are not currently eligible.
+		if node.SchedulingEligibility != api.NodeSchedulingEligible {
+			continue
+		}
+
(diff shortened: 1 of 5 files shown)


helper/scaleutils/nomad.go
```diff
@@ -0,0 +1,260 @@
...
+		return nil, err
+	}
+
+	return nodeIDMap, nil
+}
+
+// identifyTargets filters the current Nomad cluster node list and then sorts
+// and selects nodes for removal based on the specified strategy. It is
+// possible the list does not contain any many nodes as requested do to the
+// limited number available after filtering.
```
cgbaker, line 80:
```suggestion
// possible the list does not contain as many nodes as requested.
// in this case, do the limited number available after filtering.
```

helper/scaleutils/nomad.go
```diff
@@ -0,0 +1,260 @@
...
+
+		// Assign our node to a local variable as we are launching a go routine
+		// from within a for loop.
+		n := node
+
+		// Launch a routine to drain the node. Append any error returned to the
+		// error.
+		go func() {
+			if err := si.drainNode(ctx, &wg, n.NomadID, &drainSpec); err != nil {
+				result = multierror.Append(result, err)
```
cgbaker, line 207: i don't believe that `multierror.Append` is thread-safe, may want a lock here

helper/scaleutils/filter.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/nomad/api"
+)
+
+// NodeIDMap provides a mapping between the Nomad ID of a node and its remote
+// infrastructure provider specific ID.
+type NodeIDMap struct {
```
lgfa29, line 11: `NodeIDMap` sounds like it would be a `map`, which is not the case. Maybe just `NodeID` since it would scoped as `scaleutils.NodeID` outside the package.

helper/scaleutils/nomad.go
```diff
@@ -0,0 +1,260 @@
...
+		si.log.Debug("identified remote provider ID for node",
+			"node_id", nodeInfo.ID, "remote_id", id)
+
+		out = append(out, NodeIDMap{NomadID: node.ID, RemoteID: id})
+	}
+
+	return out, mErr.ErrorOrNil()
+}
+
+// drainNodes iterates the provides nodeID list and performs a drain on each
```
lgfa29, line 179:
```suggestion
// drainNodes iterates the provided nodeID list and performs a drain on each
```

helper/scaleutils/nomad.go
```diff
@@ -0,0 +1,260 @@
...
+		wg.Add(1)
+
+		// Assign our node to a local variable as we are launching a go routine
+		// from within a for loop.
+		n := node
+
+		// Launch a routine to drain the node. Append any error returned to the
+		// error.
+		go func() {
+			if err := si.drainNode(ctx, &wg, n.NomadID, &drainSpec); err != nil {
```
lgfa29, line 206: You could call `defer wg.Done()` inside the goroutine so you don't have to pass it over to the `drainNode` method.

helper/scaleutils/nomad.go
```diff
@@ -0,0 +1,260 @@
...
+	var idFunc nodeIDMapFunc
+
+	// Depending on the remote provider we are targeting, identify the function
+	// we need to use.
+	switch remoteProvider {
+	case RemoteProviderAWS:
+		idFunc = awsNodeIDMap
+	default:
+		return nil, fmt.Errorf("unsupported remote provider: %q", remoteProvider)
+	}
```
lgfa29, line 145: Since there seems to be a 1:1 relationship between provider and ID function, we can avoid having a `switch` here by using a `map[RemoteProvider]nodeIDMapFunc`, so this block would be something like:

```go
idFunc, ok := idFuncMap[remoteProvider]
if !ok {
  return nil, ...
}
```

and we would avoid extra `case`s when adding other cloud providers.

helper/scaleutils/nomad.go
jrasell: TIL thanks.

helper/scaleutils/nomad.go
jrasell: Yeah thats a much nicer idea!

helper/scaleutils/request.go
```diff
@@ -0,0 +1,52 @@
...
+
+	multierror "github.com/hashicorp/go-multierror"
+)
+
+// ScaleInReq represents an individual cluster scaling request and encompasses
+// all the information needed to perform the pre-termination tasks.
+type ScaleInReq struct {
+
+	// Num is the number of nodes we should select and prepare for termination.
+	Num int
```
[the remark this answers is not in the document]
jrasell: I don't have strong opinions but I kept it different on purpose as we treat count as an end state, this resource must be running this number. Whereas this request is defining a number of resources to be removed.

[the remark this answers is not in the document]
lgfa29: Ah, so they represent different things, makes sense to keep different 👍

Issue #175, opened 2020-06-17
Update demo to fit new policy structure


lgfa29:
The policy structure has changed and now allows multiple checks per policy.

The demo should be update to support this new format and potentially make use of this new capability.

Issue #174, opened 2020-06-17
Block policy handler until the latest evaluation is complete


lgfa29:
After a policy is sent for evaluation, it is assigned to a policy worker for evaluation and execution. If the scaling execution process takes longer than the policy's evaluation interval, it could cause multiple scaling actions to happen in parallel over the same target.

lgfa29:
The new policy eval broker (#282) handles dedup of scaling requests.

Issue #63, opened 2020-03-30
Improve control over policy evaluation


lgfa29:
Currently the agent will evaluate all registered policies in a set interval period. This could cause network and system overload as the number of policies registered increase.

A better approach is to have a queue of policies that need to be evaluated and a configurable number of workers dequeuing  them for evaluation. The number of workers can be defined as an agent configuration field.

PR #176, opened 2020-06-17
Allow multiple checks per policy

### LINKED ISSUE #158: agent: enhance policy evaluation to handle multiple checks per policy
Cluster scaling policies will allow operators to configure multiple checks per policy. This means the agent eval functionality needs to be updated to handle this, as well as perform reconciliation in the event multiple checks provide different suggestions on what scaling action should take place.

In the event two checks suggest different directions the following logic should be applied:
- ScaleOut and ScaleIn = ScaleOut
- ScaleOut and ScaleNone = ScaleOut
- ScaleIn and ScaleNone = ScaleNone

In the event two checks suggest the same direction, but by different values the following logic should be applied:
- ScaleOut(10) and ScaleOut(9) = ScaleOut(10)
- ScaleIn(3) and ScaleIn(4) = ScaleIn(4)

lgfa29:
This PR adds the capability of defining multiple `check`s per policy. This transforms the Autoscaler from a single input/single output (SISO) system into a multiple input/single output (MISO).

There are two new concepts introduced: a worker and check handler. A worker receives a policy for evaluation and executes a scaling action, if necessary. It uses one or more check handler (depending on how many checks are defined in the policy) to do the bulk of the work.

Most of the logic for these new components was extracted from `handlePolicy()` in `agent/agent.go`. This paves the way for a future pool of workers as described in #63.

The vast majority of line changes are in ` policy/nomad/test-fixtures` and can be ignored. This directory contain HCL source and JSON golden files used to facilitate testing Nomad scaling policy parsing and validation. [`policy/nomad/test-fixtures/README.md`](https://github.com/hashicorp/nomad-autoscaler/blob/gh-158/policy/nomad/test-fixtures/README.md) describes them in more detail.

Since workers are executed in a `goroutine` it's possible to hit a problem where a policy is being evaluated multiple times simultaneously. This will be addressed in a future PR and is being tracked at #174.

The demo should also be updated once we have a binary with this changes. This is being tracked at #175.

This PR also includes work done in #170 and so should be merged after that.

Closes #158.


plugins/strategy/action_test.go
```diff
@@ -199,3 +217,194 @@ func TestAction_pushReason(t *testing.T) {
...
+			name: "none vs down",
+			a: &Action{
+				Direction: ScaleDirectionNone,
+			},
+			b: &Action{
+				Direction: ScaleDirectionDown,
+			},
+			expected: &Action{
+				Direction: ScaleDirectionDown,
+			},
```
jrasell, line 250: In a none vs. down I think the none should win as none ensures greater resource availability. Thoughts?

plugins/strategy/action.go
```diff
@@ -86,3 +125,40 @@ func (a *Action) pushReason(r string) {
...
+func PreemptAction(a *Action, b *Action) *Action {
+	if a == nil {
+		return b
+	}
+
+	if b == nil {
+		return a
+	}
+
+	if b.Direction > a.Direction {
```
cgbaker, line 150: the direction checks have the assumption that they are working from the same initial count... that's true, correct?

plugins/strategy/action_test.go
lgfa29: Yeah, that is the safe operation. So we will only scale down if all checks return `ScaleDirectionDown`

lgfa29:
> One meta comment and something we could potentially look at in the future if we want; when sending a policy for evaluation, it's goes from the policy manager, to the agent to the policy worker. I am curious if we remove the agent portion in the future and just have it kept within the policy pkg.

That's the work I am planning for #63. My current thinking is to add a policy evaluation broker, similar to Nomad's `EvalBroker` to centralize and coordinate policy evaluations. In this broker we would be able to do things like dedup, prioritization, separate worker pool for different types of scaling (e.g., more cluster scaling workers since scaling takes longer) etc.

plugins/strategy/action.go
lgfa29: "yes". It's a very soft lower-case yes because it actually it isn't.

The final action is based on two values: current count and metric query result. Since each check has a different query associated we will have multiple results there and so there's a small time difference. 

We could retrieve current count just once, but then the metric is still not guarantee to be read at the time that count is valid.

The ideal scenario would be to pass a timestamp when getting target count and metric, but the Nomad APM doesn't have historical data.

So I tried to bundle these calls as close as possibles, but there isn't any guarantee that all checks are seeing the same data.

Issue #173, opened 2020-06-17
Notify when a disabled policy is detected


lgfa29:
The Autoscaler silently skips policies that are not enabled from being monitored. This can be a source of confusion for operators that may not realize that their policies could be disabled.

Issue #166, opened 2020-06-11
cluster-scaling: support Agent SIGUP to reload policy files


jrasell:
The work being conducted within #157 introduces a new policy file source. In order to support re-reading policies without restarting the agent, it would be beneficial for the agent to handle receiving the SIGHUP. Upon being notified of the SIGHUP syscall, the agent can inform the policy source of the reload request.

PR #178, opened 2020-06-18
policy: add new file source for loading policies from disk.

### LINKED ISSUE #157: policy: handle parsing and evaluation of cluster scaling policies
Cluster scaling policies will have a different format from the current task group policies. The policy parsing will need modifying to handle this work. Once a policy has been parsed, a policy handler should be started to handle evaluations in the current flow.

jrasell:
The MVP cluster scaling includes the ability to load scaling
policies from a specified directory. This functionality is off by
default but is turned on when the policy-dir flag or corresponding
config file parameter is set to a non-empty string.

The file source implements the source interface. It is also
responsible for generating UUIDs for the policies, a task which has
before been performed by Nomad. The source stores the ID mapped to
the file path md5sum. This allows the policy contents to change
without the need to update the policyID.

Funcationality is in place to allow reloading of the policies dir
by sending the agent a SIGHUP signal. This still requires plumbing
in from the agent and will be addressed in a follow up PR.

closes #157


policy/file/source.go
```diff
@@ -0,0 +1,288 @@
...
+			continue
+		}
+
+		// Get the policyID for the file add it to the policy.
+		policyID := s.getFilePolicyID(file)
+		scalingPolicy.ID = policyID.String()
+
+		// Ignore the policy if its disabled. The log line is because I
+		// (jrasell) have spent too much time figuring out why a policy doesn't
+		// get tracked here.
```
lgfa29, line 239: We've all been there: #173 😂

policy/file/source.go
```diff
@@ -0,0 +1,288 @@
...
+	reloadChan chan bool
+
+	// policyReloadChan is used internally to trigger a reload of individual
+	// policies from disk.
+	policyReloadChan chan bool
+
+	// idMap stores a mapping between between the md5sum of the file path and
+	// the associated policyID. This allows us to keep a consistent PolicyID in
+	// the event of policy changes.
+	idMap     map[[16]byte]policy.PolicyID
```
lgfa29, line 40: `[16]byte` could be a type so it's more clear what it represents (the hashed file path right?)

agent/agent.go
```diff
@@ -73,6 +67,28 @@ func (a *Agent) Run(ctx context.Context) error {
...
+	// Setup our initial default policy source which is Nomad.
+	sources := map[policy.SourceName]policy.Source{
+		policy.SourceNameNomad: nomadPolicy.NewNomadSource(a.logger, a.nomadClient, sourceConfig),
+	}
+
+	// If the operators has configured a scaling policy directory to read from
+	// then setup the file source.
+	if a.config.Policy.Dir != "" {
+		sources[policy.SourceNameFile] = filePolicy.NewFileSource(a.logger, sourceConfig, a.config.Policy.Dir, make(chan bool))
+	}
```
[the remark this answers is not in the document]
jrasell: I don't feel strongly about it, but when I added the config flag within #164 I purposely left it blank so operators had to consciously enable reading policies from disk removing any surprises of policies being read in and acted on with a default setting.

policy/manager.go
```diff
@@ -43,16 +42,18 @@ func NewManager(log hclog.Logger, ps Source, pm *manager.PluginManager) *Manager
 func (m *Manager) Run(ctx context.Context, evalCh chan<- *Evaluation) {
 	defer m.stopHandlers()
 
-	policyIDsCh := make(chan []PolicyID)
-	policyIDsErrCh := make(chan error)
+	policyIDsCh := make(chan IDMessage, 2)
+	policyIDsErrCh := make(chan error, 2)
```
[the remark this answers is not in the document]
jrasell: I thought it useful to allow policies sources to send at least 1 message each before getting blocked on the send.

policy/file/source.go
```diff
@@ -0,0 +1,292 @@
...
+		errCh <- fmt.Errorf("failed to get policy %s", ID)
+	} else {
+		resultCh <- *val.policy
+	}
+	s.policyMapLock.RUnlock()
+
+	// Technically the log message should come further up, but its quite
+	// helpful to have the file path logged with the policyID otherwise there
+	// is no way to understand the ID->File mapping.
+	log := s.log.With("policy_id", ID, "file", val.file)
```
cgbaker, line 135: 👍

policy/file/source.go
```diff
@@ -0,0 +1,292 @@
...
+// identifyPolicyIDs iterates the configured directory, identifying the
+// configured policyIDs. The IDs will be wrapped and sent to the resultCh so
+// the policy manager can do its work.
+func (s *Source) identifyPolicyIDs(resultCh chan<- policy.IDMessage, errCh chan<- error) {
+	ids, err := s.handleDir()
+	if err != nil {
+		errCh <- err
+	}
+
+	// Even is we receive an error we may have IDs to send. Otherwise it may be
```
cgbaker, line 201:
```suggestion
	// Even if we receive an error we may have IDs to send. Otherwise it may be
```

policy/file/source.go
```diff
@@ -0,0 +1,292 @@
...
+	// can handle as scaling policies.
+	files, err := fileHelper.GetFileListFromDir(s.dir, ".hcl", ".json")
+	if err != nil {
+		return nil, fmt.Errorf("failed to list files in directory: %v", err)
+	}
+
+	var (
+		policyIDs []policy.PolicyID
+		mErr      *multierror.Error
+	)
```
cgbaker, line 222: style: 4 lines instead of 2 without the `var` block?

cgbaker, line 222:
```suggestion
	var policyIDs []policy.PolicyID
	var mErr      *multierror.Error
```

policy/manager.go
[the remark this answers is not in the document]
lgfa29: Ah ok. I don't think it would make much difference though, since there's no guarantee the two messages would come from different sources 🤔 

But I think it's fine to leave it like this 👍

policy/file/source.go
```diff
@@ -0,0 +1,292 @@
...
+	config *policy.ConfigDefaults
+	dir    string
+	log    hclog.Logger
+
+	// reloadChan is the channel that the agent sends to in order to trigger a
+	// reload of all file policies.
+	//
+	// TODO(jrasell) the reloadChan is not yet triggered by anything and needs
+	//  to be hooked up via the agent.
+	reloadChan chan bool
```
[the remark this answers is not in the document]
jrasell: This is just a placeholder and will likely change when #166 is addressed.

PR #180, opened 2020-06-18
Make IndexHasChanged more resilient to Raft quirks

jippi:
Per https://www.consul.io/api-docs/features/blocking#implementation-details the index can go backward, or even reset, and thus checking for equality will be more resilient for detecting changes

This will be even more relevant with Nomad supporting snapshot restore


cgbaker:
thanks, @jippi . some of those don't apply here (for example, the way we generate the list index), but this is still a good idea.

PR #181, opened 2020-06-18
policy: fix deadlock in eval worker when handler has exited.

jrasell:
A policy worker handler can exit due to a number of reasons but
the policy handler will always attempt to send to the proceedCh.
In the event where the worker has already exited, there is no
channel reciever which means the send to the channel blocks
endlessly.

### DIFF
--- a/policy/worker.go
+++ b/policy/worker.go
@@ -101,9 +101,13 @@ func (w *Worker) HandlePolicy(ctx context.Context, p *Policy) {
 	logger.Trace(fmt.Sprintf("check %s selected", winningHandler.check.Name),
 		"direction", winningAction.Direction, "count", winningAction.Count)
 
-	// Unblock winning handler and cancel the others.
+	// Unblock winning handler and cancel the others. The default guards
+	// against the possibility of there being no receiver on the proceedCh.
 	for _, handler := range checks {
-		handler.proceedCh <- handler == winningHandler
+		select {
+		case handler.proceedCh <- handler == winningHandler:
+		default:
+		}
 	}
 
 	// Block until winning handler returns.
@@ -114,6 +118,9 @@ func (w *Worker) HandlePolicy(ctx context.Context, p *Policy) {
 	case r := <-winningHandler.results():
 		if r.err != nil {
 			logger.Error("failed to execute check", "error", r.err, "check", winningHandler.check.Name)
+			return
+		}
+		if r.action == nil {
 			return
 		}
 	}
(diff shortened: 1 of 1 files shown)


lgfa29:
Oh right.

But I think it would be better to have something like this?

```go
// Unblock winning handler and cancel the others.
for _, handler := range checks {
	select {
	case handler.proceedCh <- handler == winningHandler:
	default:
	}
}
```

This way we don't rely on some state from inside the `goroutine`. 

Or maybe use a buffered channel? (sounds like a hack though)

lgfa29:
If the winning handler is the one that fails the worker will still deadlock. So I think the current approach would be better since it could handle this as well.

It just needs an extra check after the loop.

policy/worker.go
```diff
@@ -103,6 +103,12 @@ func (w *Worker) HandlePolicy(ctx context.Context, p *Policy) {
 
 	// Unblock winning handler and cancel the others.
 	for _, handler := range checks {
+
+		// If the handler has exited do not try and send to the proceedCh as
+		// there is not a receiver and we will block the whole execution.
+		if handler.exited {
```
cgbaker, line 109: does this need a lock?
or... select default?
or... close channel? 
or ... nothing? too much caffeine, feel free to ignore me.

jrasell:
@lgfa29 it locks the `Worker.HandlePolicy` function as it sends to an unbuffered channel which doesn't have a receiver. The `checkHandler.start` function can return at numerous points within its flow [here](https://github.com/hashicorp/nomad-autoscaler/blob/master/policy/worker.go#L194-L299) which results in no receiver for the `proceedCh`.

In the event of two checks and the winning check is the first to be iterated [here](https://github.com/hashicorp/nomad-autoscaler/blob/64cc10d74da8114b9ac8715a9147b65e611c7cb6/policy/worker.go#L105), the action will be triggered but the go routine will leak as it will become blocked on the non-winning check. If the winning check is anything but the first check in the iteration, the action will never be executed.

Checking the error would provide safety for check handlers which result in an error agreed. I do not believe the nil check will provide safety across the rest of the cases. [Here](https://github.com/hashicorp/nomad-autoscaler/blob/master/policy/worker.go#L278-L281) for example the function will return, removing the channel receiver but `r.err` is nil and `r.action` is not nil.

I agree, the unbuffered channel idea is a hack and would merely mask the problem. I'll try and think of a better solution, this initial work was just an attempt to unblock the current situation to get other work moving.

jrasell:
@lgfa29 my latest commit/push works and I think incorporates your suggestion along with addressing something else I found right at the end of the flow.

PR #192, opened 2020-06-24
agent: allow capturing of SIGHUP to reload policy sources.

### LINKED ISSUE #166: cluster-scaling: support Agent SIGUP to reload policy files

jrasell:
The file source in particular benefits greatly from being able
to perform reloads. This allows the source to re-read and decode
files which may have changed as well as process new or delete
scaling policies. In order to keep the behaviour as consistent as
possible across all sources the interface has been modified with
this new functionality. The Nomad source noops a reload, but in
the future we may want to use this to reload Nomad client config
params, in particular ACL tokens.

closes #166

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -3,6 +3,9 @@ package agent
 import (
 	"context"
 	"fmt"
+	"os"
+	"os/signal"
+	"syscall"
 
 	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad-autoscaler/agent/config"
@@ -30,8 +33,12 @@ func NewAgent(c *config.Agent, logger hclog.Logger) *Agent {
 	}
 }
 
-func (a *Agent) Run(ctx context.Context) error {
+func (a *Agent) Run() error {
 	defer a.stop()
+
+	// Create context to handle propagation to downstream routines.
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
 
 	// Generate the Nomad client.
 	if err := a.generateNomadClient(); err != nil {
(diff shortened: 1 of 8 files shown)


agent/agent.go
```diff
@@ -157,3 +173,33 @@ func (a *Agent) generateNomadClient() error {
...
+func (a *Agent) handleSignals() {
+
+	signalCh := make(chan os.Signal, 3)
+	signal.Notify(signalCh, os.Interrupt, syscall.SIGTERM, syscall.SIGHUP)
+
+	// Wait to receive a signal. This blocks until we are notified.
+WAIT:
+	sig := <-signalCh
+
+	a.logger.Info("caught signal", "signal", sig.String())
```
cgbaker, line 193: 👍

policy/file/source.go
```diff
@@ -245,13 +265,13 @@ func (s *Source) handleDir() ([]policy.PolicyID, error) {
...
-		// We have had to decode the file, so store the information. This makes
-		// the MonitorPolicy function simpler as we have an easy mapping of the
+		// Store the file>id mapping if it doesn't exist. This makes the
+		// MonitorPolicy function simpler as we have an easy mapping of the
 		// policyID to the file it came from.
 		s.policyMapLock.Lock()
-		s.policyMap[policyID] = &filePolicy{file: file, policy: &scalingPolicy}
+		if _, ok := s.policyMap[policyID]; !ok {
+			s.policyMap[policyID] = &filePolicy{file: file}
+		}
```
cgbaker, line 274: why the `ok` check instead of just overwriting what's there? is there concurrent code which expects the object at `s.policyMap[policyID]` to remain unchanged (i didn't see any)?

jrasell: This part caused by about an hour of head scratching. 

This function gets called during the initial load and then on reloads of the monitor IDs routine. When we are asked to reload, if we overwrite what is stored, when we subsequently trigger reload of the individual policy monitor we can never tell whether the newly read policy differs from the stored policy.

cgbaker: :100: 

should probably add a comment to that effect 😄

policy/file/source.go
```diff
@@ -94,16 +96,10 @@ func (s *Source) MonitorIDs(ctx context.Context, resultCh chan<- policy.IDMessag
...
-			// current mapping data.
-			s.idMapLock.Lock()
-			s.idMap = make(map[pathMD5Sum]policy.PolicyID)
-			s.idMapLock.Unlock()
-
 			s.identifyPolicyIDs(resultCh, errCh)
 
 			// Tell the MonitorPolicy routines to reload their policy.
-			s.policyReloadChan <- true
+			s.policyReloadChan <- struct{}{}
```
lgfa29, line 102: I am probably missing something, but wouldn't you have to send multiple messages in this channel to trigger a reload in all `MonitorPolicy` routines?

jrasell: Ah yeah great shout, not sure how I missed that one!

PR #193, opened 2020-06-24
plugin: allow the AWS ASG plugin to be launched internally

jrasell:


cgbaker:
I started to make a comment to that affect and decided maybe this isn't
worth the scripting. 😀

PR #195, opened 2020-06-26
plugins: allow all plugins to inherit agents Nomad config.

### LINKED ISSUE #189: cluster-scaling: allow cluster target plugins to use Nomad configuration options
The horizontal cluster scaling target plugins require Nomad configuration options so they can correctly use the scaleutils helper package when performing actions. The functionality should work for plugins regardless of whether they are developed by HashiCorp or by an external developer.

The easiest option (most sensible?) would be to have plugins use their config block to configure the client with Nomad namespaced parameters. This would allow operators you use different ACL tokens and policies from the main autoscaler agent process. Another option would be to have the autoscaler agent Nomad config somehow passed to the plugin during the set-config call.

jrasell:
Plugins such as the AWS ASG target plugin setup a Nomad client in
order to provide scaleutil help. If the operator is using a Nomad
cluster which requires non-default config options, it would be
preferable that they do not have to supply them twice, once to the
agent and again to the plugin.

This changes makes it so that plugin inherit the agents Nomad
configuration by default, reducing the overhead on operators. When
merging config options within the plugin config will take
preference. This allows the operator to provide a subset of params
which are considered specific to the plugin such as ACL token, but
inherit the generic options such as address.

The generic plugin config key "nomad_config_inherit" allows
operators to override default behaviour via a boolean value.

closes #189

### DIFF
--- a/agent/plugins.go
+++ b/agent/plugins.go
@@ -1,6 +1,8 @@
 package agent
 
 import (
+	"strconv"
+
 	"github.com/hashicorp/nomad-autoscaler/agent/config"
 	nomadHelper "github.com/hashicorp/nomad-autoscaler/helper/nomad"
 	"github.com/hashicorp/nomad-autoscaler/plugins"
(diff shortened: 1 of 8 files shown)


agent/plugins.go
```diff
@@ -35,17 +37,44 @@ func (a *Agent) setupPluginConfig() map[string][]*config.Plugin {
...
+
+	// Attempt to convert the string. If the operator made an effort to
+	// configure the key but got the value wrong, log the error and do not
+	// perform the merge. The operator can fix the error and we do not make an
+	// assumption.
+	boolVal, err := strconv.ParseBool(val)
+	if err != nil {
+		a.logger.Error("failed to convert config value to bool", "error", err)
+		return
+	}
```
lgfa29, line 76: In the future I think we should probably validate the config file and exit on error. It's easy to miss a log line.

It would also lead a slightly better UX since we could point to the specific block that has the error.

PR #197, opened 2020-06-29
agent: move and fix default eval interval param to policy block.

jrasell:
The addition of a policy config block allows us to move and
evaluation interval parameter to this section. The param is only
used for policies and so makes sense in this section. A result of
this move fixes an incorrect detailing of the CLI flag which
referenced its old name.

### DIFF
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -72,7 +72,7 @@ func TestAgent_Merge(t *testing.T) {
 		Policy: &Policy{
 			Dir:                       "/etc/scaling/policies",
 			DefaultCooldown:           20 * time.Minute,
-			DefaultEvaluationInterval: 10000000000,
+			DefaultEvaluationInterval: 10 * time.Second,
 		},
 		APMs: []*Plugin{
 			{
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -18,7 +18,7 @@ func Test_Default(t *testing.T) {
 	assert.False(t, def.LogJson)
 	assert.Equal(t, def.LogLevel, "info")
 	assert.True(t, strings.HasSuffix(def.PluginDir, "/plugins"))
-	assert.Equal(t, def.DefaultEvaluationInterval, time.Duration(10000000000))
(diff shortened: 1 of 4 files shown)


jrasell:
I should have a first draft of the docs ready tomorrow, so I was planning on just making the change there and then blatting the docs dir once it was live. I'll raise a follow up with the CHANGELOG and docs update tomorrow as its not a time consuming update.

agent/config/config_test.go
```diff
@@ -18,7 +18,7 @@ func Test_Default(t *testing.T) {
 	assert.False(t, def.LogJson)
 	assert.Equal(t, def.LogLevel, "info")
 	assert.True(t, strings.HasSuffix(def.PluginDir, "/plugins"))
-	assert.Equal(t, def.DefaultEvaluationInterval, time.Duration(10000000000))
+	assert.Equal(t, def.Policy.DefaultEvaluationInterval, time.Duration(10000000000))
```
cgbaker, line 21:
```suggestion
	assert.Equal(t, def.Policy.DefaultEvaluationInterval, 10 * time.Second)
```

agent/config/config_test.go
```diff
@@ -72,8 +70,9 @@ func TestAgent_Merge(t *testing.T) {
 			SkipVerify:    true,
 		},
 		Policy: &Policy{
-			Dir:             "/etc/scaling/policies",
-			DefaultCooldown: 20 * time.Minute,
+			Dir:                       "/etc/scaling/policies",
+			DefaultCooldown:           20 * time.Minute,
+			DefaultEvaluationInterval: 10000000000,
```
cgbaker, line 75:
```suggestion
			DefaultEvaluationInterval: 10 * time.Second,
```

agent/config/config_test.go
```diff
@@ -118,8 +116,9 @@ func TestAgent_Merge(t *testing.T) {
...
```
cgbaker, line 121:
```suggestion
			DefaultEvaluationInterval: 10 * time.Second,
```

PR #198, opened 2020-07-01
scaleutils: protect Autoscaler node from scale in.

jrasell:
Until the Nomad Autoscaler has the ability to reconcile actions
which have not reached a terminal state, the scaleutils should
protect the node the Nomad Autoscaler is running on from scalein
actions. This gives better assurance that the autoscaler wont
drain the node its on and then not be able to complete the scaling
action.

Even once the autoscaler has reconcile ability, there are some
interesting dicussions to be had about what logic to apply in the
situation where the autoscaler is running as a Nomad job and
identifies its own node for scaling in. In this case, we know it
will never be fully able to complete the action.

### DIFF
--- a/helper/scaleutils/nomad.go
+++ b/helper/scaleutils/nomad.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
@@ -15,6 +16,13 @@ import (
 type ScaleIn struct {
 	log   hclog.Logger
 	nomad *api.Client
+
+	// curNodeID is the ID of the node that the Nomad autoscaler is curently
+	// running on.
+	//
+	// TODO(jrasell) this should be removed once the cluster targets and core
+	//  autoscaler components are updated to handle reconciliation.
+	curNodeID string
 }
 
 // NewScaleInUtils returns a new ScaleIn implementation which provides helper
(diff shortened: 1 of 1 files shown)


helper/scaleutils/nomad.go
```diff
@@ -250,3 +268,38 @@ func (si *ScaleIn) monitorNodeDrain(ctx context.Context, nodeID string, index ui
...
+func identifyAutoscalerNodeID(client *api.Client) (string, error) {
+
+	envVar := os.Getenv("NOMAD_ALLOC_ID")
+	if envVar == "" {
+		return "", nil
+	}
+
+	allocInfo, _, err := client.Allocations().Info(envVar, nil)
+	if err != nil {
+		return "", fmt.Errorf("failed to call Nomad allocations info: %v", err)
```
cgbaker, line 286: my question is whether this should be best-effort, in which case `identifyAutoscalerNodeID` should never return an error and block a ScaleUtils from being instantiated... or whether this error (which suggests bad config) should be permitted.

i'm leaning towards best-effort, in which case we can just log any errors.

helper/scaleutils/nomad.go
```diff
@@ -250,3 +268,38 @@ func (si *ScaleIn) monitorNodeDrain(ctx context.Context, nodeID string, index ui
...
+	}
+
+	return allocInfo.NodeID, nil
+}
+
+// TODO(jrasell) this should be removed once the cluster targets and core
+//  autoscaler components are updated to handle reconciliation.
+func filterOutNodeID(n []*api.NodeListStub, id string) []*api.NodeListStub {
+
+	var out []*api.NodeListStub
```
cgbaker, line 296:
```suggestion

	if id == "" {
		return n
	}

	var out []*api.NodeListStub
```

helper/scaleutils/nomad.go
jrasell: Yeah that makes sense to me. I'll get that changed.

PR #200, opened 2020-07-01
policy: validate key params of policy struct.

jrasell:
This adds a validate function to the Policy struct which allows
for the validation ensuring key params are not outside of the
expected.

Ideally both policy sources would use the same function, however,
the Nomad policy source performs validation on the Nomad API
policy struct so there is work to do if we wish to make the
change.

### DIFF
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -192,8 +192,12 @@ func (s *Source) handleIndividualPolicyRead(ID policy.PolicyID, path string) (*p
 	newPolicy.ID = ID.String()
 	newPolicy.ApplyDefaults(s.config)
 
+	if err := newPolicy.Validate(); err != nil {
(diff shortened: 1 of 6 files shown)


policy/file/source.go
```diff
@@ -266,8 +270,13 @@ func (s *Source) handleDir() ([]policy.PolicyID, error) {
...
 
+		if err := scalingPolicy.Validate(); err != nil {
+			mErr = multierror.Append(fmt.Errorf("failed to validate file %s: %v", file, err), mErr)
+			continue
+		}
+
 		for _, c := range scalingPolicy.Checks {
-			c.CanonicalizeAPMQuery(scalingPolicy.Target)
+			c.Canonicalize(scalingPolicy.Target)
 		}
```
lgfa29, line 280: We could also have a `func (p *Policy) Canonicalize()` that does this loop to canonicalize the checks.

PR #204, opened 2020-07-06
plugin/manager: ensure plugins are killed  during manager tests.

jrasell:
When performing tests on the plugin manager, the launched plugins
were not being killed meaning the process were left orphaned on
the host machine.


lgfa29:
Go 1.14 added [`t.Cleanup`](https://tip.golang.org/pkg/testing/#T.Cleanup), which seems like a `defer`? 

Our `go.mod` points to 1.13 but we can change that.

PR #238, opened 2020-08-10
agent: add initial telemetry implementation to emit key stats for monitoring

### LINKED ISSUE #104: [feature] emit key metrics for collection and reporting
The core application should emit metrics so operators can easily and flexibly monitor and alert on the autoscalers behaviour and status. These metrics would include core go runtime details such as NumGoroutine and application specific metrics like a scaling events triggered counter.

lib - https://github.com/armon/go-metrics

jrasell:
closes #104


agent/http/server.go
```diff
@@ -0,0 +1,194 @@
...
+const (
+	// healthRoutePattern is the Autoscaler HTTP router pattern which is used
+	// to register the health server endpoint.
+	healthRoutePattern = "/v1/health"
+
+	// metricsRoutePattern is the Autoscaler HTTP router pattern which is used
+	// to register the metrics server endpoint.
+	metricsRoutePattern = "/v1/metrics"
+
+	// healthAliveness is used to define the health of the Autoscaler agent. It
```
cgbaker, line 27: 👍 
i think later on, we'll want a more sophisticated health check for use with consul service discovery, something along the lines of "can I talk to Nomad?"

policy/file/source.go
```diff
@@ -116,7 +116,7 @@ func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq)
 	// handler which starts the evaluation ticker.
 	val, ok := s.policyMap[req.ID]
 	if !ok {
-		req.ErrCh <- fmt.Errorf("failed to get policy %s", req.ID)
+		policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy %s", req.ID), req.ErrCh)
```
cgbaker, line 119: is `string` the best we can do in `HandleSourceError`? is it possible/worthwhile passing `req.ID` in separately for indexing by the metrics destination?

policy/manager.go
```diff
@@ -199,6 +203,25 @@ func (m *Manager) ReloadSources() {
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-t.C:
+			m.lock.RLock()
+			num := len(m.handlers)
+			m.lock.RUnlock()
+			metrics.SetGauge([]string{"policy", "total_num"}, float32(num))
```
cgbaker, line 220: now witness the power of this fully instrumented autoscaler!

agent/telemetry.go
```diff
@@ -0,0 +1,126 @@
...
+	var telConfig *config.Telemetry
+	if cfg == nil {
+		telConfig = &config.Telemetry{}
+	} else {
+		telConfig = cfg
+	}
+
+	metricsConf := metrics.DefaultConfig("nomad-autoscaler")
+	metricsConf.EnableHostname = !telConfig.DisableHostname
+	metricsConf.EnableHostnameLabel = !telConfig.DisableHostname
```
lgfa29, line 32: Are these two supposed to be `telConfig.DisableHostname`?

agent/http/server_test.go
```diff
@@ -0,0 +1,49 @@
...
+			expectedRespCode: 500,
+			expectedRespBody: "random error string",
+			name:             "internal server error",
+		},
+		{
+			inputReq:         httptest.NewRequest("GET", "/v1/health", nil),
+			inputWriter:      httptest.NewRecorder(),
+			inputError:       newCodedError(418, "I'm a teapot"),
+			expectedRespCode: 418,
+			expectedRespBody: "I'm a teapot",
```
lgfa29, line 35: :tea:

policy/manager.go
```diff
@@ -39,9 +40,12 @@ func NewManager(log hclog.Logger, ps map[SourceName]Source, pm *manager.PluginMa
 
 // Run starts the manager and blocks until the context is canceled.
 // Policies that need to be evaluated are sent in the evalCh.
-func (m *Manager) Run(ctx context.Context, evalCh chan<- *Evaluation) {
+func (m *Manager) Run(ctx context.Context, evalCh chan<- *Evaluation, mInt time.Duration) {
```
lgfa29, line 43: Would `mInt` be better as part of the `Manager` `struct`? Maybe inside some `config` field.

agent/telemetry.go
jrasell: Nope, good spot!

policy/manager.go
jrasell: Yeah good shout, that makes more sense.

policy/file/source.go
jrasell: At the moment yes because of the way `ErrCh` handles messages. We can look into modifying the behaviour so that the channel accepts a custom wrapped error but this is well out of scope for this change.

jrasell: It is possible to modify the `HandleSourceError` to accept the ID and message and then form the error message before sending it to the channel. This allows us to add the policyID as a label where available on the metric request. This allows individual errors to be associated to policies which is nice.

There are two different behaviours that use this metric point; errors during listing which don't have policyIDs to associate to the metric, and then errors during policy reading which do. Maybe we need different metrics for each failure area to provide more granular information and easier to understand.

cgbaker: Yeah, that's what I was thinking. Then I read this:
https://prometheus.io/docs/practices/naming/#labels

And now I think I'm changing my mind. I'm not convinced that it's useful in Prometheus to have the counter for "get policy 7 from `nomad`". It should be a temporary situation where we are continuing to load a policy that doesn't exist on the list; and we are logging the detailed error message.

Let's leave this as is for now; we can add finer-grained monitoring if we discover a need.

PR #252, opened 2020-08-31
Datadog plugin build make target

urjitbhatia:
Adding a make target for datadog plugin build


jrasell (approved):
Thanks yet again @urjitbhatia!

PR #258, opened 2020-09-07
scaleutils: add post scale in task to optionally purge nodes.

### LINKED ISSUE #257: scaleutils: add post scale helper to run node purge after scale in
The Nomad [node purge API call](https://www.nomadproject.io/api-docs/nodes#purge-node) was recently [added to the Nomad Go SDK](https://github.com/hashicorp/nomad/pull/8636). This addition allows the Nomad Autoscaler to purge Nomad nodes once they have been scaled in. This keeps the node list tidy particuary in environments which scale on a frequent basis. Without the purge call, the nodes will eventually be removed from the list by the Nomad GC.

The behaviour should probably have a config flag, allowing the operator to decide whether or not scaled nodes are purge from the system.

jrasell:
When performing cluster scaling and removing nodes from the pool
of compute, an operator may optionally wish to have the terminated
node removed from the Nomad node list. This can be performed by
using the node purge API call and is now included as part of the
scaleutils scale in post tasks run.

The default behaviour is that the purge is not run. The operator
can change this by adding a config option "node_purge" to the
target config mapping. The RunPostScaleInTasks func handles
checking the target configuration, so that this work does not need
to be repeated across all cluster scaling plugins.

closes #257

### DIFF
--- a/helper/scaleutils/nomad.go
+++ b/helper/scaleutils/nomad.go
@@ -5,11 +5,13 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"sync"
 	"time"
 
 	hclog "github.com/hashicorp/go-hclog"
 	multierror "github.com/hashicorp/go-multierror"
+	"github.com/hashicorp/nomad-autoscaler/plugins/target"
 	"github.com/hashicorp/nomad/api"
 )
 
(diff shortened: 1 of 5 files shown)


helper/scaleutils/nomad.go
```diff
@@ -83,6 +85,50 @@ func (si *ScaleIn) RunPreScaleInTasks(ctx context.Context, req *ScaleInReq) ([]N
...
+	// Attempt to read of the node purge config parameter. If it has been set
+	// then check its value, otherwise the default stance is that node purging
+	// is disabled.
+	if val, ok := cfg[target.ConfigKeyNodePurge]; ok {
+
+		// Parse the string as a bool. If we get an error return this as the
+		// operator has attempted to configure this value.
+		boolVal, err := strconv.ParseBool(val)
+		if err != nil {
+			return fmt.Errorf("failed to parse %q: %v", target.ConfigKeyNodePurge, err)
```
cgbaker, line 102: my only concern is how this error bubbles up... the alternatives seem to be a) pre-flight to check this config during start-up; b) treat invalid as a quiet(er) `false` (may not be much difference, aside from error metrics). i don't know whether i prefer either of these.

cgbaker (approved):
👍

helper/scaleutils/nomad.go
jrasell: I hadn't considered the impact on the error metrics. I think with this in mind I probably prefer the second option, with a log message indicating the failure to convert and then exiting. The pre-flight check, unless performed when the policy is read (which doesn't seem the correct place) can only happen when the plugin is called to perform the scale in action, at which point we would need to bail on the entire scale event, which to me seems less than ideal for a malformed boolean param controlling the very end of the pipeline. Nomad will also eventually perform this work if skipped anyway, so it's eventually consistent 😄  ?

cgbaker: yeah, failure to purge isn't the end of the world.

PR #263, opened 2020-09-09
strategy: update interface to accept and respond with evaluation.

jrasell:
As part of the next major release, the strategy plugin has been updated so that it fits more into a pipeline. It therefore accepts an evaluation and responds with an evaluation which has the action populated. Whilst making this change it was found that the package layout was prohibitive. After internal discussions we decided as part of this work to move minimal items over to a new sdk package. This package provides an isolated selection of structs, vars and consts for use throughout the project as well as externally.

This should not be merged until after the next release.

Best viewed commit by commit to see the changes.

### DIFF
--- a/sdk/target.go
+++ b/sdk/target.go
@@ -1,0 +1,52 @@
+package sdk
+
+// TargetStatus is the response object when performing the Status call of the
+// target plugin interface. The response details key information about the
+// current state of the target.
+type TargetStatus struct {
+
+	// Ready indicates whether the target is currently in a state where scaling
+	// is permitted.
+	Ready bool
+
+	// Count is the current value of the target and thus performs the current
+	// state basis when performing strategy calculations to identify the
+	// desired state.
+	Count int64
+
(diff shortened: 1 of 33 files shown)


cgbaker (approved):
👍

sdk/target.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// TargetConfigKeyDrainDeadline is the config key which defines the
+	// override value to use when draining a Nomad client during the scale in
+	// action of horizontal cluster scaling.
+	TargetConfigKeyDrainDeadline = "node_drain_deadline"
+
+	// TargetConfigKeyNodePurge is the config key which defines whether or not
+	// Nomad clients are purged from Nomad once they have been terminated
+	// within their provider.
+	TargetConfigKeyNodePurge = "node_purge"
```
lgfa29, line 51: At some point I think we can avoid having all these constants and have proper fields in the policy. We can start with the internal policy representation first.

lgfa29 (approved):

PR #268, opened 2020-09-15
Update APM interface to allow time range queries

lgfa29:
This PR updates the APM interface to allow plugins to query multiple results for a give time window. 

It adds a new optional attribute to `check` called `query_window` that takes a string in the Go `time.Duration` format. This value is used to query the APM from the current time the check is evaluated and back a window period.

Example:

```diff
 check "avg_instance_sessions" {
   source       = "prometheus"
   query        = "avg((haproxy_server_current_sessions{backend=\"http_back\"}) and (haproxy_server_up{backend=\"http_back\"} == 1))"
+  query_window = "5m"
  ...
}

```

It also updates the current builtin plugins. For the target strategy, only the latest value is currently being used.

### DIFF
--- a/plugins/builtin/apm/nomad/plugin/metrics.go
+++ b/plugins/builtin/apm/nomad/plugin/metrics.go
@@ -3,6 +3,8 @@ package plugin
 import (
 	"fmt"
 	"strings"
+
+	"github.com/hashicorp/nomad-autoscaler/sdk"
 )
 
 const (
@@ -27,7 +29,7 @@ const (
 )
 
 // Query satisfies the Query function on the apm.APM interface.
-func (a *APMPlugin) Query(q string) (float64, error) {
+func (a *APMPlugin) Query(q string, _ sdk.TimeRange) (sdk.TimestampedMetrics, error) {
 
 	// Split the input query so we can understand which query type we are
 	// dealing with.
@@ -39,7 +41,7 @@ func (a *APMPlugin) Query(q string) (float64, error) {
 	case QueryTypeNode:
 		return a.queryNodePool(q)
 	default:
-		return 0, fmt.Errorf("unsupported query type %q", querySplit[0])
+		return nil, fmt.Errorf("unsupported query type %q", querySplit[0])
 	}
 }
 
--- a/plugins/builtin/apm/nomad/plugin/metrics.go
+++ b/plugins/builtin/apm/nomad/plugin/metrics.go
@@ -29,10 +29,9 @@ const (
 )
 
 // Query satisfies the Query function on the apm.APM interface.
+// The Nomad Metrics API doesn't provide historical data, so time range
+// for the query is not used.
 func (a *APMPlugin) Query(q string, _ sdk.TimeRange) (sdk.TimestampedMetrics, error) {
-	// The Nomad Metrics API doesn't provide historical data, so time range
-	// for the query is not used.
-
 	// Split the input query so we can understand which query type we are
 	// dealing with.
 	querySplit := strings.Split(q, "_")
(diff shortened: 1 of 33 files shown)


plugins/builtin/apm/nomad/plugin/metrics.go
```diff
@@ -27,7 +29,7 @@ const (
 )
 
 // Query satisfies the Query function on the apm.APM interface.
-func (a *APMPlugin) Query(q string) (float64, error) {
+func (a *APMPlugin) Query(q string, _ sdk.TimeRange) (sdk.TimestampedMetrics, error) {
```
cgbaker, line 32: because nomad doesn't support historical data, right?

lgfa29: Yup. I will highlight that in a comment.

plugins/builtin/apm/nomad/plugin/metrics.go
```diff
@@ -27,7 +29,9 @@ const (
...
+	// The Nomad Metrics API doesn't provide historical data, so time range
+	// for the query is not used.
```
jrasell, line 34: Curiosity question; is there is a trick I am missing here and some IDE magic; how comes the comment regarding the input parameter is inside the function rather than as the function comment?

policy/file/parse_test.go
```diff
@@ -114,6 +115,11 @@ func Test_decodeFile(t *testing.T) {
 			actualError := decodeFile(tc.inputFile, tc.inputPolicy)
 			assert.Equal(t, tc.expectedOutputPolicy, tc.inputPolicy, tc.name)
 			assert.Equal(t, tc.expectedOutputError, actualError, tc.name)
+
+			// Print unexpected errors.
+			if actualError != nil && tc.expectedOutputError == nil {
```
jrasell, line 120: Is this not caught by the assert on line 117?

jrasell (commented):
LGTM! 🥳 

Only comments are questions out of curiosity.

plugins/builtin/apm/nomad/plugin/metrics.go
lgfa29: At first I thought this would be more a developer-focused comment, so I didn't want it to show in the Go docs. 

But looking at it again, it's probably even more relevant to external callers, since this function will always return a single value regardless of the input range.

I will move it outside, thanks for flagging it 👍

lgfa29:
@jippi @urjitbhatia this is the change I [mentioned](https://github.com/hashicorp/nomad-autoscaler/pull/241#issuecomment-680386152) earlier.  I am curious to hear if you have any thoughts or feedback on this. 

Thanks in advance!

urjitbhatia:
Hi @lgfa29 lgtm - one teeny tiny concern is that we are setting the `to` time to `now` I have noticed that sometimes datadog doesn't behave very well if you do up to the second queries but magically works ok if you offset a bit - for example if you do between `5m` and `30sec` ago likely because their api/ingest sometimes falls behind.
Maybe we can cross that bridge when we need to

lgfa29:
> Hi @lgfa29 lgtm - one teeny tiny concern is that we are setting the `to` time to `now` I have noticed that sometimes datadog doesn't behave very well if you do up to the second queries but magically works ok if you offset a bit - for example if you do between `5m` and `30sec` ago likely because their api/ingest sometimes falls behind.
> Maybe we can cross that bridge when we need to

Oh that's really good to know, thank you! Currently I have it so that if `query_window` is not defined it will just return the latest value. But time sync is always problematic, so it would probably be better to have it as a small range (like `30s`?).

PR #271, opened 2020-09-25
switching to focal and upstream package of hey

### LINKED ISSUE #270: demo: hey install binaries are no longer available breaking usage of demos
[rakyll/hey](github.com/rakyll/hey) is used within the autoscaler demos to generate load and connections to scalable example HTTP services. Currently the [download links are broken](https://github.com/rakyll/hey/issues/222) which means the demos do not work as documented.

There are a couple of ways we could potentially solve this:
- remove use of `hey` and identify replacement, although what is to say we don't experience this with a new tool
- compile the binary from source during demo build
- compile the binary from source and host this somewhere we control

attachmentgenie:
Also applying patch for currently borked upstream vagrant box. fixes #270

### DIFF
--- a/demo/vagrant/Vagrantfile
+++ b/demo/vagrant/Vagrantfile
@@ -2,7 +2,7 @@
 # vi: set ft=ruby :
 
 Vagrant.configure("2") do |config|
-  config.vm.box = "ubuntu/bionic64"
+  config.vm.box = "ubuntu/focal64"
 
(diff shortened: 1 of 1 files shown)


jrasell (approved):
Thanks a lot @attachmentgenie. Pulled this change locally and worked perfectly!

lgfa29 (approved):
Thanks!

PR #277, opened 2020-10-02
target/nomad: support for namespaced jobs in scale and status RPC.

### LINKED ISSUE #276: target/nomad: add namespace awareness to status and scale RPCs
In environments that use Nomad namespaces, the jobID is not enough to uniquely identify a job. The namespace along with the jobID forms the unique object. The Nomad target plugin should therefore be updated so that it can handle namespaced jobs.

jrasell:
Tested locally via vagrant demo with success.

closes #276


cgbaker (approved):
👍

lgfa29 (approved):
Nice one!

PR #278, opened 2020-10-02
Add Azure VMSS target plugin

lcheruka:


jrasell:
@lcheruka thanks so much for this! It may take a little time to get this properly tested, but i'll keep you updated as its worked through.

lgfa29 (approved):

lgfa29:
Thank you so much for the contribution @lcheruka!

Sorry it took this long for us to review, but we were quite busy with Nomad 1.0 and we also had to ramp-up our knowledge of Azure in order to properly test this. 

I made small changes to the PR to:

1. automatically load the plugin in the agent
2. fallback to looking for the instance name using node meta tags since Azure fingerprinting is not released in Nomad yet, and this would also allow clusters pre-Nomad 1.0 to use this as well (although they would need to manually add the meta tag to their clients)

I also rebased the branch to pickup and fix the import path for the `helper` module, since we moved to `sdk/helper`.

There's a [WIP PR](https://github.com/hashicorp/nomad-autoscaler/pull/295) that will deploy a demo cluster. It's just pending the new release of the Nomad and Nomad Autoscaler, but if you have any feedback we would appreciate as well.

PR #282, opened 2020-10-09
Evaluation broker

lgfa29:
This PR introduces the evaluation broker, which is responsible for storing, deduping and controlling the distribution and flow of policy evaluation requests to workers.

The eval requests are stored in a set of heaps, sorted by priority and create time (FIFO).

Evals are picked from the broker by workers and they must ACK the evaluation if it completes successful or NACK it otherwise. If an ACK doesn't arrive within the deadline (5 min for now, but should probably be configurable per policy) the eval is considered NACK'd.

A new agent configuration block was also added to control the number of workers for each policy type:

```hcl
policy_workers {
  cluster = 20
  horizontal = 5
}
```


policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+			return
+		}
+	}
+
+	heap.Push(&pending, eval)
+	b.pendingEvals[queue] = pending
+
+	// Unblock any blocked dequeues.
+	select {
+	case b.waiting[queue] <- struct{}{}:
```
cgbaker, line 130: are we sure that `b.waiting[queue]` exists in the map? do we need to check-and-create?
i think it's only there because of a race condition where the workers created it before we made our first call to `Enqueue`

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+
+	// Pop heap and update reference.
+	raw := heap.Pop(&pending)
+	b.pendingEvals[queue] = pending
+
+	return raw.(*sdk.ScalingEvaluation)
+}
+
+// waitForWork blocks until queue receives an item or the context is canceled.
+func (b *Broker) waitForWork(ctx context.Context, queue string) bool {
```
cgbaker, line 202: named return might document the semantics of `true/false`

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+func (p *PendingEvaluations) Pop() interface{} {
+	n := len(*p)
+	e := (*p)[n-1]
+	(*p)[n-1] = nil
+	*p = (*p)[:n-1]
+	return e
+}
+
+// Peek is used to peek at the next element that would be popped
+func (p PendingEvaluations) Peek() *sdk.ScalingEvaluation {
```
cgbaker, line 331: it doesn't appear that we use `Peek()` here

agent/agent.go
```diff
@@ -70,6 +73,10 @@ func (a *Agent) Run() error {
 	policyEvalCh := a.setupPolicyManager()
 	go a.policyManager.Run(ctx, policyEvalCh)
 
+	// Launch eval broker and workers.
+	a.evalBroker = policyeval.NewBroker(a.logger.ResetNamed("policy_eval"), 5*time.Minute, 3)
+	a.initWorkers(ctx)
```
cgbaker, line 78: may want to make this dynamic later, for autoscaling the number of brokers or simply SIGHUP'ing config

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+	logger = logger.With("policy_id", unack.Eval.Policy.ID)
+
+	// Stop the timer, doesn't matter if we've missed it.
+	unack.NackTimer.Stop()
+
+	// Cleanup.
+	delete(b.unack, evalID)
+
+	// Check if we've hit the delivery limit.
+	if dequeues := b.enqueuedEvals[evalID]; dequeues >= b.deliveryLimit {
```
cgbaker, line 280: should the eval be GC'd from `enqueuedEvals` and `enqueuedPolicies`?

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+	// enqueuedEvals and enqueuedPolicies are indexes to quickly find if an eval
+	// or a policy is waiting be evaluated.
+	enqueuedEvals    map[string]int
+	enqueuedPolicies map[string]string
+
+	// unack tracks evaluations that have not been ack'd yet.
+	unack map[string]*unackEval
+
+	// waiting tracks Dequeue requests that are blocked waiting for work.
+	waiting map[string]chan struct{}
```
cgbaker, line 43: it might be nice to document the invariants between these fields
* every named queue should be represented as a key in `pendingEval` and `waiting`
* an eval should either be present in both `enqueuedEvals` and `unack`, or present in `pendingEvals`

cgbaker (approved):
no significant problems. one potential (but unlikely) race condition and some comments about logging.

👍👍👍

policyeval/base_worker.go
```diff
@@ -22,23 +24,75 @@ import (
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			w.logger.Info("stopping worker")
+			return
+		default:
+		}
+
+		eval, token, err := w.broker.Dequeue(ctx, w.queue)
```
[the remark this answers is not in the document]
lgfa29: Yup, it will block until there's work in the queue or return `nil` if `ctx` is closed.

policyeval/base_worker.go
```diff
@@ -22,23 +24,75 @@ import (
...
+			// Notify broker that policy eval was not successful.
+			if err := w.broker.Nack(eval.ID, token); err != nil {
+				logger.Warn("failed to NACK policy evaluation", "err", err)
+			}
+			continue
+		}
+
+		// Notify broker that policy eval was successful.
+		if err := w.broker.Ack(eval.ID, token); err != nil {
+			logger.Warn("failed to ACK policy evaluation", "err", err)
```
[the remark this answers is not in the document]
lgfa29: Yes, if the ACK fails, the broker will re-try this. But if this scaling action succeeded, the next eval will no-op because the target is already scaled. 

Also there aren't many ways the ACK would fail. I copied this from Nomad an there this call happens over RPC, so network is an issue there.

policyeval/broker.go
lgfa29: That's  a good idea and something I struggled with while reading the Nomad code 😅

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+	defer b.l.Unlock()
+	b.enqueueLocked(eval, "")
+}
+
+func (b *Broker) enqueueLocked(eval *sdk.ScalingEvaluation, token string) {
+	logger := b.logger.With(
+		"eval_id", eval.ID, "policy_id", eval.Policy.ID,
+		"queue", eval.Policy.Type, "token", token)
+
+	logger.Debug("enqueue eval")
```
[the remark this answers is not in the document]
lgfa29: This uses the logger defined above, so it logs `eval.ID`, `eval.Policy.ID`, `eval.Policy.Type` and `token`. 

Any extra info missing?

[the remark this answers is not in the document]
cgbaker: nope, that's wonderful. i missed that context. 💯

policyeval/broker.go
```diff
@@ -0,0 +1,347 @@
...
+		if _, ok := b.waiting[queue]; !ok {
+			b.waiting[queue] = make(chan struct{}, 1)
+		}
+	}
+
+	// Check if an eval for the same policy is already enqueued.
+	pendingEvalID, ok := b.enqueuedPolicies[eval.Policy.ID]
+	if !ok {
+		b.enqueuedPolicies[eval.Policy.ID] = eval.ID
+	} else if pendingEvalID != eval.ID {
```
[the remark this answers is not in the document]
lgfa29: If `pendingEvalID == eval.ID` it should actually be caught by the block above:

```go
if _, ok := b.enqueuedEvals[eval.ID]; ok {
	if token == "" {
		logger.Debug("eval already enqueued")
		return
	}
}
```

The cases for submitting the same the same ID twice (that I can think of) are:
* eval got NACK'd
* the policy manager sends it to the broker twice (for some reason....)

In the first case, the `token` is not empty, so the flow will proceed until this line and go through since the eval ID will be the same. But since it will have same `CreateTime` it will return without doing anything (something like this [test case](https://github.com/hashicorp/nomad-autoscaler/pull/282/files#diff-c600eef77639493229b591b692f5d3d7R109-R115)).

The second case will exit early in the block I mentioned earlier since the token will be empty (there's a [test case for this](https://github.com/hashicorp/nomad-autoscaler/pull/282/files#diff-c600eef77639493229b591b692f5d3d7R89-R91)).

policyeval/broker.go
lgfa29: It's created earlier in this function, when we try to find the heap for this queue from `pendingEvals`. 

But maybe it should be more explicit and without the `pendingEvals` guard?

cgbaker: ahh, marvelous. sorry, i put so many comments in here i can't read the code anymore.

policyeval/broker.go
lgfa29: Yup, good catch.

policyeval/base_worker.go
```diff
@@ -22,23 +24,75 @@ import (
...
+		logger:        l.Named("worker").With("id", id, "queue", queue),
 		pluginManager: pm,
 		policyManager: m,
+		broker:        b,
+		queue:         queue,
+	}
+}
+
+func (w *BaseWorker) Run(ctx context.Context) {
+	w.logger.Debug("starting worker")
```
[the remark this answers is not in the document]
lgfa29: It logs with the `id` and `queue`.

policyeval/broker.go
lgfa29: Ah right. In Nomad this is used by workers that can handle multiple queues, but that's not the case for use so I will remove it.

agent/agent.go
lgfa29: Yeah, I was trying to decide between a global config or a per-policy value. 

I think per-policy would be better since it can be configured as needed, but global is appealing because you can just set it to your worst-case value.

So probably both? 😄

agent/config/config.go
```diff
@@ -247,6 +251,16 @@ type Policy struct {
...
 
+// PolicyWorkers holds the configuration for the number of policy evaluation
+// workers for each policy type.
+type PolicyWorkers struct {
+	ClusterPtr *int `hcl:"cluster,optional" json:"-"`
+	Cluster    int
+
+	HorizontalPtr *int `hcl:"horizontal,optional" json:"-"`
+	Horizontal    int
+}
```
jrasell, line 262: Adding as a general comment rather than something I heavily believe in; but we do have the `config.Policy` struct and maybe these could go within it, rather than having a separate configuration block?

policyeval/broker.go
```diff
@@ -0,0 +1,371 @@
...
+//     enqueued.
+//   - the policy ID remains in `enqueuedPolicies` until the eval is ACK'd or
+//     the delivery limit is reached.
+//   - the value for the policy ID is the ID of the eval that was enqueued.
+//   - the value for the policy ID is updated if a newer eval for the policy is
+//     enqueued.
+type Broker struct {
+	logger hclog.Logger
+
+	l sync.RWMutex
```
jrasell, line 48: I think its worth adding a comment to explain which maps this lock is designed to be used for, even if its all for clarity.

sdk/eval.go
```diff
@@ -1,11 +1,19 @@
 package sdk
 
+import (
+	"time"
+
+	"github.com/hashicorp/nomad-autoscaler/helper/uuid"
```
jrasell, line 6: I would expect this to be caught by the `check-sdk` make target which disallows us from importing internal packages into the `sdk` package.

policyeval/broker.go
```diff
@@ -0,0 +1,371 @@
...
+	b.l.Lock()
+	defer b.l.Unlock()
+
+	// Generate an UUID for the token.
+	token := uuid.Generate()
+
+	// Setup Nack timer.
+	// Eval needs to be Ack'd before this timer finishes.
+	nackTimer := time.AfterFunc(b.nackTimeout, func() {
+		b.Nack(eval.ID, token)
```
jrasell, line 192: Do we need to handle the error returned by `Nack` here for at least logging purposes?

jrasell (approved):
minor comments alongside those of @cgbaker but its look very nice! 🥳

policyeval/broker.go
cgbaker: agreed. the nack/ack errors seem to fall into two categories:
* don't bother retrying (eval doesn't exist, token is wrong, already nack'd), but something may be wrong.
* don't need to do anything (in which case, maybe shouldn't have returned an error)

agent/config/config.go
lgfa29: I had it there initially, but it looked kind of off:

```hcl
policy {
  dir                = "/opt/nomad-autoscaler/policies"
  default_cooldown   = "2m"
  cluster_workers    = 8
  horizontal_workers = 4
}
```
or nested:
```hcl
policy {
  dir              = "/opt/nomad-autoscaler/policies"
  default_cooldown = "2m"
  workers {
    cluster    = 8
    horizontal = 4
  }
}
```

To me it doesn't seem like they are related, but I can change it if it's better.

PR #288, opened 2020-10-14
version: allow setting custom metadata on version output.

jrasell:
Brings the autoscaler inline with HashiCorp products and expected
version formatting.


cgbaker (approved):
👍

lgfa29 (approved):

PR #290, opened 2020-10-16
put the policy id in the scaling action metadata

cgbaker:
this supports looking up policy associated with scaling actions


jrasell (approved):
LGTM :shipit:

PR #292, opened 2020-10-19
Fix version geneartion logic with GitDescribe

lgfa29:
When the CI pipeline builds the final binary, `GitDescribe` is define with all segments already, so the CLI shouldn't add any extra info (apart from the commit hash).


cgbaker (approved):
y not

PR #295, opened 2020-10-25
demo: add Azure remote demo

lgfa29:
This PR adds cluster scaling demo on Azure using the a VMSS instance. It follows a new folder structure to better organize files per cloud provider. A following PR will adapt the AWS demo to the same structured.

It's currently a WIP because it requires a new build of the Autoscaler that includes the [Azure VMSS plugin](https://github.com/hashicorp/nomad-autoscaler/pull/278) and a new version of Nomad that includes [Azure node fingerprinting](https://github.com/hashicorp/nomad/pull/8979).

### DIFF
--- a/demo/remote/azure/terraform/control/main.tf
+++ b/demo/remote/azure/terraform/control/main.tf
@@ -1,0 +1,17 @@
+module "my_ip_address" {
+  source = "matti/resource/shell"
+
+  command = "curl https://ipinfo.io/ip"
+}
+
+module "hashistack_cluster" {
+  source = "../modules/azure-hashistack"
+
+  allowlist_ip = ["${module.my_ip_address.stdout}/32"]
+}
+
+module "hashistack_jobs" {
+  source = "../../../terraform/modules/shared-nomad-jobs"
+
+  nomad_addr = module.hashistack_cluster.nomad_addr
+}
--- a/demo/remote/azure/terraform/modules/azure-hashistack/templates/azure_autoscaler.nomad
+++ b/demo/remote/azure/terraform/modules/azure-hashistack/templates/azure_autoscaler.nomad
@@ -1,0 +1,119 @@
+job "autoscaler" {
+  datacenters = ["dc1"]
+
+  group "autoscaler" {
+    count = 1
+
+    task "autoscaler" {
+      driver = "docker"
+
+      config {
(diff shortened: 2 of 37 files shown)


demo/remote/azure/terraform/modules/azure-hashistack/templates/azure_autoscaler.nomad
```diff
@@ -0,0 +1,119 @@
...
+  datacenters = ["dc1"]
+
+  group "autoscaler" {
+    count = 1
+
+    task "autoscaler" {
+      driver = "docker"
+
+      config {
+        image   = "hashicorp/nomad-autoscaler:0.2.0"
```
lgfa29, line 11: TODO: make sure this is the right version.

lgfa29: In order to test this before everything is released, you can get a nightly build from the Autoscaler, place it in `/usr/bin` in the client and use the `exec` driver.

demo/remote/azure/terraform/control/main.tf
```diff
@@ -0,0 +1,17 @@
+module "my_ip_address" {
+  source = "matti/resource/shell"
+
+  command = "curl https://ipinfo.io/ip"
+}
+
+module "hashistack_cluster" {
+  source = "../modules/azure-hashistack"
+
+  allowlist_ip = ["${module.my_ip_address.stdout}/32"]
```
lgfa29, line 10: To test this with Azure fingerprinting, you can build Nomad from `master`, zip it and put somewhere public (like S3). Then add a new line here:

```hcl
nomad_binary = "https://...
```

lgfa29:
@jrasell any ideas on how to make Terraform destroy the jobs first? As it is, Terraform will destroy the VMs first and then try to to destroy the jobs, but it will timeout. 

I tried the new `depends_on` for modules, but Terraform was not too happy about it.

jrasell:
@lgfa29 I would expect the fact the jobs module has a requirement on an output of the cluster module to create the correct dependancy chain. Failing that and providing we use TF 0.13, again I would expect it to create the correct dependancy chain between the two modules. Lets look at this online later, hopefully we can figure out a nice solution other than having to do targeted destroys.

cgbaker (approved):
💯 🏆 🥇

PR #301, opened 2020-10-28
fix: allow empty policy type to keep compatibility with Nomad pre-1.0

lgfa29:


policy/nomad/validate.go
```diff
@@ -46,6 +46,8 @@ func validateScalingPolicyByType(policy *api.ScalingPolicy) error {
 		return validateHorizontalPolicy(policy)
 	case "cluster":
 		return validateClusterPolicy(policy)
+	case "":
+		return nil
```
cgbaker, line 50:
```suggestion
		return validateHorizontalPolicy(policy)
```

cgbaker: why not this? ^

cgbaker (approved):

policy/nomad/validate.go
cgbaker, line 50:
```suggestion
```

PR #302, opened 2020-10-29
Use Dockerhub Mirror.

RussellRollins:
Dockerhub is going to rate limit unauthenticated pulls.

Use our internal mirror for builds running through CircleCI.


cgbaker (approved):
thanks! 👍

PR #305, opened 2020-10-30
demo/das: update limits and fix nginx config file reload

lgfa29:
The nginx task was mounting the config file also as a file inside the Docker container. When Consul Template re-rendered the file, the inode reference would change, so nginx wouldn't see the updated version, even after a `SIGHUP`


angrycub (approved):
:+1:

Issue #281, opened 2020-10-08
demo: add Datadog cluster scaling query example


jrasell:
@nicgrayson reached out on hangops with the following:
> There isn't a datadog example of that query but i think that is a datadog problem not a nomad problem. Is there a place i should PR to add a dd example?

It would be useful to find the correct place to include examples like this to help users get started quicker. Initial thoughts are either the target plugin documentation, or including them within the demo so users can optionally switch.

nicgrayson:
@jrasell got datadog working. I could just comment out an example and put it here: https://github.com/hashicorp/nomad-autoscaler/blob/master/demo/remote/aws_autoscaler.nomad#L70

I could also make a new directory with no promethus in it.

jrasell:
Hi @nicgrayson. I agree with adding the example as a comment in the current jobspec and then maybe a small callout in the README explaining the possibility of using Datadog. It would probably be also worth adding an example of running the DD plugin [here](https://github.com/hashicorp/nomad-autoscaler/blob/master/demo/remote/aws_autoscaler.nomad#L37).

PR #322, opened 2020-12-03
add example of datadog usage to the demo nomad job file

nicgrayson:
closes: #281


lgfa29 (approved):
Thank you for this @nicgrayson!

I just pushed a commit to use the new syntax coming out soon in the Autoscaler v0.2.0 release.

Issue #310, opened 2020-11-05
Nomad APM plugin should support querying percentages based on allocated values


lgfa29:
Currently the Nomad APM plugin reads metrics directly from Nomad. This has the side effect of having all percentage values be relative to the host. It would be more helpful to be able to scale a group based on usage percentage of their allocated values.

lgfa29:
Closed by #324

PR #324, opened 2020-12-09
Use percent of allocated CPU for task group queries against the nomad-apm plugin

tydomitrovich:
### Overview

As per [this issue](https://github.com/hashicorp/nomad-autoscaler/issues/310) it would be useful for the nomad-apm plugin to support querying the percentage of CPU used out of the CPU that is allocated for a task group.

Currently querying the nomad-apm plugin for `avg_cpu` returns the percentage of the host's CPU an allocation uses via the `nomad.client.allocs.cpu.total_percent` metric. This is not ideal since this value can vary across hosts with different hardware configurations.

### Changes

There was not a metric in the Nomad API that returns an allocation's CPU usage out of total allocated CPU so I added the calculation to the nomad-apm plugin.

### Testing

I built a docker image with these changes, deployed it out to one of my Nomad clusters, and verified that scaling worked as expected using the new CPU percentage with the following sample job:

```

job "scaling-test-service" {

  datacenters = ["app"]

  type = "service"

  group "test-service" {

    network {
      port "http" {
        to = 3000
      }
    }

    service {
      name = "scaling-test-service"
      port = "http"
      check {
        name     = "alive"
        type     = "tcp"
        interval = "10s"
        timeout  = "2s"
      }

    }

    task "test-service" {

      driver = "docker"

      config {
        image = "kosprov/http-echo-server:latest"

        ports = ["http"]
      }

      resources {
        cpu    = 500
        memory = 256
      }
    }

    scaling {
      enabled = true
      min = 1
      max = 10
      policy {
        cooldown = "30s"

        check "cpu" {
          source = "nomad-apm"
          query  = "avg_cpu"
          strategy "target-value" {
            target = 80
          }
        }
      }
    }
  }
}
```

Load generated with `hey`.

### Other Notes

Let me know if you would like me to create a new `queryMetric` type in case it is desired to still have `avg_cpu` queries return the value of  `nomad.client.allocs.cpu.total_percent`. 

It might also be nice to have this value available as a metric from the Nomad API. If that approach is desired I can close this PR.

Thanks for looking!

---
This change is [Reviewable](https://reviewable.io/reviews/hashicorp/nomad-autoscaler/324)


tydomitrovich:
Just a heads up, I have signed the CLA but the integration seems stuck in pending.

cgbaker:
hi @tyler-domitrovich , regarding the CLA, there's a link for pending CLA. it doesn't do anything for me, but maybe it will for you?

https://cla.hashicorp.com/check/hashicorp/nomad-autoscaler?pullRequest=324

regarding the PR, my inclination is to add a new query metric... perhaps `cpuAllocated`?

tydomitrovich:
Hi @cgbaker, I was able to get the CLA to show as signed!

In regards to the PR I have gone ahead and added a new type of query metric called `cpu-allocated` as well as the required logic to ensure that metric type can only be used for task group queries. The `cpu` metric now retains its original behavior of returning `nomad.client.allocs.cpu.total_percent`.

GNUmakefile
```diff
@@ -9,7 +9,7 @@ GO_LDFLAGS := "-X github.com/hashicorp/nomad-autoscaler/version.GitCommit=$(GIT_
 .PHONY: tools
 tools: ## Install the tools used to test and build
 	@echo "==> Installing tools..."
-	GO111MODULE=on cd tools && go get -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.24.0
```
tydomitrovich, line 12: I had to remove the update flag since [one of the linters golangci-lint aggregates](https://github.com/dominikh/go-tools/tree/v0.1.0) introduced a breaking change in a minor version release that gets picked up when `-u` is set. This was causing the build to fail.

lgfa29 (approved):
Thank you very much for the contribution @tydomitrovich! 

I tested it locally and all looks good. I also created a follow-up PR on top of this to add the same functionality for memory.

calh:
Hi @lgfa29 and  @tydomitrovich,

My Google searching landed me here, trying to find this feature.  The documentation on the [official site](https://developer.hashicorp.com/nomad/tools/autoscaling/plugins/apm/nomad) does not have the `cpu-allocated` or `memory-allocated`  queries listed.  I had to dig through your patches here to find out how to use this.  Can that be updated?

lgfa29:
Thanks for letting us know @calh. I've documented the missing configuration in https://github.com/hashicorp/nomad/pull/15299

PR #326, opened 2021-01-03
Update example for nomad 1.0

trilom:
Since this example uses the latest nomad it no longer works with 1.0 because of the network stanzas.

I went through and got it working with the following configuration, but have not tested it against a networked cluster to verify it's working as it is intended to on a scale greater than 1.

There might need to be more work done I am working on another project that will ultimately validate this work but is not complete.

Thanks

### DIFF
--- a/demo/vagrant/horizontal-app-scaling/jobs/autoscaler.nomad
+++ b/demo/vagrant/horizontal-app-scaling/jobs/autoscaler.nomad
@@ -4,13 +4,24 @@ job "autoscaler" {
   group "autoscaler" {
     count = 1
 
+    network {
+      port "http" {
+        to = 8080
+        static = 8080
+      }
+      port  "promtail"{
+        to = 9080
+        static = 9080
+      }
+    }
+
     task "autoscaler" {
       driver = "docker"
 
       config {
         image   = "hashicorp/nomad-autoscaler:0.1.1"
         command = "nomad-autoscaler"
-
+        network_mode = "host"
         args = [
           "agent",
           "-config",
@@ -18,10 +29,7 @@ job "autoscaler" {
           "-http-bind-address",
           "0.0.0.0",
         ]
-
-        port_map {
-          http = 8080
-        }
+        ports = ["http"]
       }
 
       ## Alternatively, you could also run the Autoscaler using the exec driver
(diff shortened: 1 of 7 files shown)


demo/vagrant/horizontal-app-scaling/jobs/autoscaler.nomad
```diff
@@ -96,15 +99,12 @@ strategy "target-value" {
 
       config {
         image = "grafana/promtail:1.5.0"
-
+        network_mode = "host"
```
jrasell, line 102: What is the reason the network mode param was added? I don't believe we need it.

demo/vagrant/horizontal-app-scaling/jobs/haproxy.nomad
```diff
@@ -59,6 +72,15 @@ EOF
           interval = "10s"
           timeout  = "2s"
         }
+
+        check {
+          name = "host"
+          type     = "http"
+          address_mode = "driver"
```
jrasell, line 79: I believe we can omit this, and allow the driver to dictate the address_mode for the check.

jrasell (changes requested):
Thanks so much for working on this @trilom.

A meta comment, which duplicates across the job specification files is that additions are not formatted in accordance to HCL formatting and would be great to have fixed. Taking the new http port addition to the `autoscaler.nomad` file, the formatted port addition would look like:
```hcl
port "http" {
  to     = 8080
  static = 8080
}
```

It also seems the new `nomad_autoscaler_telemetry.json` file is unused and a duplicate of the file with the same name under the `shared` directory. Can this addition therefore be removed?

trilom:
Thanks for the feedback, I'll follow up with each one.

As I work through the remote example I realize the authors intent of the shared files more.

For example in the remote demo, the shared files are used exclusively by the azure demo, but ignored by the aws demo.  In this demo the other dashboard is mentioned but ultimately never included anywhere from the shared code.

Once I finish up using the remote part for learning I'll circle back to clean up what you have indicated based on your feedback and my discovery.

lgfa29:
Hi @trilom,

Thanks for this work, it was helpful to flag to us where things needed to change. Since we're moving our demos to their own repo, I moved these changes and finished the remaining work in https://github.com/hashicorp/nomad-autoscaler-demos/pull/3, with you as the co-author, and so I am going to close this one.

Thank you for the contribution once again!

PR #329, opened 2021-01-04
test: fix-up tooling installation and dependancy management.

jrasell:
I don't know why this worked and stopped after the new year. This
change however fixes the seen errors (locally and CI) when
installing the tools and fixes the module name.

This should fix the lint failure seen in #328


cgbaker (approved):
lgtm

PR #334, opened 2021-01-05
Query Nomad APM for memory values relative to task group

lgfa29:
Following the work on #324, this PR adds a new query type for the Nomad APM to use percentage values based on allocated totals rather than looking at the entire cluster.


cgbaker (approved):
👍

PR #346, opened 2021-01-13
test: use gotestsum to capture test output for easier viewing.

jrasell:
The standard go test output has increasingly become long and hard
to grok. Inline with Nomad, the Nomad Autoscaler would benefit
from using gotestsum to create a much more user friendly test
output.


cgbaker (approved):
👍

PR #347, opened 2021-01-14
ci: use single CircleCI config file.

jrasell:
This work gave me a chance to remove some items we didn't need and generally vet the config. I moved the Docker class to small which impacts the overall CI time by roughly 10-20s only. I'll raise a follow up PR to cover the Enterprise side of things.


cgbaker (approved):
👍

lgfa29 (approved):
Seeing them side-by-side, the single file is easier to under. 

Thanks!

PR #349, opened 2021-01-15
agent: add pprof HTTP debug endpoints.

jrasell:
The Nomad Autoscaler is mildly complex and debugging problems can be tricky when they are related to items such as go routine deadlocks. The optional pprof endpoints allow developers and operators to easily obtain profiling data from the Nomad Autoscaler, no matter the persons ability with the Go programming language.

We have no control over the API, therefore they do not sit under the `/v1` prefix.

p.s. @cgbaker sorry for adding more endpoints!

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -32,6 +32,9 @@ type Agent struct {
 	// LogJson enables log output in JSON format.
 	LogJson bool `hcl:"log_json,optional"`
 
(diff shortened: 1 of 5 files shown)


cgbaker (approved):
cool. good call.

lgfa29 (approved):
💯 

I need to learn how to use this 😄

PR #352, opened 2021-01-21
Print metrics received from APM to debug log

baxor:
This does make the Debug logs a little more chatty than they usually are, as it logs all metrics in the time series.  However, it is something I need for debugging right now and figured it may be desirable to others as well. 

```2021-01-21T22:34:32.662Z [DEBUG] policy_eval.worker.check_handler: received metrics from APM: check=high-cpu-applications-scale-cpu-utilization id=2f898ada-ae90-4212-ab3c-25a1ec622db7 policy_id=476ad545-9fa3-ecf4-3ee5-262a6ef7fb20 queue=cluster source=datadog strategy=target-value target=aws-asg metrics=["{2021-01-21 22:29:40 +0000 UTC 26.886199373834266}", "{2021-01-21 22:29:50 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:00 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:10 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:20 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:30 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:40 +0000 UTC 26.886199373834266}", "{2021-01-21 22:30:50 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:00 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:10 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:20 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:30 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:40 +0000 UTC 26.886199373834266}", "{2021-01-21 22:31:50 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:00 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:10 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:20 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:30 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:40 +0000 UTC 26.886199373834266}", "{2021-01-21 22:32:50 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:00 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:10 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:20 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:30 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:40 +0000 UTC 26.886199373834266}", "{2021-01-21 22:33:50 +0000 UTC 26.886199373834266}", "{2021-01-21 22:34:00 +0000 UTC 26.886199373834266}", "{2021-01-21 22:34:10 +0000 UTC 26.886199373834266}", "{2021-01-21 22:34:20 +0000 UTC 26.886199373834266}"]```

### CODE REVIEW

File: policyeval/base_worker.go [hunk 1]
```diff
@@ -323,6 +323,8 @@ func (h *checkHandler) start(ctx context.Context) {
 		return
 	}
 
+	h.logger.Debug("received metrics from APM", "metrics", h.checkEval.Metrics)
```
lgfa29, line 326:
```suggestion
	for _, m := range h.checkEval.Metrics {
		h.logger.Trace("metric result", "ts", m.Timestamp, "value", m.Value)
	}
```

lgfa29 (changes requested):
Thanks for the contribution @baxor, this does seem helpful.

But it could get quite verbose, so a `TRACE` log level would be better. Also, to make it easier to parse the output, each metric could be printed in its own line, which would allow you to do something like this:

```shell-session
$ nomad-autoscaler agent 2>&1 | grep "metric result" | grep check=avg_instance_sessions | cut -d' ' -f 13-17
ts="2021-01-22 10:49:35 -0500 EST" value=0
ts="2021-01-22 10:49:36 -0500 EST" value=0
ts="2021-01-22 10:49:37 -0500 EST" value=0
ts="2021-01-22 10:49:38 -0500 EST" value=0
ts="2021-01-22 10:49:39 -0500 EST" value=0
ts="2021-01-22 10:49:40 -0500 EST" value=0
ts="2021-01-22 10:49:41 -0500 EST" value=0
```

What do you think of this approach?

policyeval/base_worker.go
baxor: Great Idea, will make these changes

policyeval/base_worker.go
```diff
@@ -323,6 +325,10 @@ func (h *checkHandler) start(ctx context.Context, currentStatus *sdk.TargetStatu
 		return &sdk.ScalingAction{Direction: sdk.ScaleDirectionNone}, nil
 	}
 
+	for _, m := range h.checkEval.Metrics {
+		h.logger.Trace("metric result", "ts", m.Timestamp, "value", m.Value)
+	}
```
jrasell, line 330: I wonder if this is worth wrapping with `if h.logger.IsTrace() {}` so that we only enter the loop if we are going to print the metrics and this avoids looping tens/hundreds of times when we don't need to.

policyeval/base_worker.go
```diff
@@ -86,6 +87,7 @@ func (w *BaseWorker) Run(ctx context.Context) {
 		// Notify broker that policy eval was successful.
 		if err := w.broker.Ack(eval.ID, token); err != nil {
 			logger.Warn("failed to ACK policy evaluation", "err", err)
+			os.Exit(1)
```
jrasell, line 90: Is this a change we are expecting to include? If so, could you provide a little context on the change and consider moving it to a separate PR?

baxor: Ahh woops, this snuck in! -- I was testing this weeks ago to restart the autoscaler when the evaluation threads would hang.  My bad, removing that.

jrasell (approved):
LGTM thanks @baxor

PR #365, opened 2021-01-28
Use default values for the Nomad client as provided by the `api` package

### LINKED ISSUE #357: Read NOMAD_ADDR and NOMAD_REGION environment variables
It's common to use `NOMAD_*` environment variables to configure Nomad clients. The Autoscaler does call [`api.DefaultConfig`](https://github.com/hashicorp/nomad-autoscaler/blob/v0.2.1/agent/agent.go#L151), which handles environment variables, but there are [hardcoded defaults](https://github.com/hashicorp/nomad-autoscaler/blob/master/agent/config/config.go#L344-L347) for region and address that end up overriding the environment variables.

lgfa29:
The Autoscaler was accidentally overriding the defaults returned by Nomad's [`api.DefaultConfig()`](https://github.com/hashicorp/nomad/blob/v1.0.2/api/api.go#L243). 

I couldn't see a reason why the Autoscaler would need to set default values on its own, so I just removed them. But I could be missing something.

It's also not a breaking change since the default values being set are the same as `api.DefaultConfig()` defines.

Closes #357

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -294,14 +294,6 @@ const (
 	// the PluginDir default value.
 	defaultPluginDirSuffix = "/plugins"
 
-	// defaultNomadAddress is the default address used for Nomad API
-	// connectivity.
-	defaultNomadAddress = "http://127.0.0.1:4646"
-
-	// defaultNomadRegion is the default Nomad region to use when performing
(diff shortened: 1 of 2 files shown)


cgbaker (approved):
👍

PR #378, opened 2021-02-10
scaleutils: filter nodes to ensure unstable pools do not run eval.

### LINKED ISSUE #376: plugins/target: remote target plugins don't track in-flight draining processes
The current implementation of the Nomad remote target plugins (`aws-asg`, `azure-vmss` and `gcp-mig`) do not perform any state tracking when draining nodes before interacting with the remote resources. This causes undesired behaviour in situations where the draining takes longer than the policy evaluation interval. Specifically; the Autoscaler will request another scaling action and select new nodes to drain alongside the original action.

The Nomad Autoscaler and the remote target plugins should have functionality that allows it to track when a scaling action is in-flight no matter what stage that is. This will then allow the core workflow to skip policy evaluation, as the target is not ready.

As a work around until this is fixed, it is advisable to set the evaluation interval of the remote target policy at a value higher than that of the drain deadline.

jrasell:
In situations where cluster scaling evaluations are triggered at
a rate faster than cluster scaling actions can complete,
the autoscaler would drain more and more nodes, eventually leaving
the pool without any.

The fix adds better, more thorough node pool filtering to error in
situations where the nodes within a pool are considered unstable.
This ensures the Nomad Autoscaler will only perform cluster scaling
evaluations when we can be sure the state is stable and known.

closes #376

### DIFF
--- a/sdk/helper/scaleutils/filter.go
+++ b/sdk/helper/scaleutils/filter.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	multierror "github.com/hashicorp/go-multierror"
 	"github.com/hashicorp/nomad/api"
 )
 
@@ -20,12 +21,22 @@ type PoolIdentifier struct {
 	Value         string
 }
 
+// Validate is used to check the validation of the PoolIdentifier object.
+func (p *PoolIdentifier) Validate() error {
+	switch p.IdentifierKey {
+	case IdentifierKeyClass:
+		return nil
+	default:
+		return fmt.Errorf("unsupported node pool identifier: %q", p.IdentifierKey)
+	}
+}
+
 // IdentifyNodes filters the supplied node list based on the PoolIdentifier
 // params.
 func (p *PoolIdentifier) IdentifyNodes(n []*api.NodeListStub) ([]*api.NodeListStub, error) {
 	switch p.IdentifierKey {
 	case IdentifierKeyClass:
-		return filterByClass(n, p.Value), nil
+		return filterByClass(n, p.Value)
 	default:
 		return nil, fmt.Errorf("unsupported node pool identifier: %q", p.IdentifierKey)
 	}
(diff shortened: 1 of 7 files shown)


sdk/helper/scaleutils/filter.go
```diff
@@ -88,37 +99,78 @@ const nodeAttrGCEZone = "platform.gce.zone"
...
+		// Filter out all nodes which do not match the target class first.
+		if node.NodeClass != "" && node.NodeClass != id ||
+			node.NodeClass == "" && id != defaultClassIdentifier {
 			continue
 		}
 
-		// Ignore nodes that are not currently eligible.
-		if node.SchedulingEligibility != api.NodeSchedulingEligible {
+		// We should class an initializing node as an error, this is caused by
+		// node registration and could be sourced from scaling out.
```
cgbaker, line 127: should we be nervous here? is there a guarantee in nomad that a client will eventually move from init to ready? otherwise, this could block cluster scaling forever.

cgbaker (approved):
lgtm. just a couple of typos and a few questions. 👍

sdk/helper/scaleutils/filter.go
```diff
@@ -88,37 +99,90 @@ const nodeAttrGCEZone = "platform.gce.zone"
...
+
 	for _, node := range n {
 
-		// Ignore nodes that are not in a ready state.
-		if node.Status != api.NodeStatusReady {
+		// Track till 10 nodes in an unstable state so that we have some
+		// efficiency, whilst still responding with useful information. It also
+		// avoids error logs messages which are extremely long and potentially
+		// unsuitable for log aggregators.
+		if err != nil && err.Len() == 10 {
```
cgbaker, line 124:
```suggestion
		if err != nil && err.Len() >= 10 {
```

cgbaker: a bit safer, just in case we (e.g.) add two errors and skip over length 10

lgfa29 (approved):

baxor:
I just tried the nightly build and am still getting a build up of ineligible nodes that are not removed from the ASG.  The autoscaler adds more to compensate for evicted/blocked jobs and eventually hits the max nodes.  Currently the drain timeout is 5m, evaluation interval is 7m, and cooldown is 10m.  

Interestingly, a nightly build from 2/9 with the above interval/timeout/cooldown settings did not exhibit this behavior, so that workaround is now ineffective. 

Perhaps a scale-in-maximum would be useful, basically never allow more than N nodes to be marked as ineligible/draining/scaling out at any given time?

jrasell:
@baxor unfortunately there is a problem with the nightly builds (#384) which means the [last build](https://github.com/hashicorp/nomad-autoscaler/releases/tag/nightly) ran 10 days ago and therefore doesn't include this fix. Are you able to try running a build from the current master branch?

baxor:
yes, correct - building via https://github.com/hashicorp/nomad-autoscaler/archive/nightly.tar.gz

baxor:
This appears to be behaving correctly with a build from master -- nodes have been marked ineligible but no further nodes.   There is an issue of the ASG not being scaled in.  I will open a separate issue.

PR #383, opened 2021-02-15
plugin/apm/prometheus: only access config map once to set config.

jrasell:
This little manoeuvre's gonna save us 4 nanoseconds.


cgbaker (approved):
don't spend it all in one place 😄

PR #390, opened 2021-02-19
Expand SIGHUP handling

### LINKED ISSUE #375: Reload Credentials on SIGHUP
We're currently running the autoscaler as a Nomad job with an AWS ASG configured as a target. The autoscaling job is using STS credentials for AWS authentication and a Nomad ACL token that are both retrieved from Vault. Naturally, Nomad restarts the autoscaler alloc as these credentials approach their TTLs and are renewed. When this occurs in the middle of a scale in event, we end up with drained and ineligible, but not terminated, nodes left in the ASG seemingly indefinitely. The series of events goes something like:

1. autoscaler starts draining a node
2. Before the node drain is complete, Nomad restarts the autoscaler alloc to template in new Vault credentials
3. The new autoscaler allocation seemingly has no state tracking the started-but-not-completed node scale in activity

If the behavior of autoscaler + a SIGHUP signal was extended to cover reloading any potential credential sources that would resolve our immediate issue. Alternatively, if there is an existing mechanism to support this usage, please do let us know! 😄

lgfa29:
This PR expands the effect of sending a `SIGHUP` signal to reload the Autoscaler configuration.

It specifically reloads plugin configurations, stops/starts plugins if their blocks are removed/added, and updates instances of Nomad clients (withing plugins and the policy source).

Other configuration values are also updated in memory, but have no effect (for example, increasing the number of workers for a queue).

Closes  #375

### DIFF
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -23,6 +23,7 @@ import (
 type Agent struct {
 	logger        hclog.Logger
 	config        *config.Agent
+	configPaths   []string
 	nomadClient   *api.Client
 	pluginManager *manager.PluginManager
 	policyManager *policy.Manager
@@ -35,11 +36,12 @@ type Agent struct {
 	nomadCfg *api.Config
 }
 
-func NewAgent(c *config.Agent, logger hclog.Logger) *Agent {
+func NewAgent(c *config.Agent, configPaths []string, logger hclog.Logger) *Agent {
 	return &Agent{
-		logger:   logger,
-		config:   c,
-		nomadCfg: nomadHelper.MergeDefaultWithAgentConfig(c.Nomad),
+		logger:      logger,
+		config:      c,
+		configPaths: configPaths,
+		nomadCfg:    nomadHelper.MergeDefaultWithAgentConfig(c.Nomad),
 	}
 }
 
@@ -164,8 +166,25 @@ func (a *Agent) generateNomadClient() error {
 // reload triggers the reload of sub-routines based on the operator sending a
 // SIGHUP signal to the agent.
 func (a *Agent) reload() {
+	a.logger.Info("reloading Autoscaler configuration")
+
+	// Reload config files from disk.
+	newCfg, err := config.LoadPaths(a.configPaths)
+	if err != nil {
+		a.logger.Error("failed to reload Autoscaler configuration", "error", err)
+		return
+	}
+
+	a.config = newCfg
+	a.nomadCfg = nomadHelper.MergeDefaultWithAgentConfig(newCfg.Nomad)
+
 	a.logger.Debug("reloading policy sources")
 	a.policyManager.ReloadSources()
+
+	a.logger.Debug("reloading plugins")
+	if err := a.pluginManager.Reload(a.setupPluginsConfig()); err != nil {
+		a.logger.Error("failed to reload plugins", "error", err)
+	}
 }
 
 // handleSignals blocks until the agent receives an exit signal.
--- a/agent/agent.go
+++ b/agent/agent.go
@@ -26,6 +26,7 @@ type Agent struct {
 	configPaths   []string
 	nomadClient   *api.Client
 	pluginManager *manager.PluginManager
+	policySources map[policy.SourceName]policy.Source
 	policyManager *policy.Manager
 	inMemSink     *metrics.InmemSink
 	evalBroker    *policyeval.Broker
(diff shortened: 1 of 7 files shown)


agent/agent.go
```diff
@@ -164,8 +168,35 @@ func (a *Agent) generateNomadClient() error {
...
+	if err != nil {
+		a.logger.Error("failed to reload Autoscaler configuration", "error", err)
+		return
+	}
+
+	a.config = newCfg
+	a.nomadCfg = nomadHelper.MergeDefaultWithAgentConfig(newCfg.Nomad)
+
+	if err := a.generateNomadClient(); err != nil {
+		a.logger.Error("failed to reload Autoscaler configuration", "error", err)
```
cgbaker, line 184: i wasn't sure about this error message being the same as above, but it look `generateNomadClient` adds "failed to instantiate Nomad client" to any message from Nomad, so that's fine

policy/nomad/source.go
```diff
@@ -71,56 +85,70 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
+		// still listen for the context closing or a reload request.
+		blockingQueryCompleteCh := make(chan struct{})
+		go func() {
+			// Obtain a handler now so we can release the lock before starting
+			// the blocking query.
+			s.nomadLock.RLock()
+			scaling := s.nomad.Scaling()
+			s.nomadLock.RUnlock()
+
+			policies, meta, err = scaling.ListPolicies(q)
```
cgbaker, line 105: i'm still annoyed that we can't send a context into these blocking queries...

policy/nomad/source.go
```diff
@@ -71,56 +85,70 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
-			if err != nil {
-				policy.HandleSourceError(s.Name(), fmt.Errorf("failed to call the Nomad list policies API: %v", err), req.ErrCh)
-				time.Sleep(10 * time.Second)
-				continue
-			}
+		// If we get an errors at this point, we should sleep and try again.
+		// TODO(jrasell) in the future maybe use a better method than sleep.
+		if err != nil {
+			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to call the Nomad list policies API: %v", err), req.ErrCh)
+			time.Sleep(10 * time.Second)
```
cgbaker, line 123: while we're here, maybe this should be a select over `time.After` and `ctx.Done`, to keep an error from holding up our shutdown for 10s?

policy/nomad/source.go
```diff
@@ -139,59 +167,79 @@ func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq)
...
-			// If the index has not changed, the query returned because the timeout
-			// was reached, therefore start the next query loop.
-			if !blocking.IndexHasChanged(meta.LastIndex, q.WaitIndex) {
-				continue
-			}
+		// If we get an errors at this point, we should sleep and try again.
+		// TODO(jrasell) in the future maybe use a better method than sleep.
+		if err != nil {
+			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy: %v", err), req.ErrCh)
+			time.Sleep(10 * time.Second)
```
cgbaker, line 211: as above, so below

cgbaker (approved):
lgtm, good job cleaning up all the detected race conditions

agent/agent.go
```diff
@@ -164,8 +168,35 @@ func (a *Agent) generateNomadClient() error {
 // reload triggers the reload of sub-routines based on the operator sending a
 // SIGHUP signal to the agent.
 func (a *Agent) reload() {
+	a.logger.Info("reloading Autoscaler configuration")
+
+	// Reload config files from disk.
+	newCfg, err := config.LoadPaths(a.configPaths)
+	if err != nil {
+		a.logger.Error("failed to reload Autoscaler configuration", "error", err)
+		return
```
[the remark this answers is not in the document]
lgfa29: I thought about this, but I am not 100% what would be the best approach. I think the options are:

#### 1. Current approach
I think it's fine, but it does hide errors from operators.

#### 2. Ignore config error and reload sources
I think this would also be fine but still hides the error. 

The only potential probably I see is that we would need to be careful in the future to recognize that, in this state, the configuration being provided to the Autoscaler is invalid and should be ignored, but everyone reads comments right 😄 

#### 3. Just crash
In this state, the Autoscaler wouldn't be able to start, so there could be an argument that it shouldn't be able to restart either. 

This is the most operator-visible option, and quite disruptive. But maybe desirable? Given that the Autoscaler sits in the control plane, it might be better for it to crash sooner and loudly (failed allocs will be displayed in the UI) rather than during an outage caused by a policy not being evaluated?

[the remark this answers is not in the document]
lgfa29: After some discussion we decided to go with option 3 for now. This could be improved in future releases once we have a better mechanism to surface errors.

policy/nomad/source.go
lgfa29: :nod:

PR #392, opened 2021-02-19
plugin/target/aws-asg: check instances in non-running state.

### LINKED ISSUE #391: plugin/target/aws-asg: checking terminated instances loop exits too early
The confirmation of AWS EC2 instance termination is ending too quickly, particularly in action events with multiple nodes and results in node purge problems as the autoscaler attempts to trigger this before the node is classed as down.

jrasell:
It could also be worth adding logic to `RunPostScaleInTasks` to check the node state before purging on some retry interval to ensure the purge call has the desired affects. I'll raise a followup issue to track this.

closes #391


cgbaker (approved):
👍

lgfa29 (approved):

PR #395, opened 2021-02-22
scaleutils: refactor to make external plugins possible and easier

### LINKED ISSUE #259: scaleutils: remove setup burden on target plugin authors
The current way in which the scaleutils are written puts a larger than required burden on the target plugin to parse and validate config params in order to populate a scale request object before calling scale util functions. Each horizontal cluster scaling target plugin will need to replicate this work, meaning it should be moved into the scaleutils so it happens in a single place.

This specifically relates to the creation of the [ScaleInReq](https://github.com/hashicorp/nomad-autoscaler/blob/d854b629fda7c630ad89dc82abe65bec4d9d63aa/helper/scaleutils/request.go#L18).

### LINKED ISSUE #377: ensure that scaleutils can be used by external plugins
blocker for closing #370 , need to ensure that scaleutils can be effectively used by external plugins

jrasell:
Refactor of the scaleutils package so that external plugins are possible without need to modify the core codebase. The setup of the package is also easier, reducing the strain on plugin developers performing tasks such as config map parsing where possible.

The old functionality has been marked as deprecated and the files moved to use a deprecated prefix. This has been done in the hope to ensure the deprecated code is removed in a timely manner, and not missed easily.

There is a new nodepool package that handles the node pool identification methods. This is wrapped in an interface, allowing us to add additional methods in the future if we desire. This is internal on purpose as I don't see any real reason why this should be the responsibility of plugins, or to even allow that possibility.

All multierrors from scaleutils are now wrapped so their output is formatted in a much easier to read manner. This is currently only used within this package, but would be worth extending to the whole application in the future.

closes #259
closes #377


sdk/helper/scaleutils/cluster.go
```diff
@@ -0,0 +1,296 @@
...
+	var (
+		out  []NodeResourceID
+		mErr *multierror.Error
+	)
+
+	for _, node := range nodes {
+
+		// Read the full node object from the API which will contain the full
+		// information required to identify the remote provider ID. If we get a
+		// single error here, its likely we won't be able to perform any of the
```
cgbaker, line 166:
```suggestion
		// single error here, it's likely we won't be able to perform any of the
```

cgbaker (approved):
lgtm  💯

lgfa29 (approved):
:shipit:

PR #399, opened 2021-02-23
demos: move demos to a separate repo

lgfa29:
Having the demos in-tree along with the rest of the Autoscaler source code required users to clone the entire repo in order to follow them. 

By moving the demos to a separate repo it becomes easier to retrieve just the files required to run them and it provides a way to do better version control.


cgbaker (approved):
👍

jrasell (approved):

PR #416, opened 2021-03-09
agent: return req error if prometheus metrics are disabled.

### LINKED ISSUE #398: telemetry: Prometheus config flag `prometheus_metrics` doesn't alter endpoint behaviour
Using the CLI flag `-telemetry-prometheus-metrics` or equivalent config file option does not alter whether or not metrics are exposed in Prometheus format via the API.

jrasell:
If the user has disabled Prometheus metrics and a request is
sent to the metrics endpoint requesting Prometheus formatted
metrics, then the request should fail.

The returned response code copies that of Consul and Nomad,
to provide a consistent user experience.

closes #398

### DIFF
--- a/agent/http/server.go
+++ b/agent/http/server.go
@@ -51,6 +51,10 @@ type Server struct {
 	mux *http.ServeMux
 	srv *http.Server
 
+	// promEnabled tracks whether Prometheus formatted metrics should be
+	// enabled.
+	promEnabled bool
+
 	// aliveness is used to describe the health response and should be set
 	// atomically using healthAlivenessReady and healthAlivenessUnavailable
 	// const declarations.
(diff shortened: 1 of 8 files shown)


cgbaker (approved):
👍

agent/http/server.go
```diff
@@ -62,12 +66,13 @@ type Server struct {
 }
 
 // NewHTTPServer creates a new agent HTTP server.
-func NewHTTPServer(debug bool, cfg *config.HTTP, log hclog.Logger, agent AgentHTTP) (*Server, error) {
+func NewHTTPServer(debug, prom bool, cfg *config.HTTP, log hclog.Logger, agent AgentHTTP) (*Server, error) {
```
lgfa29, line 69: I think we're getting to the point where the input to `NewHTTPServer` would be better as a struct 😄

lgfa29 (approved):
Minor comment on config sprawl on the HTTP server, but not a blocker 👍

agent/http/server.go
jrasell: Agreed and this did cross my mind that we just pass the whole config object. I'll follow this up in a new PR to refactor the func signature.

lgfa29: I think passing the whole config would be too much, just the current params in a `HTTPServerConfig` struct should be enough.

PR #418, opened 2021-03-12
agent: fix issue causing incorrect debug/prom HTTP server config.

jrasell:
big sigh, sorry.


lgfa29 (approved):
big sight, sorry x2 for missing it in the code review

PR #421, opened 2021-03-15
add CLI flags for PolicyEval config

### LINKED ISSUE #291: Document and add new CLI flags for the policy_eval config
The new `policy_eval` configuration block should be documented and have corresponding CLI flags.

lgfa29:
The `EvaluateAfter` config will be moved to a new configuration block, so it was left out of this PR.

Closes #291

### DIFF
--- a/command/agent.go
+++ b/command/agent.go
@@ -121,6 +121,18 @@ Policy Options:
     The default evaluation interval that will be applied to all scaling policies
     which do not specify an evaluation interval.
 
+Policy Evaluation Options:
+
+  -policy-eval-ack-timeout=<dur>
+    The time limit that an eval must be ACK'd before being considered NACK'd.
+
+  -policy-eval-delivery-limit=<num>
+    The maximum number of times a policy evaluation can be dequeued from the broker.
+
+  -policy-eval-workers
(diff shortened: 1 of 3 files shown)


command/agent.go
```diff
@@ -121,6 +121,18 @@ Policy Options:
...
+
+  -policy-eval-ack-timeout=<dur>
+    The time limit that an eval must be ACK'd before being considered NACK'd.
+
+  -policy-eval-delivery-limit=<num>
+    The maximum number of times a policy evaluation can be dequeued from the broker.
+
+  -policy-eval-workers=<key:value>
+    The number of workers to initialize for each queue, formatted as
+    <queue1>:<num>,<queue2>:<num>.
```
jrasell, line 134: I'm curious if it would be worth detailing the possible queues here so operators can easily discover utilising this flag?

jrasell (approved):
LGTM

Minor question on the flag help, but certainly not a blocker.

cgbaker (approved):
lgtm

command/agent.go
lgfa29: I added to the the [docs](https://github.com/hashicorp/nomad/pull/10183), but you are right, it's good to have it here as well.

PR #422, opened 2021-03-16
Add configuration and CLI flags related to DAS

lgfa29:
Initial work to allow operators to customize internal settings for Dynamic Application Sizing.


agent/config/config.go
```diff
@@ -431,6 +469,41 @@ func (a *Agent) Validate() error {
 	return result.ErrorOrNil()
 }
 
+func (d *DynamicApplicationSizing) merge(b *DynamicApplicationSizing) *DynamicApplicationSizing {
+	if d == nil {
+		return b
+	}
```
cgbaker, line 475: 👍 😉

agent/config/config.go
```diff
@@ -327,8 +359,9 @@ func Default() (*Agent, error) {
 	}
 
 	return &Agent{
-		LogLevel:  defaultLogLevel,
-		PluginDir: pwd + defaultPluginDirSuffix,
+		LogLevel:                 defaultLogLevel,
+		PluginDir:                pwd + defaultPluginDirSuffix,
+		DynamicApplicationSizing: &DynamicApplicationSizing{},
```
cgbaker, line 364: no defaults for these?

cgbaker (approved):
lgtm

agent/config/config.go
lgfa29: It's almost like my previous PRs were a giant ball of mess that I had to split up 😬

agent/config/config.go
lgfa29: I thought it would be better to set the defaults in the ENT repo, so for OSS this config block is completely ignored. 

The parsing of config and CLI flags is done here to avoid errors if an OSS binary is used when loading a config that has ENT values. 

We could check if ENT-only configs are used and emit a warning or exit early with a nicer error message.

PR #425, opened 2021-03-17
plugin/target/aws-asg: use single ASG call rather than split ASG/EC2 on scale-in.

### LINKED ISSUE #361: plugin/aws-asg: use TerminateInstanceInAutoScalingGroupRequest rather than separate calls
In its current implementation, the AWS ASG target plugin uses `DetachInstancesRequest` and `TerminateInstancesRequest` to perform a scaling in action. This has the benefit that each calls accepts a list of instanceIDs, but has the downside of being two, long lived calls.

Moving the plugin to use `TerminateInstanceInAutoScalingGroupRequest` means we can avoid the use of two separate calls to perform scaling in, thus reducing the possibility of the autoscaler getting interrupted and leaving orphaned instances running within the AWS account. The downside is that the call only accepts a single instanceID, therefore some thought should be put into how to handle partial failures when dealing with > 1 instance.

### LINKED ISSUE #412: Scaling down > 20 instances at once in AWS ASG fails
Hello! 👋 

We are using the `nomad-autoscaler` to effectively scale up for running large batch jobs and then back down again at the end. However we are seeing the autoscaler throw errors when trying to scale down more than 20 instances at a time (anything less works a-ok!)

**Steps To Reproduce:**

1. Create an AWS ASG with a maximum capacity of > 20.
2. Create a policy like this:

```hcl
scaling "my_amazing_scaling" {
  enabled = true
  min     = 0
  max     = 21

  policy {
    cooldown            = "2m"
    evaluation_interval = "30s"

    check "desired" {
      source = "prometheus"
      query  = "21" # <- change this to 0 and see what happens
      strategy "target-value" {
        target = 0
      }
    }

    target "aws-asg" {
      dry-run             = "false"
      aws_asg_name        = "your_asg"
      node_class          = "your_asg"
      node_drain_deadline = "5m"
    }
  }
}
```
3. Wait for boxes to scale up.
4. Change `query` from `"21"` to `"0"`.

**Expected Behaviour:**

The autoscaler should correctly terminate all 21 instances.

**Actual Behaviour:**

The autoscaler fails to terminate the instances with the following error:

```
2021-03-03T18:09:08.148Z [ERROR] policy_eval.worker: failed to evaluate policy: eval_id=<id> eval_token=<id> id=<id> policy_id=<id> queue=cluster err="failed to scale target: failed to perform scaling action: failed to scale in AWS AutoScaling Group: failed to detach intances from Autoscaling Group: ValidationError: List exceeds the maximum number of instances to detach. Please retry your request with fewer instances.
	status code: 400, request id: <id>"
```

This is because there is a maximum limit of 20 instances in a single detch call in the AWS API:

> You can use AttachInstances, DetachInstances, EnterStandby, and ExitStandby with at most 20 instance IDs at a time.

https://docs.aws.amazon.com/autoscaling/ec2/userguide/as-account-limits.html

jrasell:
closes #361 
closes #412 

This changes the AWS ASG plugin to use `TerminateInstanceInAutoScalingGroup` when scaling in nodes rather than a call against the ASG to detach instances and then another to terminate instances. This solves a core bug, whereby decrementing an ASG accepts a maximum of 20 instance IDs.

The downside of moving to this function, is that unlike Azure and GCP, AWS ASG only accepts a single instance ID per call to terminate an instance from an ASG. In order to accommodate this, the plugin includes some additional logic in order to understand and correctly respond to situations where, in a single scaling event, some termination calls might succeed, and others might fail. This creates a partial success. In partial success situations, the plugin attempts to toggle the eligibility on the instances that failed to terminate, in order to reconcile the node pool.

When performing the AWS API call, the ASG name does not need to be passed. This creates a potential situation where a node is misconfigured to reside in the pool. To counter this, before scaling in, the plugin ensures all selected instances do in-fact belong to the target ASG. This has resulted in exporting the functions that were hidden behind the scaleutils `RunPreScaleInTasks` so the workflow can be better controlled.

### DIFF
--- a/plugins/builtin/target/aws-asg/plugin/aws.go
+++ b/plugins/builtin/target/aws-asg/plugin/aws.go
@@ -2,13 +2,14 @@ package plugin
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/aws/external"
 	"github.com/aws/aws-sdk-go-v2/service/autoscaling"
-	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils"
 	"github.com/hashicorp/nomad/api"
 )
 
@@ -55,8 +56,7 @@ func (t *TargetPlugin) setupAWSClients(config map[string]string) error {
 		cfg.Credentials = aws.NewStaticCredentialsProvider(keyID, secretKey, session)
 	}
 
-	// Set up our AWS clients.
-	t.ec2 = ec2.New(cfg)
+	// Set up our AWS client.
 	t.asg = autoscaling.New(cfg)
 
 	return nil
@@ -93,18 +93,36 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, asg *autoscaling.AutoScalin
 
 func (t *TargetPlugin) scaleIn(ctx context.Context, asg *autoscaling.AutoScalingGroup, num int64, config map[string]string) error {
 
-	ids, err := t.clusterUtils.RunPreScaleInTasks(ctx, config, int(num))
+	// The AWS plugin utilises the individual scaleutils calls due to the way
+	// in which the TerminateInstanceInAutoScalingGroupRequest call works. We
+	// need to ensure the nodes are part of the target ASG before performing
+	// actions, and draining the nodes just to get an error at this phase is
+	// wasteful and means we need reconciliation.
+	if t.clusterUtils.ClusterNodeIDLookupFunc == nil {
+		return errors.New("required ClusterNodeIDLookupFunc not set")
+	}
+
+	nodes, err := t.clusterUtils.IdentifyScaleInNodes(config, int(num))
 	if err != nil {
-		return fmt.Errorf("failed to perform pre-scale Nomad scale in tasks: %v", err)
+		return err
 	}
 
-	// Grab the instanceIDs once as it is used multiple times throughout the
-	// scale in event.
-	var instanceIDs []string
-
-	for _, node := range ids {
-		instanceIDs = append(instanceIDs, node.RemoteResourceID)
+	nodeResourceIDs, err := t.clusterUtils.IdentifyScaleInRemoteIDs(nodes)
+	if err != nil {
+		return err
 	}
+
+	// Any error received here indicates misconfiguration between the ASG and
+	// the Nomad node pool.
+	instanceIDs, err := instancesBelongToASG(asg, nodeResourceIDs)
+	if err != nil {
+		return err
+	}
+
+	if err := t.clusterUtils.DrainNodes(ctx, config, nodeResourceIDs); err != nil {
+		return err
+	}
+	t.logger.Info("pre scale-in tasks now complete")
 
 	// Create the event writer and write that the drain event has been
 	// completed which is part of the RunPreScaleInTasks() function.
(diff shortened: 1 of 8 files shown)


plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -93,105 +93,168 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, asg *autoscaling.AutoScalin
...
+		if err := t.clusterUtils.RunPostScaleInTasks(ctx, config, result.successfulIDs()); err != nil {
+			return fmt.Errorf("failed to perform post-scale Nomad scale in tasks: %v", err)
+		}
 	}
 
-	return nil
+	if result.lenFailure() > 0 && result.lenSuccess() > 0 {
+		log.Warn("partial scaling success",
+			"success_num", result.lenSuccess(), "failed_num", result.lenFailure())
+		return nil
```
jrasell, line 170: I would appreciate peoples view on this part in particular!

cgbaker: so, as long as we had at least one successful scale in, we don't return an error here, with the assumption that we'll retry.

one thing that i'm not sure about is that we don't keep the error if we failed to reconcile the node state; `RunPostScaleInTasksOnFailure` doesn't return an error. so it's possible that we got to this point and we have nodes that were not destroyed but are still ineligible. any thoughts on that? should that error case prevent us from returning a null on partial success?

cgbaker (approved):
one question about partial-partial success, otherwise, LGTM!

plugins/builtin/target/aws-asg/plugin/aws.go
jrasell: Ah yes this is a great idea.

So collect any errors from `RunPostScaleInTasksOnFailure`. In the event there are some and we have a partial result, return these errors to the caller specifying the scaling action failed because we were unable to reconcile the state thus indicating the operator needs to take action.

jazzyfresh:
This change requires an update to the nomad-autoscaler-demos repo with new IAM permissions

lgfa29 (approved):
Tested and updated demo and docs for the new permission requirement 👍

plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -159,15 +162,22 @@ func (t *TargetPlugin) scaleIn(ctx context.Context, asg *autoscaling.AutoScaling
...
 
 		// Run any post scale in tasks that are desired.
-		if err := t.clusterUtils.RunPostScaleInTasks(ctx, config, result.successfulIDs()); err != nil {
-			return fmt.Errorf("failed to perform post-scale Nomad scale in tasks: %v", err)
-		}
+		successTaskErr = t.clusterUtils.RunPostScaleInTasks(ctx, config, result.successfulIDs())
+	}
+
+	// The tasks run on nodes that have been successfully terminated should not
+	// cause a failure of the scaling pipeline.
```
cgbaker, line 169: 👍

cgbaker:
changes look good!

bansawbanchee:
The whole point of using an auto scaling resource is to provide the ability to fully customize how your cluster scales in and out. The strategies that are prebuilt will cover most users requirements out of the box. You can use custom metrics within the terraform resources as It is built in. [Scaling the size of your Auto Scaling group]
(https://docs.aws.amazon.com/autoscaling/ec2/userguide/scaling_plan.html)

You would also need to account for what the end user configures across the resources used to build the autoscaling group before just terminating an instance. You could seriously damage something. [The docs](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group) start off with warning too.

Adjusting the `capacity` variables on the asg will fire off a `terminating` event that you can query for that tells you which instance is going to be `terminated` in two minutes. That gives you two minutes to migrate/drain the jobs on that instance. 
[Waiting for capacity](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/autoscaling_group#waiting-for-capacity)  has some more information on capacity management in terraform.

If you must manage capacity for aws resources that somehow can't be managed dynamically by aws then you should use the `ec2` instance resource. This allows you to manage the instances without having to deal with an actual aws asg, which could prove fatal.. You already group the nodes by `tags` on the instances so you could use those to identify a "pool".

The modules used in your demo just need to be built up with the rest of the resources aws provides for handling autoscaling.

jrasell:
Hi @bansawbanchee and thank you for the feedback. You’re absolutely correct that autoscaling could be achieved using AWS provided tooling. That being said, having finer grained control and internal control loops give the Nomad Autoscaler several features that are utilised by operators. Particularly the ability to select which instances are terminated, and therefore ensure they are successfully drained, is vitally important. The two minute window provided natively by ASG’s in many cases is not sufficient enough time to drain a node to completion. Utilising plugins and a single workflow also allows the Nomad Autoscaler to work across providers in a common interface. It also allows us to support providers which may not include/utilise native autoscaling features, such as this [Openstack Senlin](https://github.com/dkt26111/nomad-senlin-autoscaler) plugin.

PR #428, opened 2021-03-20
Check for nil result to prevent panic

lgfa29:
In some situations (such as agent shutdown) the policy monitor might return a `nil` error, causing a panic.


cgbaker (approved):
👍

PR #435, opened 2021-03-26
scaleutils: add nodeselector package to select nodes suitable for termination

### LINKED ISSUE #417: feature: allow user to specify the strategy used for node identification during scale-in
When performing cluster scale in actions, nodes are identified for termination by picking off the top of the list as returned by [/v1/nodes](https://www.nomadproject.io/api-docs/nodes#list-nodes). This works fine for nodes which are running stateless and redundant jobs. This method, however, begins to cause issues when dealing with nodes which can be running stateless jobs, long lived batch jobs which are required to complete, or jobs that are simply classed as being important and ideally should not be interrupted.

The Nomad Autoscaler scaleutils package, which performs node selection for scale in has a basic concept of node selection strategy. This strategy can be used to alter the way in which nodes are identified to be appropriate for scaling in. It could therefore be beneficial to extend this functionality and provide a configuration mechanism to operators to alter this behaviour as desired.

#### Back of Slack Ideas
Some of the initial implementation thoughts we have had to aid further discussions. These are by no means fully formed.

Configuration would be exposed by a key within the policy target config mapping such as `node_scale_in_ident_strategy` whose value marks the strategy to use. The value of this could be a single strategy, or a list of strategies which forms a priority list. The former option is easier, the latter more complex; both open to discussion.

The available strategies could include:
- `no_allocs` whereby only nodes with 0 running allocs are eligible for identification
- `least_allocs` whereby the nodes with the least allocs are chosen
- `newest` where the newest nodes are picked
- `oldest` where the oldest nodes in the pool are identified for removal

jrasell:
closes #417 

This work allows operators to specify the strategy which is used to select nodes for termination during a scale in action. Up until now, the Nomad Autoscaler would simply pick nodes from the top of the list as returned from the Nomad node list API endpoint. This method is still included and is now known as `newest_create_index`.

In order to explicitly configure the selector behaviour, a new policy target configuration option `node_selector_strategy` has been made available. This config option is parsed by the scaleutils helper and therefore is available to all cluster target plugins which utilise it.

In the absence of an operator configured option, the value defaults to `least_busy`. This strategy calculates the percentage allocation of each node in the pool, selecting those which have the lowest value and therefore are allocated the least amount of resources. The resources are CPU and memory. 

The second selector strategy available is `empty`. This strategy will only select nodes for termination that do not have non-terminal allocations. This is ideal for worker pools running batch workloads which should not be interrupted or migrated. 

Admin notes:

- Documentation on the Nomad site will be required as a follow up step.
- To those internally testing this, I have some new TF I have been working on that will allow easier testing of different architectures such as this example. I will share this privately for now as it is very alpha.
- This PR only implements a single strategy whereas the original issue states several. I will raise follow up issues to individually track additional selector strategies. This is ideal candidates for first-issues and will be marked as such.


sdk/helper/scaleutils/nodeselector/defualt.go
```diff
@@ -0,0 +1,36 @@
...
+	return sdk.TargetNodeSelectorStrategyDefault
+}
+
+// Select satisfies the Select function on the ClusterScaleInNodeSelector
+// interface.
+func (d *defaultClusterScaleInNodeSelector) Select(nodes []*api.NodeListStub, num int) []*api.NodeListStub {
+
+	out := make([]*api.NodeListStub, num)
+
+	for i := 0; i <= num-1; i++ {
```
cgbaker, line 32: what if `len(nodes) < num`?

not sure if earlier code protects, but still probably worth handling.

sdk/helper/scaleutils/nodeselector/defualt_test.go
```diff
@@ -0,0 +1,40 @@
+package nodeselector
+
+import (
+	"testing"
+
+	"github.com/hashicorp/nomad/api"
+	"github.com/stretchr/testify/assert"
+)
+
+func Test_newestCreateIndexClusterScaleInNodeSelectorName(t *testing.T) {
```
cgbaker, line 10: if this is actually a test for `newest create index`, maybe we should go ahead and call the strategy that?

sdk/helper/scaleutils/nodeselector/defualt_test.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func Test_newestCreateIndexClusterScaleInNodeSelector_Select(t *testing.T) {
+	testCases := []struct {
+		inputNodes     []*api.NodeListStub
+		inputNum       int
+		expectedOutput []*api.NodeListStub
+		name           string
+	}{
+		{
```
cgbaker, line 21: maybe worth adding a test for `len(inputNodes) < inputNum`

sdk/helper/scaleutils/nodeselector/empty_nodes.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, alloc := range allocs {
+		if !allocClientTerminalStatus(alloc) || !allocServerTerminalStatus(alloc) {
+			e.log.Debug("selector skipped node with running allocs", "node_id", id)
+			return true
+		}
+	}
+	return false
+}
+
+// TODO (jrasell) remove once Nomad#10230 is released.
```
cgbaker, line 77: is release necessary or is merged sufficient?

sdk/helper/scaleutils/nodeselector/empty_nodes.go
```diff
@@ -0,0 +1,95 @@
...
+
+	allocs, _, err := e.client.Nodes().Allocations(id, nil)
+	if err != nil {
+		e.log.Error("failed to detail node allocs", "error", err)
+		return true
+	}
+
+	for _, alloc := range allocs {
+		if !allocClientTerminalStatus(alloc) || !allocServerTerminalStatus(alloc) {
+			e.log.Debug("selector skipped node with running allocs", "node_id", id)
```
cgbaker, line 70:
```suggestion
			e.log.Debug("selector skipped node with non-terminal allocs", "node_id", id)
```

sdk/helper/scaleutils/nodeselector/selector_test.go
```diff
@@ -0,0 +1,60 @@
...
+			inputCfg: map[string]string{
+				"node_selector_strategy": "empty_nodes",
+			},
+			expectedSelectorName: "empty_nodes",
+			expectedOutputError:  nil,
+			name:                 "empty nodes configured",
+		},
+		{
+			inputCfg: map[string]string{
+				"node_selector_strategy": "aliens",
```
cgbaker, line 41: 👽 👽 👽

cgbaker (approved):
one array bounds question and a few comments 👍

sdk/helper/scaleutils/nodeselector/defualt_test.go
jrasell: I should rename this test; the strategy itself makes no effort to ensure the description is correct which I why I used default. I just missed this. thanks.

sdk/helper/scaleutils/nodeselector/empty_nodes.go
jrasell: This depends if we are happy to upgrade the API dependancy to main, rather than using the release SHA?

sdk/helper/scaleutils/nodeselector/defualt.go
jrasell: Good shout. It's protected from the current caller, but this doesn't protect future implementations.

sdk/helper/scaleutils/nodeselector/empty_nodes.go
lgfa29: Personally I wouldn't mind using a non-release SHA if there's something useful to us 👍

lgfa29 (approved):
Minor comments, but it seems to be working 🎉

sdk/helper/scaleutils/nodeselector/least_busy.go
```diff
@@ -0,0 +1,199 @@
...
+		// allow for network hiccups.
+		resource, err := l.computeNodeResources(node)
+		if err != nil {
+			l.log.Error("failed to calculate node resource usage", "error", err)
+			return nil
+		}
+
+		// Set the index based on the node list position and add this entry to
+		// the heap. We do not need to use Push() and this is more efficient.
+		resource.index = i
```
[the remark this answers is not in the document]
jrasell: Nah not missing; I need to remove this as it is a remnant of code I previously added and removed.

sdk/helper/scaleutils/nodeselector/least_busy.go
```diff
@@ -0,0 +1,199 @@
...
+	// provides an easy way to pop entries in the way we need to return a list
+	// to the caller.
+	node *api.NodeListStub
+
+	// index tracks the index on the heap that this entry currently occupies.
+	index int
+}
+
+// nodeResourceConsumption is our container/heap implementation.
+type nodeResourceConsumption []*nodeResourceStats
```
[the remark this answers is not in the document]
jrasell: I don't really have a strong opinion either way. That being said, I don't see a possibility for nil pointer panics in the current code and from other examples I have seen using a pointer seems common place.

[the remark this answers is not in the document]
lgfa29: It doesn't happen right now because you don't use `heap.Push`, but as soon as you do you would get a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x12cb98f]

goroutine 82 [running]:
github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodeselector.nodeResourceConsumption.Less(...)
        /Users/laoqui/go/src/github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodeselector/least_busy.go:170
container/heap.down(0x13f5ea0, 0xc00000e040, 0x31, 0x64, 0x12ee0a0)
        /nix/store/iadyfrad1x7c05frqsqgw60z6mkjf9wl-go-1.14.8/share/go/src/container/heap/heap.go:112 +0xd3
container/heap.Init(0x13f5ea0, 0xc00000e040)
        /nix/store/iadyfrad1x7c05frqsqgw60z6mkjf9wl-go-1.14.8/share/go/src/container/heap/heap.go:46 +0x78
github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodeselector.BenchmarkSelectHeap2(0xc000014380)
        /Users/laoqui/go/src/github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodeselector/least_busy_test.go:261 +0x1e6
testing.(*B).runN(0xc000014380, 0x64)
        /nix/store/iadyfrad1x7c05frqsqgw60z6mkjf9wl-go-1.14.8/share/go/src/testing/benchmark.go:191 +0xe8
testing.(*B).launch(0xc000014380)
        /nix/store/iadyfrad1x7c05frqsqgw60z6mkjf9wl-go-1.14.8/share/go/src/testing/benchmark.go:321 +0xea
created by testing.(*B).doBench
        /nix/store/iadyfrad1x7c05frqsqgw60z6mkjf9wl-go-1.14.8/share/go/src/testing/benchmark.go:276 +0x55
exit status 2
FAIL    github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodeselector        12.671s
FAIL
```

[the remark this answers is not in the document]
jrasell: Resolving this issue. After some internal discussion and benchmarking we are moving from heap to sort.

PR #440, opened 2021-04-01
makefile: add targets for fixed-value and pass-through plugins.

jrasell:
I will raise a followup issue to improve this process; ideally in a way that just iterates the directory rather than requiring manual updates to the makefile.


cgbaker (approved):
👍

PR #441, opened 2021-04-01
policyeval: log error messages with error key.

jrasell:
This ensures consistency across the codebase and removes the use of what is a Golang specific style making log aggregation and parsing easier.


cgbaker (approved):
👍

PR #442, opened 2021-04-01
Make `query` optional inside checks

lgfa29:
Some strategy plugins don't require an APM query result (for example, `fixed-value`), so the PRs modifies the scaling policy validation rules to allow checks without a `query`.

### DIFF
--- a/policy/nomad/validate_horizontal.go
+++ b/policy/nomad/validate_horizontal.go
@@ -8,6 +8,7 @@ import (
 	"github.com/hashicorp/nomad/api"
 )
 
+
 func validateHorizontalPolicy(policy *api.ScalingPolicy) error {
 	var result *multierror.Error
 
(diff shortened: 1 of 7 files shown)


policy/nomad/validate_horizontal.go
```diff
@@ -49,14 +49,7 @@ func validateChecksHorizontal(in map[string]interface{}, path string) error {
...
-	// Validate Query.
-	//   1. Query key must exist.
-	if _, ok := c[keyQuery]; !ok {
-		result = multierror.Append(result, fmt.Errorf("%s.%s is missing", path, keyQuery))
-	}
-
-	return result.ErrorOrNil()
+func validateCheckHorizontal(c map[string]interface{}, path string, label string) error {
+	// Nothing to check at this time
+	return nil
```
lgfa29, line 54: We decided to keep this function so we have the structure in place in case we need this type of validation in the future.

cgbaker (approved):
oof, lotta work for such a simple goal.

looks good to me!

PR #443, opened 2021-04-01
Return no-op scaling action when there are no metrics

lgfa29:
With [`query` being optional](#442), there's a possibility that strategy plugins are called with empty metric results. So return a no-op scaling action.


cgbaker (approved):
👍 cool.

PR #450, opened 2021-04-05
Fix `empty` node selector strategy

lgfa29:
The `empty` node selector strategy had a wrong check that ended causing the opposite of the expected result. It would also fail to scale-in if system jobs were deployed.

This PR fixes the check and also ignores system job allocs when using the `empty` strategy. It also adds some basic tests using test-fixtures retrieved from a live Nomad cluster.


jrasell (approved):
LGTM, thanks!

I have not tested it E2E, but everything looks great.

PR #452, opened 2021-04-06
ci: lint test file and fix problems

jrasell:


agent/agent_test.go
```diff
@@ -27,7 +28,7 @@ func TestAgent_generateNomadClient(t *testing.T) {
 				},
 			},
 			expectedOutputEr: errors.New(`failed to instantiate Nomad client: invalid address '	': parse "\t": net/url: invalid control character in URL`),
-			name:             "invalid input Nomad address",
+			name:             "invalid input Nomad address", //nolint
```
jrasell, line 31: this shows as an error I think because of the line above, but is correctly formatted.

cgbaker (approved):
👍

PR #456, opened 2021-04-06
Upgrade policy file monitoring log to info

lgfa29:
When running the Autoscaler at `INFO` level (the default), there's no indication that a policy file has been properly loaded, which is a useful information to have early-on.


cgbaker (approved):
lgtm

jrasell (approved):
👍🏻

PR #458, opened 2021-04-08
add consul client to aget

cgbaker:
configuration, parsing and creation of a consul client for use by the autoscaler for... whatever we decide to use it for 🤷


agent/agent.go
```diff
@@ -165,6 +172,29 @@ func (a *Agent) generateNomadClient() error {
...
+	client, err := consulapi.NewClient(cfg)
+	if err != nil {
+		return fmt.Errorf("failed to instantiate Consul client: %v", err)
+	}
+	a.consulClient = client
+
+	return nil
+}
+
+
```
jrasell, line 197:
```suggestion
```

agent/config/config.go
```diff
@@ -149,6 +157,169 @@ type Nomad struct {
...
+// Consul contains the configuration information necessary to
+// communicate with a Consul server.
+type Consul struct {
+	// Addr is the HTTP endpoint address of the local Consul agent
+	//
+	// Uses Consul's default and env var.
+	Addr string `hcl:"address"`
+
+	// TimeoutHCL is used by Consul HTTP Client
+	TimeoutHCL string        `hcl:"timeout" json:"-"`
```
jrasell, line 169:
```suggestion
	TimeoutHCL string `hcl:"timeout" json:"-"`
```

agent/config/config.go
```diff
@@ -149,6 +157,169 @@ type Nomad struct {
...
+
+func (c *Consul) MergeWithDefault() (*consulapi.Config, error) {
+	if c == nil {
+		return nil, nil
+	}
+
+	cfg := consulapi.DefaultConfig()
+
+	if c.TimeoutHCL != "" {
+	}
```
jrasell, line 217: Does this need to be removed? I notice the same conditional is below (L225) and parses the duration.

jrasell (changes requested):
Looking good.

Couple of lint failures and a question about a conditional which I think can be removed. Currently imports in the repo are only split by stdlib / non-stdlib and then sorted within these two sections.

agent/config/config.go
cgbaker: oops, leftover copy-pasta 🤦

lgfa29 (approved):
LGTM.

We tend to add CLI flags to all of our configs (like [Nomad](https://github.com/hashicorp/nomad-autoscaler/blob/main/command/agent.go#L391-L402)), but this can come in a future PR.

jrasell (approved):
LGTM

PR #460, opened 2021-04-13
new policy.Source for filtering policies for HA capability

cgbaker:
this provides a new `policy.Source`, which accepts a policy filter and an upstream `policy.Source` and applies the filter to the results of `MonitorIDs` on the upstream.

### DIFF
--- a/policy/ha/filter.go
+++ b/policy/ha/filter.go
@@ -14,7 +14,7 @@ type MonitorFilterRequest struct {
 // PolicyFilter defines the interface for policy filters
 // used by the autoscaler's HA capability.
 type PolicyFilter interface {
-	// MonitorFilter accepts a context and a channel for informing the caller
+	// MonitorFilterUpdates accepts a context and a channel for informing the caller
 	// of asynchronous updates to the underlying filter.
 	MonitorFilterUpdates(ctx context.Context, req MonitorFilterRequest)
 	// ReloadFilterMonitor indicates that the filter should be reloaded
--- a/policy/ha/filter.go
+++ b/policy/ha/filter.go
@@ -1,0 +1,27 @@
+package ha
+
+import (
+	"context"
+
+	"github.com/hashicorp/nomad-autoscaler/policy"
+)
+
+type MonitorFilterRequest struct {
+	ErrCh    chan<- error
+	UpdateCh chan<- struct{}
+}
+
+// PolicyFilter defines the interface for policy filters
+// used by the autoscaler's HA capability.
+type PolicyFilter interface {
+	// MonitorFilter accepts a context and a channel for informing the caller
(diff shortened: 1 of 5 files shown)


policy/ha/filter.go
```diff
@@ -0,0 +1,27 @@
...
+
+type MonitorFilterRequest struct {
+	ErrCh    chan<- error
+	UpdateCh chan<- struct{}
+}
+
+// PolicyFilter defines the interface for policy filters
+// used by the autoscaler's HA capability.
+type PolicyFilter interface {
+	// MonitorFilter accepts a context and a channel for informing the caller
```
jrasell, line 17:
```suggestion
	// MonitorFilterUpdates accepts a context and a channel for informing the caller
```

policy/ha/source.go
```diff
@@ -0,0 +1,120 @@
...
+			policyIDs = newUpstreamIDs.IDs
+			haveFirstPolicies = true
+
+		case <-filterUpdateCh:
+			haveFirstFilter = true
+		}
+
+		if !haveFirstPolicies || !haveFirstFilter {
+			continue
+		}
```
jrasell, line 91: I am curious about this as bit, as I would have expected that we would only get this far if either `filterUpdateCh` or `upstreamPolicyCh` had fired so curious what I am missing.

policy/ha/source_test.go
```diff
@@ -0,0 +1,198 @@
...
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/stretchr/testify/require"
+
+	"github.com/hashicorp/nomad-autoscaler/policy"
```
jrasell, line 12:
```suggestion
	"github.com/hashicorp/go-hclog"
	"github.com/hashicorp/nomad-autoscaler/policy"
	"github.com/stretchr/testify/require"
```

policy/source.go
```diff
@@ -65,6 +67,9 @@ const (
 
 	// SourceNameFile is the source for policies that are loaded from disk.
 	SourceNameFile SourceName = "file"
+
+	// SourceNameFiltered is the source for HA policy sources
```
jrasell, line 71:
```suggestion
	// SourceNameHA is the source for HA policy sources
```

jrasell (approved):
LGTM

A couple of minor suggestions and a question for my clarification but nothing blocking.

policy/ha/source.go
cgbaker: we'll only get this far if one of them had fired... this checks that both of them have fired. i don't know why i wrote this instead of the equivalent `!(haveFirstPolicies && haveFirstFilter)`

lgfa29 (approved):
💯

Issue #459, opened 2021-04-12
nodepool: add Datacenter nodepool configuration option


jrasell:
Horizontal cluster scaling currently relies on utilising the client `NodeClass` parameter in order to group clients into logical pools of scalable resources. The `Datacenter` parameter is another way to perform this grouping, and in some cases is preferred as it is a require parameter on both the client and within the job specification.

It would therefore be beneficial to add a new `ClusterNodePoolIdentifier` which utilises `Datacenter` as the client pool grouper. Besides the new implementation, a small amount of logic will be involved to extract the config to identify which method is desired.

cc @angrycub

PR #462, opened 2021-04-15
sdk/helper: create error pkg to help with multierrors.

jrasell:
Promote the multierror formatting used in scaleutils. This will be useful for work I am doing within #459 and will allow us to consistently format our errors in a nice way.


cgbaker (approved):
👍

PR #463, opened 2021-04-15
chore: don't use fmt.Errorf when not formatting the error.

jrasell:
Using the stdlib `errors.New` call to create errors that are not formatted is quicker and more memory efficient. This change therefore moves all non-formatted error creations to use this func.

Just imagine what we can achieve with the 80ns we save each time.

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"errors"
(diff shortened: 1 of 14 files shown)


cgbaker (approved):
:til: 
would be nice if the linter would remind us of this.

Issue #473, opened 2021-04-21
test/aws-asg: improve config testing using EC2 metadata service mock


jrasell:
The outcome of the `SetConfig` function within the AWS ASG target plugin depends on whether the application is running on an EC2 instance or not. This is because the AWS client setup utilises the EC2 metadata service, if possible. Testing this function therefore requires mocking this metadata service to fully and reliably test.

Potential applications and libraries which could provide the required functionality include [`aws/amazon-ec2-metadata-mock`](https://github.com/aws/amazon-ec2-metadata-mock) and [`nytimes/mock-ec2-metadata`](https://github.com/nytimes/mock-ec2-metadata).

PR #474, opened 2021-04-21
plugins/target/aws-asg: correctly order region config setting.

### LINKED ISSUE #472: `aws_region` not respected
## The Bug

Upon setting the `aws_region` config setting in the autoscaler config, we noticed that it didn't actually seem to be respected by the autoscaler. Instead it seemed to default to the region that the instance was currently deployed in. After further debugging, we also found that we could affect the region by setting `AWS_DEFAULT_REGION` in the autoscaler's environment. 

## Context

We are setting three different plugin instances, all using `aws-asg` as driver, and each with a different `aws_region` set. Only the plugin dispensed with the `aws_region` set to it's home region ended up working. The other two fail with the following:
```log
failed to describe AWS Autoscaling Group: expected 1 Autoscaling Group, got 0
```

We also did some source diving and patched in a few extra logger statements so that we could determine that the region is not being set properly. Here is some output from our custom logger to show more clearly the issue:

```log
Dispense plugin: client-us-east-2-c5-4xlarge
external_plugin.aws-asg: ASG client region: eu-central-1
```

Here you can see we are dispensing plugin intended for us-east-2 region, with it's `aws_region` being likewise set, yet logging the actual value of the region at that point in the code shows it is still using eu-cental-1 regardless of the config setting.

jrasell:
A user specific region configuration option should take precedence
over other options. This change therefore updates the flow in
which the AWS config is populated. In situation where AWS and the
operator do not provide an option, the default is applied.

The change itself is hard to automatically test at the moment as
the behaviour changes depending whether you have the EC2 metadata
service available. The test added is aimed at situations where the
plugin is not running with the EC2 metadata service available.

#473 has been raised to track better testing where the EC2 metadata
service is available.

closes #472

### DIFF
--- a/plugins/builtin/target/aws-asg/plugin/aws_test.go
+++ b/plugins/builtin/target/aws-asg/plugin/aws_test.go
@@ -2,14 +2,55 @@ package plugin
 
 import (
 	"errors"
+	"os"
 	"testing"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/autoscaling"
+	hclog "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils"
 	"github.com/hashicorp/nomad/api"
 	"github.com/stretchr/testify/assert"
 )
+
+func TestTargetPlugin_SetConfig_NonAWS(t *testing.T) {
+	if os.Getenv("NOMAD_AUTOSCALER_AWS") != "" {
+		t.Skip("skipping aws-asg non-aws environment test, NOMAD_AUTOSCALER_AWS set")
+	}
+
+	testCases := []struct {
+		inputCfg          map[string]string
+		expectedOutputErr error
+		name              string
+	}{
+		{
+			inputCfg:          map[string]string{},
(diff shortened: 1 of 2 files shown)


plugins/builtin/target/aws-asg/plugin/aws_test.go
```diff
@@ -2,15 +2,56 @@ package plugin
...
+	hclog "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils"
 	"github.com/hashicorp/nomad/api"
 	"github.com/stretchr/testify/assert"
 )
 
+func TestTargetPlugin_SetConfig_NonAWS(t *testing.T) {
+	if os.Getenv("NOMAD_AUTOSCALER_AWS") != "" {
+		t.Skip("skipping aws-asg non-aws environment test, NOMAD_AUTOSCALER_AWS set")
+	}
```
jrasell, line 19: this test will only pass where the EC2 metadata is not available; therefore I have added this placeholder as a reminder when adding additional tests that we could use an env var to control which get run depending on the situation.

lgfa29 (approved):
Oof, nice catch!

PR #483, opened 2021-04-30
threshold strategy plugin

### LINKED ISSUE #438: `threshold` strategy
The `threshold` strategy will allow for different scaling actions based on a set of tiers defined by upper and lower bound metric values. Each tier can define a scaling action to be taken when the metric enters the tier.

The type of scaling action can be one of the following:

* **Percentage** will modify the current count by a percent factor, rounding up the result to keep the maximum amount of infrastructure remaining. Defined by using the `percentage` property.
* **Instances** will add (positive value) or remove (negative value) the number of instances defined in the policy. Defined by using the `delta` property.
* **Absolute** will set the value specified in the policy directly as the new count. Defined by using the `value` property.

### Example
```hcl
scaling {
  ...
  check "scale_out_high_threshold_zone" {
    source = "prometheus"
    query  = "avg(cpu_usage_pct)[5m]"

    strategy "threshold" {
      # If our metric value is above 70 (inc.).
      lower_bound = 70
      # Add 30 new instances.
      delta       = 30
    }
  }

  check "scale_out_low_threshold_zone" {
    source = "prometheus"
    query  = "avg(cpu_usage_pct)[5m]"

    strategy "threshold" {
      # If our metric value is between 60 (inc.) and 70 (excl.).
      lower_bound = 60
      upper_bound = 70
      # Increase count by 10 percent.
      percentage  = 10
    }
  }

  check "scale_in_low_threshold_zone" {
    source = "prometheus"
    query  = "avg(cpu_usage_pct)[5m]"

    strategy "threshold" {
      # If our metric value is between 10 (inc.) and 30 (excl.).
      lower_bound = 10
      upper_bound = 30
      # Reduce count by 30%.
      percentage  = -30
    }
  }

  check "scale_in_high_threshold_zone" {
    source = "prometheus"
    query  = "avg(cpu_usage_pct)[5m]"

    strategy "threshold" {
      # If our metric value is below 10.
      upper_bound = 10
      # Set count to 2.
      value       = 2
    }
  }
}
```

jazzyfresh:
Closes #438 

Should be merged after #484 to avoid a panic.

### DIFF
--- a/plugins/builtin/strategy/threshold/plugin/plugin.go
+++ b/plugins/builtin/strategy/threshold/plugin/plugin.go
@@ -1,0 +1,179 @@
+package plugin
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad-autoscaler/plugins"
+	"github.com/hashicorp/nomad-autoscaler/plugins/base"
+	"github.com/hashicorp/nomad-autoscaler/plugins/strategy"
+	"github.com/hashicorp/nomad-autoscaler/sdk"
+)
+
+const (
+	// pluginName is the unique name of the this plugin amongst strategy
+	// plugins.
+	pluginName = "threshold"
+
+	// These are the keys read from the RunRequest.Config map.
+	runConfigKeyUpperBound    = "upper-bound"
+	runConfigKeyLowerBound = "lower-bound"
+	runConfigKeyDelta = "delta"
+)
+
+var (
+	PluginID = plugins.PluginID{
+		Name:       pluginName,
+		PluginType: sdk.PluginTypeStrategy,
+	}
+
+	PluginConfig = &plugins.InternalPluginConfig{
+		Factory: func(l hclog.Logger) interface{} { return NewThresholdPlugin(l) },
+	}
+
+	pluginInfo = &base.PluginInfo{
+		Name:       pluginName,
+		PluginType: sdk.PluginTypeStrategy,
+	}
+)
+
+// Assert that StrategyPlugin meets the strategy.Strategy interface.
+var _ strategy.Strategy = (*StrategyPlugin)(nil)
+
+// StrategyPlugin is the Threshold implementation of the strategy.Strategy
+// interface.
+type StrategyPlugin struct {
+	config map[string]string
+	logger hclog.Logger
+}
+
+// NewTargetValuePlugin returns the Threshold implementation of the
+// strategy.Strategy interface.
+func NewThresholdPlugin(log hclog.Logger) strategy.Strategy {
+	return &StrategyPlugin{
+		logger: log,
+	}
+}
+
+// SetConfig satisfies the SetConfig function on the base.Base interface.
+func (s *StrategyPlugin) SetConfig(config map[string]string) error {
+	s.config = config
+	return nil
+}
+
+// PluginInfo satisfies the PluginInfo function on the base.Base interface.
+func (s *StrategyPlugin) PluginInfo() (*base.PluginInfo, error) {
+	return pluginInfo, nil
+}
+
+// Run satisfies the Run function on the strategy.Strategy interface.
+func (s *StrategyPlugin) Run(eval *sdk.ScalingCheckEvaluation, count int64) (*sdk.ScalingCheckEvaluation, error) {
+
+	// Read and parse threshold bounds from req.Config.
+	upperbound := eval.Check.Strategy.Config[runConfigKeyUpperBound]
+	lowerbound := eval.Check.Strategy.Config[runConfigKeyLowerBound]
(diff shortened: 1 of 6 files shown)


plugins/builtin/strategy/threshold/plugin/plugin.go
```diff
@@ -0,0 +1,296 @@
...
+// runPercentage returns the next count for a percentage check.
+func runPercentage(count int64, pct float64) int64 {
+	newCount := float64(count) * (1 + pct/100)
+	return int64(math.Ceil(newCount))
+}
+
+// runValue returns the next count for a value check.
+func runValue(count int64, v float64) int64 {
+	return int64(v)
+}
```
jrasell, line 283: It doesn't seem that the `count` argument is used; is this expected? If so can we remove it from the function signature?

plugins/builtin/strategy/threshold/plugin/plugin.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+	if nonEmpty == 0 {
+		return nil, fmt.Errorf("missing required field, must have either %q, %q or %q", runConfigKeyDelta, runConfigKeyPercentage, runConfigKeyValue)
+	}
+	if nonEmpty != 1 {
+		return nil, fmt.Errorf("only one of %q, %q or %q must be provided", runConfigKeyDelta, runConfigKeyPercentage, runConfigKeyValue)
+	}
+
+	if deltaStr != "" {
+		// Delta must be an interger.
```
jrasell, line 193:
```suggestion
		// Delta must be an integer.
```

plugins/builtin/strategy/threshold/plugin/plugin.go
```diff
@@ -0,0 +1,296 @@
...
+// Assert that StrategyPlugin meets the strategy.Strategy interface.
+var _ strategy.Strategy = (*StrategyPlugin)(nil)
+
+// StrategyPlugin is the Threshold implementation of the strategy.Strategy
+// interface.
+type StrategyPlugin struct {
+	logger hclog.Logger
+}
+
+// NewTargetValuePlugin returns the Threshold implementation of the
```
jrasell, line 67:
```suggestion
// NewThresholdPlugin returns the Threshold implementation of the
```

plugins/builtin/strategy/threshold/plugin/plugin.go
```diff
@@ -0,0 +1,296 @@
...
+		if s != "" {
+			nonEmpty++
+		}
+	}
+	if nonEmpty == 0 {
+		return nil, fmt.Errorf("missing required field, must have either %q, %q or %q", runConfigKeyDelta, runConfigKeyPercentage, runConfigKeyValue)
+	}
+	if nonEmpty != 1 {
+		return nil, fmt.Errorf("only one of %q, %q or %q must be provided", runConfigKeyDelta, runConfigKeyPercentage, runConfigKeyValue)
+	}
```
jrasell, line 190: I wonder if it is worth simplifying this into a single `!= 1` check rather than two checks to slightly change the error message grammar.

plugins/builtin/strategy/threshold/plugin/plugin.go
```diff
@@ -0,0 +1,296 @@
...
+	logger = logger.With("new_count", newCount)
+
+	// Identify the direction of scaling, and exit early if none.
+	eval.Action.Direction = calculateDirection(count, newCount)
+	if eval.Action.Direction == sdk.ScaleDirectionNone {
+		return eval, nil
+	}
+
+	logger = logger.With("direction", eval.Action.Direction)
+	logger.Trace("calculated scaling strategy results")
```
jrasell, line 130: Two sub-loggers get created to log a single trace line here. I wonder if it would be better to add `new_count` and `direction` onto the log line as inline contextual arguments like `logger.Trace("msg", "new_count", newCount)`?

jrasell (approved):
LGTM.

Couple of minor points but nothing blocking.

plugins/builtin/strategy/threshold/plugin/plugin.go
lgfa29: At some point there was a common function type for these three functions, but I thought it would couple them too much. I missed removing it here 👍

plugins/builtin/strategy/threshold/plugin/plugin.go
lgfa29: I think there used to be a log message between the two, but I guess I got rid of it at some point 😅

plugins/builtin/strategy/threshold/plugin/plugin.go
lgfa29: Hum...not sure if I follow. `== 0` is for the case where no action type is provided. `!= 1` is for when more than one is provided. I guess it could be `> 1` for clarity.

But the error message in both cases would be different.

PR #484, opened 2021-05-01
Fix to return `nil` instead of `none` action from strategies

lgfa29:
There is a small difference between a strategy determining that the count value shouldn't change as opposed to the strategy not being able to compute it.

In the first case, the new count is successfully calculated but it happens to be equal the current count. In the second case, the strategy is unable to determine what to do.

This difference is important because a `none` scaling direction takes precedence over a `down` action. Return `none` instead of `nil` can cause a `down` check to be ignored.


jrasell (approved):
LGTM pending fixing the broken tests.

PR #494, opened 2021-05-25
helper: add metric pkg to allow setting default labels.

jrasell:
When running in HA and therefore when each agent has a unique ID,
it will be useful to have this ID as a default label. This allows
easier understanding across telemetry, logs and service discovery
processes.

The new package wraps the metrics library calls, ensuring default
and additional labels are always added where required.

The new code isn't currently utilised, this work will be performed in
a follow up PR targeting the feature branch.

### DIFF
--- a/sdk/helper/metrics/metrics.go
+++ b/sdk/helper/metrics/metrics.go
@@ -1,0 +1,46 @@
+package metrics
+
+import (
+	"sync/atomic"
+	"time"
+
+	m "github.com/armon/go-metrics"
+)
+
+// defaultLabels are the label set that should be applied to every data point
+// emitted. Use an atomic value so that we protect against concurrent access in
+// situations where the default labels are being updated after telemetry
+// initialization.
+var defaultLabels atomic.Value
+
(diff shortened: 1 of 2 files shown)


sdk/helper/metrics/metrics.go
```diff
@@ -0,0 +1,46 @@
...
+// importing both packages when emitting metrics.
+type Label = m.Label
+
+// SetDefaultLabels sets defaultLabels with the configured default set of
+// labels.
+func SetDefaultLabels(labels []Label) { defaultLabels.Store(labels) }
+
+// SetGauge wraps m.SetGaugeWithLabels and sets the default labels on the
+// emitted metric.
+func SetGauge(key []string, val float32) {
```
lgfa29, line 26: Would it be worth adding a `SetGaugeWithLabels`?

lgfa29 (approved):
Nice!

sdk/helper/metrics/metrics.go
jrasell: I created the functions based on the autoscaler's current usage of the go-metrics package and was of the opinion we could just add new ones as/when we need them.

PR #502, opened 2021-06-15
sdk: add new hash pkg with ring member and hash implementation.

jrasell:


lgfa29 (approved):
The code LGTM, but I am curious if we actually need these interfaces since only the `ha` package would use them.

jrasell:
> The code LGTM, but I am curious if we actually need these interfaces since only the `ha` package would use them.

I find they help keep the code structured and isolated even if they are simple and only to be used by a single caller. It also helps me personally design and write the code in a better style. We can revisit before merging to main if we find there are overriding factors why we shouldn't use these.

PR #504, opened 2021-06-21
ha: add policy filter interface and implementation.

jrasell:


policy/filter.go
```diff
@@ -0,0 +1,155 @@
...
+// Therefore, we'll continue to use the existing name.
+func (fs *FilteredSource) Name() SourceName {
+	return fs.upstreamSource.Name()
+}
+
+// ReloadIDsMonitor implements policy.Source by calling the appropriate
+// reload method on the underlying policy source and filter.
+func (fs *FilteredSource) ReloadIDsMonitor() {
+	fs.upstreamSource.ReloadIDsMonitor()
+	fs.policyFilter.ReloadFilterMonitor()
```
lgfa29, line 154: Do we need to reset `haveFirstPolicies` and `haveFirstFilter` here? I'm not sure how they relate, but I'm wondering if it could cause the unfiltered policies to be sent on reload.

policy/filter_test.go
```diff
@@ -0,0 +1,196 @@
...
+			IDs:    append(expected, unexpected...),
+			Source: "test",
+		}
+	}()
+
+	// should not receive a message before BOTH filter and upstream have messaged
+	select {
+	case <-outputCh:
+		require.Fail("received message before filter sent message")
+	case <-time.After(2 * time.Second):
```
lgfa29, line 59: Maybe reduce these timeouts (or make it configurable on the number of tests running? I'm not even sure if that's possible 😅) 

2s doesn't seem like a lot but it could compound when running the full test suite.

lgfa29 (approved):
Looking good! I left some minor comments.

policy/filter_test.go
jrasell: The only way I can figure off the top of my head is to use env vars; but that in itself will likely get lost and be tricky/annoying to use in local setups. I'll make a note to keep this in mind in the future.

policy/filter.go
jrasell: We don't. Those two variables ensure that we have both members and policy responses available before performing the first set of assignments. The current reloading which does need broader work to improve, is a different starting point and has potential to change something like a TLS certificate only. If we were to reset these values, it would be possible policy changes are not processed because the pool membership has not changed, and therefore the blocking query not returned.

This potentially changes in the future when we enhance the reload capabilities; but that feels like a large piece of work outside this particular feature.

PR #510, opened 2021-07-13
ha: add Consul pool discovery and agent ID generation

jrasell:


consul/pool_discovery.go
```diff
@@ -0,0 +1,207 @@
...
+				return
+			}
+
+			if err != nil {
+				errCh <- err
+			} else {
+				q.WaitIndex = qm.LastIndex
+				resCh <- services
+			}
+		}()
```
lgfa29, line 61: Would it be better to have these run just once instead of recreating everything on each loop? So there would be one long-lived goroutine with a `for` loop and another loop surrounding the `select` below

consul/pool_discovery.go
```diff
@@ -0,0 +1,207 @@
...
+		return fmt.Errorf("failed to register service: %v", err)
+	}
+
+	// Retry and backoff when attempting to register the service check within
+	// Consul.
+	checkFunc := func() error { return c.client.Agent().CheckRegister(&checkReg) }
+
+	if err := backoff.Retry(checkFunc, generateRetryBackoff()); err != nil {
+		return fmt.Errorf("failed to register service check: %v", err)
+	}
```
lgfa29, line 167: Probably a minor thing, but these backoffs would hold a read lock for the Consul client even if not using it right? Maybe the lock and unluck should happen inside each function?

Though I don't see where the lock is being acquired for writing.

lgfa29 (approved):
Minor comments, but LGTM!

PR #513, opened 2021-07-15
HA: add initial consistent hashring implementation.

jrasell:
The ring focusses on consistency and doesn't include functionality
such as load balancing.

The consistent hashring is used for partitioning policies between
autoscaler agents. It keeps an internal tracking of load
distribution across known members, allowing telemetry to identify
inconsistency.

The current implementation is specific to Consul and the manner in
which agents are discovered. If we ever moved to an alternative
discovery mechanism, we would need some additional functions but
the logic is sound.

### DIFF
--- a/sdk/helper/hash/ring.go
+++ b/sdk/helper/hash/ring.go
@@ -1,0 +1,159 @@
+package hash
+
+import (
+	"sort"
+	"sync"
+
+	"github.com/hashicorp/nomad-autoscaler/policy"
+)
+
+// ConsistentRing is a basic implementation of a consistent hash ring providing
+// consistent assignment of keys to ring members.
+type ConsistentRing struct {
+	hasher Hasher
+
+	// mu should be used to ensure ring consistency when updating or
+	// identifying assignments. In the search for stronger consistency this is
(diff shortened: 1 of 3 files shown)


sdk/helper/hash/ring.go
```diff
@@ -0,0 +1,159 @@
...
+
+// getMemberAssignmentWithLock identifies which ring member the passed key
+// should be assigned to.
+func (cr *ConsistentRing) getMemberAssignmentWithLock(key string) RingMember {
+
+	// The fact we have to convert the policyID from a string to a byte array
+	// accounts for a large amount time comparable to the work being done. This
+	// is seen within the cpu profile for the benchmarks particularly when
+	// running GetAssignments().
+	hash := cr.hasher.Hash32([]byte(key))
```
jrasell, line 132: This single line contributes the large portion of time complexity; I have thoughts on how to improve this but it's certainly not something we need right away and would be quite a piece of work.

lgfa29 (approved):
Minor comments, but looking forward to seeing it all hooked-up 😄

PR #525, opened 2021-09-17
Allow disabling the nomad source

### LINKED ISSUE #524: Allow to disable the nomad source
Hi, I think it would be useful to to allow disabling the nomad source.

In our case, we want to deploy two instances of nomad-autoscaler, one dealing with allocation scaling and another with node pools autoscaling. We created several policies so each of them only has whats needed (the allocation one has scale over namespaces and the node autoscaler has permissions over nodes only). Right now as they always get the policies from nomad, the one dealing with node pools (that gets the policies from the filesystem) gets and error constantly because it's trying to get them.

```
[ERROR] policy_manager: failed to call the Nomad list policies API: Unexpected response code: 403 (Permission denied)
```

I think allowing to disable the sourcing of policies from nomad would be a nice addition.

jorgemarey:
Fixes #524.


jrasell:
Hi @jorgemarey and thanks so much for raising this PR. Apologies for the delay in reviewing, I am hoping we can take an initial look over the coming few weeks.

lgfa29:
Thank you for the PR @jorgemarey, it looks great!

But while I was reviewing it, I thought of a different approach that seems more flexible and would allow for new features to implemented on top of it. 

I'm going to close this without merging in favour of #544, but just to make it clear, there's nothing wrong with this implementation, and I was about to accept it until I thought about that alternate approach. Feel free to share your thoughts on that PR as well!

Thanks again for the contribution!

Issue #424, opened 2021-03-17
Scaling up from 0 not possible


dpogorzelski:
Hello,
please consider the following scenario: 
after the autoscaler scaled a GCP MIG to 0 nodes it won't be able to scale up again when a job is submitted because of:
```
2021-03-17T10:44:16.749Z [WARN]  policy_eval.worker: failed to run check: id=c912b170-5aa7-a36d-83fd-079f1fdb35b5 policy_id=323c8cc5-e238-f91a-7a22-96abdd2824f3 queue=cluster target=gce-mig err="failed to query source: no nodes identified within pool"
```
the policy looks like this:
```hcl
scaling "cluster_policy" {
  enabled = true
  min     = 0
  max     = 3

  policy {

    cooldown            = "1m"
    evaluation_interval = "1s"

    check "cpu_allocated_percentage" {
      query = "percentage-allocated_cpu"

      strategy "target-value" {
        target = 70
      }
    }

    check "mem_allocated_percentage" {
      query = "percentage-allocated_memory"

      strategy "target-value" {
        target = 70
      }
    }

    target "gce-mig" {
      project             = "foobar"
      region              = "europe-north1"
      mig_name            = "mymig"
      node_class          = "default"
      node_drain_deadline = "5m"
    }
  }
}
```
Any ideas on how to solve this? If scaling down to 0 is permitted then it makes sense it should also be possible to scale up from 0 :)

jrasell:
Hi @dpogorzelski and thanks for opening this. This is certainly a problem within the Nomad APM and how it should handle this situation.

jrasell:
Hi again @dpogorzelski. This issue has been discussed internally and we have identified https://github.com/hashicorp/nomad/issues/9046 as something which is precursor any updates within this repository. The issue is currently being looked into and we will be sure to keep these issues updates so progress can be tracked.

lgfa29:
Hi @dpogorzelski 👋 

After investigating this further I realized that there's really not good way to support scaling to zero clients with the Nomad APM....

#534 goes into more details and provide a fix to prevent the cluster from being stuck to zero. Feel free to add any comments or feedback on the approach I took there 🙂

dpogorzelski:
Hey @lgfa29 , thanks for the followup, I think this makes sense :)

lgfa29:
Nice! Feel free to reach out again if you have any other question 🙂

PR #534, opened 2021-11-11
Add policy mutators to prevent cluster scaling to zero with the Nomad APM

### LINKED ISSUE #530: Cluster autoscale fails to scale up from 0 nodes
**Background**
- Nomad version 1.x
- Nomad autoscaler version 2.1.0
- Target AWS ASG
- APM: Nomad

It seems its possible to scale down to zero nodes, but the autoscaler breaks at that point and can't scale back up again. Log messages like the following appear

```
2021-10-04T16:22:36.460Z [WARN]  policy_eval.worker: failed to run check: id=xxxxx policy_id=xxxxx queue=cluster target=aws-asg error="failed to query source: no nodes identified within pool"
```

**Desired outcomes**
If scale to zero isn't supported I suppose the a configuration like 

```
scaling "example" {
   min = 0
   max = x
   ....
}
```

Should probably fail fast, or the autoscaler should ignore min=0 and just scale down to 1. If scale to zero _is_ supported then something is broken I think. Perhaps its trying to query nomad APM to scale up but the query really needs to change when the node count == 0. If _any_ job is scheduled for the node_class we need to scale up, instead of trying to measure CPU and memory on nodes that don't exist

lgfa29:
The Nomad APM uses the Nomad API to retrieve information about allocations and nodes. Usually, the choice of metric source should not impact the choice of target defined in a policy, but when using the Nomad APM to perform cluster scaling to `0` presents there is a problem to scale back up since, once the cluster reaches zero clients, there's nothing to query for metrics. 

There is also no meaningful default value to return in this case. The metrics supported by the Nomad APM for cluster scaling are percentage of CPU and memory allocated. For zero clients these values are 0% and 100% at the same time, and different policy strategies would require interpreting in one or another. Or not mater at all!

For example, using the `target-value` strategy with `target = 70` (i.e., 70% CPU usage as target) would never scale back from zero since the next count to be calculated is also always zero (`next count = 0/70 * 0` or `next count = 100/70 * 0`)

The root of the problem is that the Nomad APM was never intended to be used in this scenario. Its goal is to provide a quick way to scale apps and clusters based on memory and CPU usage. More advanced scenarios, where these metrics are not useful, require a different APM. 

The [On-demand Batch Job Cluster Autoscaling tutorial](https://learn.hashicorp.com/tutorials/nomad/horizontal-cluster-scaling-on-demand-batch?in=nomad/autoscaler) provides an example on how to scale a Nomad cluster to zero and back up using the Prometheus APM.

There are a few possible solutions to avoid this problem.

### Validate policies for this condition

Policy validation is currently done by policy sources, but don't take plugin-specific aspects into consideration. Validating a policy to cover this scenario would be possible, but a failed validation would prevent the policy from being evaluated altogether. A log message is emitted in these situations, but a re easy to be missed by operators.

### Query servers for metrics

With no clients to be queried, servers present an option for metric source. But this setup requires significant additional logic since the `/v1/metrics` endpoint only returns values specific to the agent being queried. Iterating over all servers scraping their metrics using tools like Datadog or Prometheus is the recommended way to consume these values, and so in this scenario, a different APM plugin should be used.

Another potential problem is that, the Nomad Autoscaler, when running as an allocation within Nomad, may not have direct access to servers.

### Modify policies to never scale to zero

This option is similar to the validation approach, in which plugin-aware checks are performed in each policy, but with the advantage that, instead of never performing any evaluations at all, modifying the policy to change `min = 0` to `min = 1` would keep the policy active, but without ever reaching zero clients.

Given that expanding the Nomad APM plugin to query more metrics is not feasible, modifying the policies to prevent scaling to zero seems to be the best approach. 

This PR adds the concept of policy mutators that are used by policy handlers to modify incoming policies if necessary. If a policy is modified, a log line is emitted to notify operators:

```
2021-11-11T13:37:54.946-0500 [INFO]  file_policy_source: starting file policy monitor: file=bin/policies/cluster.hcl name=policy-test policy_id=9d121526-7660-dc19-463b-5b2fee1a6b6d
2021-11-11T13:37:54.946-0500 [INFO]  policy_manager.policy_handler: policy modified: policy_id=9d121526-7660-dc19-463b-5b2fee1a6b6d modification="min value set to 1 since scaling cluster to 0 is not supported by the Nomad APM"
```

Despite the potential for low visibility into catching this log line, mutators don't prevent the policy from being evaluated, so missing these lines should be safe.

Closes #530 #424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+ * policy: Prevent scaling cluster to zero when using the Nomad APM [[GH-534](https://github.com/hashicorp/nomad-autoscaler/pull/534)]
+
 BUG FIXES:
  * scaleutils: Fixed `least_busy` node selector on clusters running servers older than v1.0.0 [[GH-508](https://github.com/hashicorp/nomad-autoscaler/pull/508)]
 
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -36,6 +36,9 @@ type Handler struct {
 	// is responsible for.
 	policySource Source
 
+	// mutators is a list of mutations to apply to policies.
+	mutators []Mutator
+
 	// ticker controls the frequency the policy is sent for evaluation.
 	ticker *time.Ticker
 
@@ -68,11 +71,14 @@ func NewHandler(ID PolicyID, log hclog.Logger, pm *manager.PluginManager, ps Sou
 		log:           log.Named("policy_handler").With("policy_id", ID),
 		pluginManager: pm,
 		policySource:  ps,
-		ch:            make(chan sdk.ScalingPolicy),
-		errCh:         make(chan error),
-		doneCh:        make(chan struct{}),
-		cooldownCh:    make(chan time.Duration),
-		reloadCh:      make(chan struct{}),
+		mutators: []Mutator{
+			NomadAPMMutator{},
+		},
+		ch:         make(chan sdk.ScalingPolicy),
+		errCh:      make(chan error),
+		doneCh:     make(chan struct{}),
+		cooldownCh: make(chan time.Duration),
+		reloadCh:   make(chan struct{}),
 	}
 }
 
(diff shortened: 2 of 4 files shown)


jrasell (approved):
Wonderful change and comments, thanks!

PR #535, opened 2021-11-12
Allow datacenter and node class filtering

### LINKED ISSUE #531: aws-asg: allow filtering by node_class and datacenter
I was looking at the documentation for `aws-asg` https://www.nomadproject.io/docs/autoscaling/plugins/target/aws-asg
and I noticed that when selecting Nodes for scale-in events, the autoscaler can only select by `node_class` or by `datacenter` and not both.

Our cloud setup is such that we have multiple AWS regions, with the same ASG setup terraformed in each. For example, we may have:
```
datacenter "us-east-1" that exists in AWS region us-east-1
asg 1: node_class_a
asg 2: node_class_b

datacenter "us-west-1" that exists in AWS region us-west-1
asg 1: node_class_a
asg 2: node_class_b

etc etc
```

When a policy goes to scale in and select nodes to do so for the target ASG using `node_class`, it may select nodes from a different datacenter, fail to find the instance in the ASG and so fail to scale in, eventually hitting the `delivery_limit` and failing for good. Likewise, if you were to filter by `datacenter` instead then you may end up selecting Nodes that belong to a different ASG and also fail for the same reason.

Do you see value in being able to allow filtering by both `datacenter` and `node_class` rather than making them mutually exclusive in the config? That way you can be sure you are selecting the correct Node for a given ASG.

(Off the back of this discussion thread https://discuss.hashicorp.com/t/nomad-autoscaler-aws-asg-filtering-by-node-class-and-datacenter/30358)

lgfa29:
The initial implementation for the node filtering would only take either node class or datacenter as the filter input. But as explained in #531, there are valid use cases where defining both is required in order to properly match Nomad clients with cloud instances.

This PR creates a new `ClusterNodePoolIdentifier` called `combinedClusterPoolIdentifier` that can be used to merge other identifiers, applying a simple `and` or `or` logic. 

When both `node_class` and `datacenter` are present in a scaling policy, the Autoscaler will use this combined identifier to merge the existing `NewNodeDatacenterPoolIdentifier` and `NewNodeDatacenterPoolIdentifier` logic with the `and` operator. 

In the future we may add an option to use the `or` operator, but it's not clear to me if there is any real benefit since it would increase the node pool instead of reducing it. Regardless, the capability is there if needed.

Closes #531

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -9,7 +9,6 @@ import (
 	"github.com/hashicorp/nomad-autoscaler/plugins"
 	nomadAPM "github.com/hashicorp/nomad-autoscaler/plugins/builtin/apm/nomad/plugin"
 	"github.com/hashicorp/nomad-autoscaler/sdk"
-	errHelper "github.com/hashicorp/nomad-autoscaler/sdk/helper/error"
 )
 
 // Processor helps process policies and perform common actions on them when
@@ -65,28 +64,7 @@ func (pr *Processor) ValidatePolicy(p *sdk.ScalingPolicy) error {
 		mErr = multierror.Append(mErr, errors.New("policy Min must not be greater Max"))
 	}
 
-	if err := pr.validateHorizontalClusterPolicy(p); err != nil {
-		mErr = multierror.Append(mErr, err)
-	}
-
 	return mErr.ErrorOrNil()
-}
-
-// validateHorizontalClusterPolicy validates that the target config for a
-// horizontal cluster scaling policy is valid. Calling this function is safe in
-// situations where it is not known whether the policy is a horizontal cluster
-// target or not.
-func (pr *Processor) validateHorizontalClusterPolicy(p *sdk.ScalingPolicy) error {
-
-	// Protect against the policy not targeting a horizontal cluster target so
-	// callers don't have to perform this.
-	if !p.Target.IsNodePoolTarget() {
-		return nil
-	}
-
-	// There are no extra validation needed at this point.
-	var mErr *multierror.Error
-	return errHelper.FormattedMultiError(mErr)
 }
 
 // CanonicalizeCheck sets standardised values on fields.
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -84,16 +84,8 @@ func (pr *Processor) validateHorizontalClusterPolicy(p *sdk.ScalingPolicy) error
 		return nil
 	}
 
+	// There are no extra validation needed at this point.
 	var mErr *multierror.Error
-
-	class := p.Target.Config[sdk.TargetConfigKeyClass]
-	dc := p.Target.Config[sdk.TargetConfigKeyDatacenter]
-
-	if class != "" && dc != "" {
-		mErr = multierror.Append(mErr, fmt.Errorf("target config must contain only one of %s",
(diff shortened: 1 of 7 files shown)


sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -0,0 +1,69 @@
...
+
+// Value satisfies the Value function on the ClusterNodePoolIdentifier
+// interface.
+func (c *combinedClusterPoolIdentifier) Value() string {
+	values := make([]string, 0, len(c.poolIdentifiers))
+	for _, identifier := range c.poolIdentifiers {
+		values = append(values, fmt.Sprintf("%s is %s", identifier.Key(), identifier.Value()))
+	}
+	return strings.Join(values, fmt.Sprintf(" %s ", c.mode))
+}
```
lgfa29, line 69: I wasn't sure how this should look like. It seems to only be used for log output, so I opted for a more human-friendly representation:

```
2021-11-12T10:13:17.228-0500 [INFO]  policy_eval.worker: scaling target: id=a21d3449-cf30-680b-add1-d1cc4f56616d policy_id=0501a099-feb7-99b7-3b2c-6daa7baa9567 queue=cluster target=aws-asg from=1 to=0 reason="scaling down because metric is 0" meta=map[nomad_policy_id:0501a099-feb7-99b7-3b2c-6daa7baa9567]
2021-11-12T10:13:17.347-0500 [DEBUG] internal_plugin.aws-asg: performing node pool filtering: combined_identifier="node_class is learn and datacenter is batch_workers"
2021-11-12T10:13:17.409-0500 [DEBUG] internal_plugin.aws-asg: performing node selection: selector_strategy=empty_ignore_system
```

I thought of just using `node_class=class and datacenter=dc1`, but that could make parsing harder since `=` is also used as the separator for the other labels. Maybe `node_class:class and datacenter:dc1`?

policy/policy.go
```diff
@@ -84,16 +84,8 @@ func (pr *Processor) validateHorizontalClusterPolicy(p *sdk.ScalingPolicy) error
...
-
-	class := p.Target.Config[sdk.TargetConfigKeyClass]
-	dc := p.Target.Config[sdk.TargetConfigKeyDatacenter]
-
-	if class != "" && dc != "" {
-		mErr = multierror.Append(mErr, fmt.Errorf("target config must contain only one of %s",
-			strings.Join(sdk.TargetConfigConflictingClusterParams, ", ")))
-	}
-
 	return errHelper.FormattedMultiError(mErr)
```
jrasell, line 89: Do we still need this multierror and call to FormattedMultiError as currently it will always be nil?

sdk/helper/scaleutils/nodepool/combined.go
jrasell: Yeah this is a special case and agree using the equals is probably going to get in the way. I personally find using the colon a little easier to grok but this is very subjective and I think it's fine like it is if you're happy!

jrasell (approved):
Great addition!

Minor comment inline but certainly not a blocker and I expect the answer is future proofing.

sdk/helper/scaleutils/nodepool/combined.go
lgfa29: I'm good either way, so colon it is!

PR #536, opened 2021-11-12
Dispense fixed-value, pass-through, and threshold strategy plugins by default

### LINKED ISSUE #532: Failed to dispense plugin threshold:
Hi!

I think this is a simple config error somewhere, but I couldn't find it. So the Autoscaler agent is working fine if I don't use `Threshold` strategy. As soon as I add this strategy the following error would appear ([This plugin](https://www.nomadproject.io/docs/autoscaling/plugins/strategy/threshold) is installed by default, right?)
```
2021-10-15T22:08:10.731Z [ERROR] agent: failed to start agent: error="failed to setup plugins: 1 error occurred:
	* failed to dispense plugin threshold: failed to instantiate plugin threshold client: fork/exec /plugins/threshold: no such file or directory
```

And here is my agent template:
```
job "autoscaler" {
  type = "service"

  datacenters = ["preproduction"]

  group "autoscaler" {
    count = 1

    network {
      port "http" {}
    }

    task "autoscaler" {
      driver = "exec"

      config {
        command = "/usr/local/bin/nomad-autoscaler"
        args    = [
          "agent",
          "-config",
          "${NOMAD_TASK_DIR}/config.hcl",
          "-policy-dir",
          "$${NOMAD_TASK_DIR}/policies/"
        ]
      }

      artifact {
        source      = "https://releases.hashicorp.com/nomad-autoscaler/0.3.0/nomad-autoscaler_0.3.0_linux_amd64.zip"
        destination = "/usr/local/bin"
      }

      template {
        data = <<EOF
          policy_eval {
            ack_timeout    = "10m"
            delivery_limit = 4

            workers = {
              cluster    = 2
              horizontal = 2
            }
          }

          nomad {
            address = "http://{{env "attr.unique.network.ip-address" }}:4646"
          }

          apm "datadog" {
            driver = "datadog"

            config = {
              dd_api_key = "KEY"
              dd_app_key = "KEY"
            }
          }
          
          strategy "threshold" {
            driver = "threshold"
          }
 
          target "aws-asg" {
            driver = "aws-asg"
            config = {
              aws_region = "eu-central"
            }
          }
          EOF

        destination = "${NOMAD_TASK_DIR}/config.hcl"
      }

      template {
        data = <<EOF
          scaling "cluster_policy" {
            min     = 2
            max     = 5
            enabled = true

            policy {
              cooldown            = "30s"
              evaluation_interval = "60s"

              check "free_memory" {
                source = "datadog"
                query  = "max:system.mem.free{host:nomad-*,env:preproduction}"

                strategy "threshold" {
                  upper_bound = 8000000000
                  lower_bound = 3500000000
                }
              }

              target "aws-asg" {
                ...
              }
            }
          }
        EOF
        destination = "${NOMAD_TASK_DIR}/policies/hashistack.hcl"
      }

      resources {
        cpu    = "1000"
        memory = "250"
      }

      service {
        name = "autoscaler"
        port = "http"

        check {
          type     = "http"
          path     = "/v1/health"
          interval = "3s"
          timeout  = "1s"
        }
      }
    }
  }
}
```

This is the use case: I want Autoscaler to monitor the free memory of each node and make sure there is enough memory for future allocations, also kill nodes if too much memory is available. So I chose threshold strategy for this with these two conditions: (data is provided by Datadog)
- The maximum free memory in any node is less than X -> more resource needed, the max free memory on any node is not enough for a new allocation.
- Or maximum free memory on any node is more than Y -> there are nodes with full free memory, so kill them.

If you think I can handle this by other strategies, please guide. Thanks so much for this product and your help :)

lgfa29:
When these new strategy plugins were added, we forgot to update the agent to auto-dispense them. 

I wonder if we should just auto-dispense _all_ builtin plugins and add a new flag to allow disabling specific ones if needed, but some of them require additional configuration (like `aws-asg`, `prometheus` etc), so the user would have to declare them anyway.

It would be cool if we could detect plugins that don't require additional config and auto-dispense them, but I couldn't think of a way to easily check this.

Closes #532

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## UNRELEASED
 
 IMPROVEMENTS:
+ * agent: Dispense `fixed-value`, `pass-through`, and `threshold` strategy plugins by default [[GH-536](https://github.com/hashicorp/nomad-autoscaler/pull/536)]
  * policy: Prevent scaling cluster to zero when using the Nomad APM [[GH-534](https://github.com/hashicorp/nomad-autoscaler/pull/534)]
  * scaleutils: Add combined filter to allow filtering by node class and datacenter [[GH-535](https://github.com/hashicorp/nomad-autoscaler/pull/535)]
 
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -547,9 +547,14 @@ func Default() (*Agent, error) {
 			AckTimeout:    defaultPolicyEvalAckTimeout,
 			Workers:       defaultPolicyEvalWorkers,
 		},
-		APMs:       []*Plugin{{Name: plugins.InternalAPMNomad, Driver: plugins.InternalAPMNomad}},
-		Strategies: []*Plugin{{Name: plugins.InternalStrategyTargetValue, Driver: plugins.InternalStrategyTargetValue}},
-		Targets:    []*Plugin{{Name: plugins.InternalTargetNomad, Driver: plugins.InternalTargetNomad}},
+		APMs: []*Plugin{{Name: plugins.InternalAPMNomad, Driver: plugins.InternalAPMNomad}},
+		Strategies: []*Plugin{
+			{Name: plugins.InternalStrategyFixedValue, Driver: plugins.InternalStrategyFixedValue},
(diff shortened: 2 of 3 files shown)


jrasell (approved):
LGTM!

PR #538, opened 2021-11-12
Fix error message when job scale request fails

lgfa29:
Read the right config key when logging the error message.


jrasell (approved):
LGTM!

PR #541, opened 2021-11-19
Prevent the Nomad target plugin from being blocked

### LINKED ISSUE #514: nomad-autoscaler sits idle without scaling jobs or evaluating policies
Hi folks,

A few times now the nomad-autoscaler (`0.3.3`) simply stops performing any actions, but the job stays up and running. There are no real logs or errors that seem to indicate the reason.

I've included my HCL and logs https://gist.github.com/evandam/c439d08a24b3d150a76335510d08a91f - note the last logs are from `2021-07-14T19:27:06.302` and the current time is `Jul 19 16:29:02 UTC 2021`. The job is still up and running, though.

You can see that up until the last log there was some activity and then it drops to zero.

Screen Shot 2021-07-19 at 9 21 11 AM

I'll drop logs to debug level to see if there's any insight, but this only seems to happen every few days/weeks sporadically.

lgfa29:
The Nomad target plugin caches job and group status locally to avoid querying the Nomad API every time a policy is evaluated. This is done by having background gorotuines that use blocking queries to update the local state whenever the job scale status changes.

The current implementation has a bug that can be triggered when the Nomad API returns an error at the first run. The blocking query look gets stuck in [this loop](https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.3/plugins/builtin/target/nomad/plugin/state.go#L135-L157) causing [`jsh.handleFirstRun()`](https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.3/plugins/builtin/target/nomad/plugin/state.go#L172) to never be called. This, in turn, blocks the entire plugin since the plugin `Status` call would be waiting indefinitely for the [channel signal](https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.3/plugins/builtin/target/nomad/plugin/plugin.go#L182) whilst [holding the `statusHandlersLock`](https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.3/plugins/builtin/target/nomad/plugin/plugin.go#L170).

This PR makes a few changes to avoid this situation.

#### Add a timeout to limit how long handlers can take to start

Once a handler starts, it never stops unless it receives a `404` from the API, meaning the job being monitored doesn't exist anymore. But in order for the plugin to work properly, it must return the target status at least once. If the API calls fails continually at first run, we should consider the handler as invalid and return an error. The target can then checked again in the next policy evaluation.

With this timeout, the plugin is guarantee to unblock and release the lock even if an error occurs in the first run.

#### Internalize access to internal handler state

The handlers internal state was being accessed directly by the plugin in some cases, such as calls to `Status` and the garbage collector. Since calls to the plugin are made concurrently by the policy workers, without any synchronization mechanism there were a high potential for data races to happen. 

This PR internalizes the state of handlers, such that they are only accessed via methods that must acquire an internal mutex lock beforehand.

#### Reducing thundering herd problems

Policies are evaluated at an interval set by their own properties. The starting time that the evaluations happen are set when they are first read. When multiple policies are read at the same time, and have the same `evaluation_interval`, they would all be evaluated at the same time, every time, causing a spike in load on the Autoscaler.

This PR adds a small random delay to distribute evaluations of policies registered at the same time.

Closes #514

### DIFF
--- a/plugins/builtin/target/nomad/plugin/state.go
+++ b/plugins/builtin/target/nomad/plugin/state.go
@@ -23,6 +23,13 @@ const (
 	metaKeyJobStoppedSuffix = ".stopped"
 )
 
+var (
+	// statusHandlerInitTimeout is the time limit a status handler must
+	// initialize before considering the operation a failure.
+	// Declared as a var instead of a const to allow overwriting it in tests.
+	statusHandlerInitTimeout = 30 * time.Second
+)
+
 // jobScaleStatusHandler is an individual handler on the /v1/job/<job>/scale
 // GET endpoint. It provides methods for obtaining the current scaling state of
 // a job and task group.
@@ -55,14 +62,27 @@ type jobScaleStatusHandler struct {
 	lastUpdated int64
 }
 
-func newJobScaleStatusHandler(client *api.Client, ns, jobID string, logger hclog.Logger) *jobScaleStatusHandler {
-	return &jobScaleStatusHandler{
+func newJobScaleStatusHandler(client *api.Client, ns, jobID string, logger hclog.Logger) (*jobScaleStatusHandler, error) {
+	jsh := &jobScaleStatusHandler{
 		client:      client,
 		initialDone: make(chan bool),
 		jobID:       jobID,
 		namespace:   ns,
 		logger:      logger.With(configKeyJobID, jobID),
 	}
+
+	go jsh.start()
+
+	// Wait for initial status data to be loaded.
+	// Set a timeout to make sure callers are not blocked indefinetely.
+	select {
+	case <-jsh.initialDone:
+	case <-time.After(statusHandlerInitTimeout):
+		jsh.setStopState()
+		return nil, fmt.Errorf("timeout while waiting for job scale status handler")
+	}
+
+	return jsh, nil
 }
 
 // status returns the cached scaling status of the passed group.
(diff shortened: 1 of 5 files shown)


plugins/builtin/target/nomad/plugin/state.go
```diff
@@ -51,18 +62,33 @@ type jobScaleStatusHandler struct {
...
 		initialDone: make(chan bool),
 		jobID:       jobID,
 		namespace:   ns,
 		logger:      logger.With(configKeyJobID, jobID),
 	}
+
+	go jsh.start()
+
+	// Wait for initial status data to be loaded.
+	// Set a timeout to make sure callers are not blocked indefinetely.
```
jrasell, line 77:
```suggestion
	// Set a timeout to make sure callers are not blocked indefinitely.
```

policy/handler.go
```diff
@@ -309,6 +310,12 @@ func (h *Handler) updateHandler(current, next *sdk.ScalingPolicy) {
 	// policy's evaluation interval has changed.
 	if current == nil || current.EvaluationInterval != next.EvaluationInterval {
 		h.ticker.Stop()
+
+		// Add a small random delay to spread the first evaluation of policies
+		// that are evaluated at the same time.
```
jrasell, line 315: It might be because it's early; I feel it would be nice for the comment to detail what the delay range could be to save the reader from having to work this out themselves.

jrasell (approved):
LGTM! No blocking comments.

PR #542, opened 2021-11-22
Downgrade error message from app scaling due to active deployment

### LINKED ISSUE #515: Drop error log for "job scaling blocked due to active deployment"
Just a slight quality-of-life request - can logs for this be dropped to WARN or maybe even INFO? 

To me it seems this is expected behavior, and wouldn't be considered an error. It would be nice to drop the severity on this one so alerts for real errors won't trip over this and it won't have to be filtered otherwise.

```
2021-07-19T16:37:34.048Z [ERROR] policy_eval.worker: failed to evaluate policy: eval_id=70c8ea00-1c40-9ffb-f732-29cddb212153 eval_token=8e13fa3c-8c68-2aa8-da50-be561332f67e id=3fe0dfca-376c-6a37-deba-dd84d95e9a2b policy_id=0b03eba3-153e-4c7f-dab9-325d3cff760e queue=horizontal error="failed to scale target: failed to scale group /: Unexpected response code: 400 (job scaling blocked due to active deployment)"
```

Low priority - just a thought. Thanks folks!

lgfa29:
The Autoscaler is fairly likely to encounter active deployments during policy evaluations, either due to previous scaling actions that are taking longer than the policy `evaluation_internval` and `cooldown`, or due to manual job updates by operators.

This situation is usually not detrimental to the overall scaling progress since the deployment is expected to eventually complete and future policy evaluations will correct the group `count` if necessary.

This PR reduces the log level for when this happen from an `ERROR` to just `INFO`. 

It also introduces a new concept of `TargetScalingNoOpError` which is a special error type that target plugins can use in situations like this, where the scaling action request didn't actually complete, but that's not really a problem.

Closes #515

### DIFF
--- a/sdk/target.go
+++ b/sdk/target.go
@@ -1,5 +1,29 @@
 package sdk
 
+import (
+	"fmt"
+)
+
+// TargetScalingNoOp is a special error type that can be used by target plugins
+// to indicate that a scaling request didn't result in any action, but didn't
+// fail either.
+// This can be used to avoid post-scaling actions such as placing the policy in
+// cooldown.
+type TargetScalingNoOpError struct {
+	Err error
+}
+
+// NewTargetScalingNoOpError returns a new target scaling no-op error with the
+// provided formatted message.
+func NewTargetScalingNoOpError(msg string, args ...interface{}) *TargetScalingNoOpError {
+	return &TargetScalingNoOpError{Err: fmt.Errorf(msg, args...)}
+}
+
+// Error implements the error interface.
+func (n *TargetScalingNoOpError) Error() string {
+	return n.Err.Error()
+}
+
 // TargetStatus is the response object when performing the Status call of the
 // target plugin interface. The response details key information about the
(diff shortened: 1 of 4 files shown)


sdk/target.go
```diff
@@ -1,5 +1,29 @@
...
 
+import (
+	"fmt"
+)
+
+// TargetScalingNoOp is a special error type that can be used by target plugins
+// to indicate that a scaling request didn't result in any action, but didn't
+// fail either.
+// This can be used to avoid post-scaling actions such as placing the policy in
+// cooldown.
```
jrasell, line 11:
```suggestion
// TargetScalingNoOpError is a special error type that can be used by target
// plugins to indicate that a scaling request didn't result in any action, but
// didn't fail either.
// This can be used to avoid post-scaling actions such as placing the policy in
// cooldown.
```

jrasell (approved):
LGTM!

PR #544, opened 2021-11-23
Allow enabling and disabling policy sources

### LINKED ISSUE #524: Allow to disable the nomad source

lgfa29:
This PR builds on top of the approach taken in #525 but allows a more flexible way to define which policy sources should be enabled.

It introduces a new configuration block called `source` that can be placed within the existing `policy` block. Currently this block takes only a label that specifies which source it controls and a boolean attribute to enable or disable it:

```hcl
policy {
  source "nomad" {
    enabled = false
  }
}
```

Both `file` and `nomad` sources are initially enabled by default, with `file` requiring that `policy -> dir` is also defined to be fully enabled, so everything remains backwards compatible.

The CLI flag parsing is a bit different than others, and it special cases the existing policy sources. This seems fine to me, as all policy sources are internal and having CLI flags to disable them sounds pretty useful.

If no policy source is enabled, the Autoscaler agent will fail and exit, since it won't be able to do any work without policies. In the future, we may want to implement a full reload of policy sources via `SIGHUP` so that sources can be enabled/disabled dynamically.

### Possible future enhancements

This approach allows further customization of policy sources, such as moving the current `policy -> dir` into a new `config` map inside the `file` source:

```hcl
source "file" {
  # Possible future enhancement.
  config = {
    dir = "./policies"
  }
}
```

With this, we could start supporting multiple configurations for the same policy source, such as reading policies from multiple paths:

```hcl
source "file" "aws" {
  # Possible future enhancement.
  config = {
    dir = "./policies/aws"
  }
}

source "file" "gcp" {
  # Possible future enhancement.
  config = {
    dir = "./policies/gcp"
  }
}
```

Another possibility would be to start supporting external policy source plugins:

```hcl
source "my-plugin" {
  # Possible future enhancement.
  driver = "my_source_plugin"
  config = {
    key = "value"
  }
}
```

This pattern of disabling internal components via a `enabled` attribute could also be used for plugins:

```hcl
strategy "target-value" {
  driver  = "target-value"
  # Possible future enhancement.
  enabled = false
}
```

Closes #524

### DIFF
--- a/agent/config/config.go
+++ b/agent/config/config.go
@@ -65,10 +65,9 @@ type Agent struct {
 	// Consul is used to configure a consul API client
 	Consul *Consul `hcl:"consul,block"`
 
-	PolicySources []*PolicySource `hcl:"source,block"`
-	APMs          []*Plugin       `hcl:"apm,block"`
-	Targets       []*Plugin       `hcl:"target,block"`
-	Strategies    []*Plugin       `hcl:"strategy,block"`
+	APMs       []*Plugin `hcl:"apm,block"`
+	Targets    []*Plugin `hcl:"target,block"`
+	Strategies []*Plugin `hcl:"strategy,block"`
 }
 
 // DynamicApplicationSizing contains configuration values to control the
@@ -458,6 +457,9 @@ type Policy struct {
 	// `evaluation_interval` is not defined in a policy.
 	DefaultEvaluationInterval    time.Duration
 	DefaultEvaluationIntervalHCL string `hcl:"default_evaluation_interval,optional" json:"-"`
+
+	// Sources store configuration for policy sources.
+	Sources []*PolicySource `hcl:"source,block"`
 }
 
 // PolicyEval holds the configuration related to the policy evaluation process.
@@ -560,15 +562,15 @@ func Default() (*Agent, error) {
 		Policy: &Policy{
 			DefaultCooldown:           defaultPolicyCooldown,
 			DefaultEvaluationInterval: defaultEvaluationInterval,
+			Sources: []*PolicySource{
+				{Name: policySourceFile, Enabled: true},
+				{Name: policySourceNomad, Enabled: true},
+			},
 		},
 		PolicyEval: &PolicyEval{
 			DeliveryLimit: defaultPolicyEvalDeliveryLimit,
 			AckTimeout:    defaultPolicyEvalAckTimeout,
 			Workers:       defaultPolicyEvalWorkers,
-		},
-		PolicySources: []*PolicySource{
-			{Name: policySourceFile, Enabled: true},
-			{Name: policySourceNomad, Enabled: true},
 		},
 		APMs: []*Plugin{
 			{Name: plugins.InternalAPMNomad, Driver: plugins.InternalAPMNomad},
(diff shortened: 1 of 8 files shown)


agent/config/config.go
```diff
@@ -65,9 +65,10 @@ type Agent struct {
 	// Consul is used to configure a consul API client
 	Consul *Consul `hcl:"consul,block"`
 
-	APMs       []*Plugin `hcl:"apm,block"`
-	Targets    []*Plugin `hcl:"target,block"`
-	Strategies []*Plugin `hcl:"strategy,block"`
+	PolicySources []*PolicySource `hcl:"source,block"`
```
jrasell, line 68: `source` feels ambiguous when considering the agent already includes a `policy` entry. I think having `source` under the `policy` block would make more sense.

```hcl
policy {
  dir = "/opt/policies"
  
  source "nomad" {
    enabled = false
  } 
}
```

agent/config/config.go
```diff
@@ -475,6 +476,13 @@ type PolicyEval struct {
 	Workers map[string]int `hcl:"workers,optional"`
 }
 
+// PolicySource is an individual configured policy source.
+type PolicySource struct {
+	Name       string `hcl:"name,label"`
+	EnabledPtr *bool  `hcl:"enabled,optional"`
+	Enabled    bool
```
jrasell, line 483: I am curious about the reasoning for using a pointer along with a non-pointer for the value? Is this so we don't have to dereference every time when reading the value out? If so, could we add a comment to clarify this.

agent/config/config.go
```diff
@@ -510,6 +518,17 @@ const (
 	defaultPolicyEvalAckTimeout = 5 * time.Minute
 )
 
+// TODO: there's an unexpected import cycle that prevents us from using the
+// values defined in policy/source.go.
```
jrasell, line 522: A candidate to move to the `sdk` package?

agent/config/config.go
lgfa29: Good point, I will make this change.

agent/config/config.go
lgfa29: I used a pointer so that an empty block (like `source "nomad" {}`) would have `enabled` set to `true`. 

I think it wouldn't be possible to distinguish between an explicit `enabled = false` and a missing `enabled` otherwise? 🤔

agent/config/config.go
lgfa29: I'm not sure to be honest (hence the `TODO` to punt it to future Luiz 😬).

I think the root of the problem is this import: https://github.com/hashicorp/nomad-autoscaler/blob/v0.3.3/sdk/helper/nomad/config.go#L7

The `sdk` package should never require anything outside itself (and I thought we had a check for that), so whole agent config is probably what we would want to move to `sdk`, but that would require quite a bit of refactoring (mostly copy and paste though).

agent/config/config.go
jrasell: Yeah I understand the use of the pointer, just curious of the reason for using it alongside a non-pointer rather than referencing the pointer field.

jrasell (approved):

agent/config/config.go
lgfa29: Ahh got it. Yeah, that makes sense. I think I was just copy and pasting code 😅

PR #547, opened 2021-11-24
Add support for TLS CA certificate on Prometheus APM

### LINKED ISSUE #497: Support providing CA Certificate for Prometheus APM
We run Prometheus with only a TLS listener using a certificate from an internal CA. It would be nice if we could provide that CA certificate for the Prometheus APM like we can for the Nomad configuration.

lgfa29:
Manually testes using the instructions in https://prometheus.io/docs/guides/tls-encryption/ but using [these steps](https://gist.github.com/croxton/ebfb5f3ac143cd86542788f972434c96) to generate the certificate, otherwise I would get this error from the Prometheus client:

```
2021-11-24T10:07:53.393-0500 [WARN]  policy_eval.worker: failed to run check: id=377a4f0d-834f-e87c-9cbe-8a05d4e91b57 policy_id=7fe5c606-3d3b-ecd0-ed70-06566acd04bb queue=horizontal target=nomad-target error="failed to query source: failed to query: Post "https://example.com:55001/api/v1/query_range": x509: certificate relies on legacy Common Name field, use SANs instead"
```

Closes #497

### DIFF
--- a/plugins/builtin/apm/prometheus/plugin/plugin.go
+++ b/plugins/builtin/apm/prometheus/plugin/plugin.go
@@ -2,9 +2,14 @@ package plugin
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/pem"
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"math"
+	"strconv"
 	"time"
 
 	hclog "github.com/hashicorp/go-hclog"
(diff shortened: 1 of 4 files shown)


plugins/builtin/apm/prometheus/plugin/plugin.go
```diff
@@ -145,6 +163,41 @@ func (a *APMPlugin) QueryMultiple(q string, r sdk.TimeRange) ([]sdk.TimestampedM
 	}
 }
 
+func generateTLSConfig(config map[string]string) (*tls.Config, error) {
+	tlsConfig := tls.Config{}
+
+	// Load the CA certificate if present.
+	caCertPath := config[configKeyCACert]
+	fmt.Println(caCertPath)
```
jrasell, line 171: Assuming this is a forgotten debugging line?

jrasell (approved):
LGTM with a minor comment.

plugins/builtin/apm/prometheus/plugin/plugin.go
lgfa29: 🤦

PR #552, opened 2021-12-08
[APM Datadog] return 429 specific error

adrianlop:
err is != nil for 429s too so the code will never return the specific 429 error msg


jrasell (approved):
LGTM, thanks @adrianlop!

lgfa29 (approved):
Thanks @adrianlop!

I pushed a commit to add a changelog entry and had to update from `main` to fix a conflict.

PR #566, opened 2022-02-16
Add `on_error` and `on_check_error` config

### LINKED ISSUE #562: Autoscaler should have an option to avoid scaling when a check fails
Hi,

We are using the nomad-autoscaler to scaler an AWS ASG and we saw that when a check fails the nomad-autoscaler scales down the node pool.

```
2022-01-20T19:34:36.413Z [WARN]  policy_eval.worker: failed to run check: id=1ba3ad63-96d3-98dd-3d40-b461e1c8b5dd policy_id=44818048-3983-d50a-5c84-47c02e084c86 queue=cluster target=aws-asg error="failed to query source: failed to query: Post "http://localhost:9090/api/v1/query_range": EOF"
2022-01-20T19:34:36.419Z [INFO]  policy_eval.worker: scaling target: id=1ba3ad63-96d3-98dd-3d40-b461e1c8b5dd policy_id=44818048-3983-d50a-5c84-47c02e084c86 queue=cluster target=aws-asg from=26 to=23 reason="capped count from 21 to 23 to stay within limits" meta="map[nomad_autoscaler.count.capped:true nomad_autoscaler.count.original:21 nomad_autoscaler.reason_history:[scaling down because factor is 0.798563] nomad_policy_id:44818048-3983-d50a-5c84-47c02e084c86]"
```

The reason for that is that we have several checks (one for memory and one for cpu) and normally, the nodes are limited by CPU and they usually have more free memory. So the scaling is mostly limited by CPU. 

```
    check "cpu_allocated_percentage" {
      source = "prometheus"
      query  = "scalar(nomad:allocated_cpu:ratio{node_class=\"default\",node_status=\"ready\"})"

      strategy "target-value" {
        target = 87
      }
    }

    check "mem_allocated_percentage" {
      source = "prometheus"
      query  = "scalar(nomad:allocated_memory:ratio{node_class=\"default\",node_status=\"ready\"})"

      strategy "target-value" {
        target = 87
      }
    }
```

The problem is that if the CPU check fails, the nomad-autoscaler currently continues and uses only the value of the memory check. I think that at least a configuration option should be provided that allows to set that if one check fails the scaling should be aborted.

We saw this in 0.3.4, but seeing the code in main seems that the same would happen. Maybe something like the following would be nice. In the meantime, as a workaround, we're leaving only the CPU check in the policy.

```
	// Start check handlers.
	for _, checkEval := range eval.CheckEvaluations {
		checkHandler := newCheckHandler(logger, eval.Policy, checkEval, w.pluginManager)

		// Wrap target status call in a goroutine so we can listen for ctx as well.
		var action *sdk.ScalingAction
		var err error
		doneCh := make(chan interface{})

		go func() {
			defer close(doneCh)
			action, err = checkHandler.start(handlersCtx, currentStatus)
		}()

		select {
		case <-ctx.Done():
			w.logger.Info("stopping worker")
			return nil
		case <-doneCh:
		}

		if err != nil {
			logger.Warn("failed to run check", "error", err)
			
			// NEW CODE
                        if eval.Policy.AbortOnCheckFail {
                          winningAction = nil
                          break
                        }
                        // NEW CODE
			
			continue
		}

		winningAction = sdk.PreemptScalingAction(winningAction, action)
		if winningAction == action {
			winningHandler = checkHandler
		}
	}

	// At this point the checks have finished. Therefore emit of metric data
	// tracking how long it takes to run all the checks within a policy.
	metrics.MeasureSinceWithLabels([]string{"scale", "evaluate_ms"}, evalStartTime, labels)

	if winningHandler == nil || winningAction == nil || winningAction.Direction == sdk.ScaleDirectionNone {
		logger.Debug("no checks need to be executed")
		return nil
	}

```

lgfa29:
Introduce two new configuration options (`on_error` and `on_check_error`) to allow operators fine-grain control on how to handle errors when checks are being evaluated.

`on_check_error` is set at policy level and is applied to all `check` by default. Each `check` may override this setting using `on_error`.

For example, this policy will fail its evaluation if _any_ `check` fails:

```hcl
scaling "batch" {
  enabled = true
  min     = 0
  max     = 5

  policy {
    cooldown            = "1m"
    evaluation_interval = "10s"
    on_check_error      = "fail"

    check "batch_jobs_in_progess" {
      source = "datadog"
      query  = "avg:proxy.backend.response.time{proxy-service:web-app}"

      strategy "noop-strategy" {}
    }

    check "noop" {
      source = "noop-apm"
      query  = "fixed:5"

      strategy "noop-strategy" {}
    }

    target "noop-target" {}
  }
}
```

But we can also set a fine-grained behaviour per `check`:

```hcl
scaling "batch" {
  enabled = true
  min     = 0
  max     = 5

  policy {
    cooldown            = "1m"
    evaluation_interval = "10s"
    on_check_error      = "fail"

    check "batch_jobs_in_progess" {
      source = "datadog"
      query  = "avg:proxy.backend.response.time{proxy-service:web-app}"

      strategy "noop-strategy" {}
    }

    check "noop" {
      source   = "noop-apm"
      query    = "fixed:5"
      on_error = "ignore"

      strategy "noop-strategy" {}
    }

    target "noop-target" {}
  }
}
```

Closes #562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 IMPROVEMENTS:
  * plugins/target/aws-asg: Support EC2 role credentials [[GH-564](https://github.com/hashicorp/nomad-autoscaler/pull/564)]
+ * policy: Add `on_check_error` and `on_error` configuration [[GH-566](https://github.com/hashicorp/nomad-autoscaler/pull/566)]
 
 ## 0.3.5 (January 20, 2022)
 
--- a/policy/file/parse_test.go
+++ b/policy/file/parse_test.go
@@ -26,6 +26,7 @@ func Test_decodeFile(t *testing.T) {
 					Max:                100,
 					Cooldown:           10 * time.Minute,
 					EvaluationInterval: 1 * time.Minute,
+					OnCheckError:       "error",
 					Checks: []*sdk.ScalingPolicyCheck{
 						{
 							Name:        "cpu_nomad",
@@ -40,9 +41,10 @@ func Test_decodeFile(t *testing.T) {
 							},
 						},
 						{
-							Name:   "memory_prom",
-							Source: "prometheus",
-							Query:  "nomad_client_allocated_memory*100/(nomad_client_allocated_memory+nomad_client_unallocated_memory)",
+							Name:    "memory_prom",
+							OnError: "ignore",
+							Source:  "prometheus",
+							Query:   "nomad_client_allocated_memory*100/(nomad_client_allocated_memory+nomad_client_unallocated_memory)",
 							Strategy: &sdk.ScalingPolicyStrategy{
 								Name: "target-value",
 								Config: map[string]string{
--- a/policy/file/test-fixtures/full-cluster-policy.hcl
+++ b/policy/file/test-fixtures/full-cluster-policy.hcl
@@ -8,6 +8,7 @@ scaling "full-cluster-policy" {
 
     cooldown            = "10m"
     evaluation_interval = "1m"
+    on_check_error      = "error"
 
     check "cpu_nomad" {
       source       = "nomad_apm"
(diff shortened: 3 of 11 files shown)


jrasell (approved):
LGTM!

We should update the Nomad docs also to reference this.

PR #588, opened 2022-05-26
make sure target is always between min and max

lgfa29:
The current policy evaluation always attempts to query the APM for
metrics and apply the policy strategy before making a scaling decision.

But in some situations the target may be out of the specified min and
max range from the beginning. In these situations, we would want to
correct the target first so it's in within the expected range before
applying the rest of the policy.

This change short-circuits the most the policy evaluation if the target
count status is outside this range.


DerekStrickland (approved):
:shipit:

jrasell (approved):
LGTM!

PR #594, opened 2022-08-09
feat(aws-asg): add `retry_attempts` configuration to customize the maximum waiting time

### LINKED ISSUE #518: aws-asg: configureable retry interval/retry
Hey, 

first of all, thank you for the autoscaler, seems to be designed really well! 

We are using AWS autoscaler and we have some ELB connection draining, which might take more than 2 minutes in some cases. The autoscaler gives up after couple of retries, would it be possible to make retry count and interval configurable?

https://github.com/hashicorp/nomad-autoscaler/blob/main/plugins/builtin/target/aws-asg/plugin/aws.go#L16-L20

Thanks!

alopezsanchez:
The maximum waiting time for checking the status of the ASG is 2 minutes, which is not configurable.
When using the ASG target plugin and downscaling, the ELB connection draining could take more
than 2 minutes. If this happens, Nomad will purge the nodes without waiting for the ASG to remove
the instances.

Logs:
```
2022-07-28T15:48:51.939Z [ERROR] internal_plugin.aws-asg: failed to ensure all activities completed: action=scale_in asg_name=preproduction/nomad error="reached retry limit"
2022-07-28T15:48:52.123Z [INFO]  internal_plugin.aws-asg: successfully purged Nomad node: node_id=34862212-f89a-b107-903d-f4036a44543d nomad_evals=["88a3c562-31e6-f303-cae3-1289bd924958", "b0b08d76-0c2c-e5f3-1b8d-d6f032cd89aa"]
```

So, this commit adds a new config property called `retry_attempts`, which is the maximum number
of attempts the plugin will make to ensure that the ASG has completed all its actions.

Apart from that, if this PR is accepted, I will update the [documentation](https://www.nomadproject.io/tools/autoscaling/plugins/target/aws-asg#policy-configuration-options) accordingly.

Resolves #518

### DIFF
--- a/plugins/builtin/target/aws-asg/plugin/aws.go
+++ b/plugins/builtin/target/aws-asg/plugin/aws.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -18,9 +19,17 @@ import (
 
 const (
 	defaultRetryInterval  = 10 * time.Second
-	defaultRetryLimit     = 15
 	nodeAttrAWSInstanceID = "unique.platform.aws.instance-id"
 )
+
+func getConfigValue(config map[string]string, key string, defaultValue string) string {
+	value, ok := config[key]
+	if !ok {
+		return defaultValue
+	}
+
+	return value
+}
 
 // setupAWSClients takes the passed config mapping and instantiates the
 // required AWS service clients.
@@ -300,7 +309,12 @@ func (t *TargetPlugin) ensureActivitiesComplete(ctx context.Context, asg string,
 		return false, fmt.Errorf("waiting for %v activities to finish", len(ids))
 	}
 
-	return retry(ctx, defaultRetryInterval, defaultRetryLimit, f)
+	retryLimit, err := strconv.Atoi(getConfigValue(t.config, configKeyRetryAttempts, configValueRetryAttemptsDefault))
+	if err != nil {
+		return err
+	}
+
+	return retry(ctx, defaultRetryInterval, retryLimit, f)
 }
 
 func (t *TargetPlugin) ensureASGInstancesCount(ctx context.Context, desired int64, asgName string) error {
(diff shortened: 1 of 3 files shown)


plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -18,10 +19,18 @@ import (
 
 const (
 	defaultRetryInterval  = 10 * time.Second
-	defaultRetryLimit     = 15
```
alopezsanchez, line 21: I removed this from here since there will be a configuration for it.

plugins/builtin/target/aws-asg/plugin/plugin.go
```diff
@@ -28,10 +28,12 @@ const (
 	configKeySessionToken       = "aws_session_token"
 	configKeyASGName            = "aws_asg_name"
 	configKeyCredentialProvider = "aws_credential_provider"
+	configKeyRetryAttempts      = "retry_attempts"
 
 	// configValues are the default values used when a configuration key is not
 	// supplied by the operator that are specific to the plugin.
-	configValueRegionDefault = "us-east-1"
+	configValueRegionDefault        = "us-east-1"
+	configValueRetryAttemptsDefault = "15"
```
alopezsanchez, line 36: `config` values must be strings. Hence the default value should also be a string.

plugins/builtin/target/aws-asg/plugin/plugin.go
```diff
@@ -220,3 +232,12 @@ func processLastActivity(activity types.Activity, status *sdk.TargetStatus) {
...
 }
+
+func getConfigValue(config map[string]string, key string, defaultValue string) string {
+	value, ok := config[key]
+	if !ok {
+		return defaultValue
+	}
+
+	return value
+}
```
lgfa29, line 243: This is a nice little helper, good one 🙂

lgfa29 (approved):
Thanks for the PR @alopezsanchez, and apologies for the long delay in getting this reviewed.

I pushed a commit with some small refactoring to parse the config value earlier in the `SetConfig` method. This allows for quicker detection of invalid values (like non-integer strings), otherwise the problem would only be caught at scaling time.

PR #595, opened 2022-08-11
Instance refresh consideration

tirimia:


tirimia:
Made a small mistake, will return with a proper PR

PR #608, opened 2023-01-05
build: update to Go 1.19

lgfa29:
Update toolchain to use Go 1.19.3.

The failing CircleCI will be fixed once https://github.com/hashicorp/nomad-autoscaler/pull/609 is merged.

PR #609, opened 2023-01-05
ci: move tests to GHA

lgfa29:
This work builds on top of https://github.com/hashicorp/nomad-autoscaler/pull/608 to make sure CI works after the toolchain update.


jrasell (approved):
LGTM!

tgross (approved):

PR #621, opened 2023-03-01
log something useful when dry-running cluster scaling policies

erulabs:
When using `dry-run` for cluster autoscaling, it's easy to get a bit confused because we see no output whatsoever!

The change means that AWS/GKE/Azure cluster autoscaling will print an INFO log letting us know what *would* have happened.

### DIFF
--- a/plugins/builtin/apm/datadog/plugin/plugin.go
+++ b/plugins/builtin/apm/datadog/plugin/plugin.go
@@ -156,7 +156,7 @@ func (a *APMPlugin) QueryMultiple(q string, r sdk.TimeRange) ([]sdk.TimestampedM
 
 	series := queryResult.GetSeries()
 	if len(series) == 0 {
-		a.logger.Warn("empty time series response from datadog, try a wider query window")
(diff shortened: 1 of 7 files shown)


erulabs:
Ah, it seems that I should have tried TRACE level logging first:

https://github.com/hashicorp/nomad-autoscaler/blob/04c6cf689fb9aac43b36ae2c4bb055652f4a323b/plugins/builtin/strategy/target-value/plugin/plugin.go#L136-L142

I'd suggest that this is probably the most important thing nomad-autoscaler could log! Particularly in dry-run mode, it's very useful to know what would happen and why.

I'll change these to INFO level logs and that ought to be enough for me!

lgfa29:
Hi @erulabs 👋 

Thanks for the PR, but yeah, we have log a lot of information at `TRACE` level because the Autoscaler can be quite chatty depending on how many policies you have and how often they are evaluated.

But I agree that they are easy to miss, specially since the default level is `INFO`. I created https://github.com/hashicorp/nomad-autoscaler/issues/624 to describe the problem and also propose some options.

Feel free to add any more comments there or take on it if you are interested 🙂

Issue #622, opened 2023-03-08
Fail-safe scale-up when metrics plugin returns no data


erulabs:
You may be aware that [datadog is experiencing a large outage today](https://status.datadoghq.com/). This means that nomad-autoscaler, when using datadog as a `source`, is unable to collect any metrics.

There is currently a `on_error` option to either `ignore` or `fail` (ignore moving on to other checks and fail stopping all actions). It seem to me a 3rd option here seems reasonable, perhaps called `scale`.

The idea would be that if a check fails to get any metrics, it could be set to `on_error = "scale"` which would consider the check active. In this example, if datadog goes offline or no metrics are reported, the nomad-autoscaler would trigger a scale-up and add additional instances (according to our `delta` in this example).

```
policy {
  check "api_needs_uppies" {
    source = "datadog"
    on_error = "scale"
    query = "ewma_3(avg:api.concurrent_request_per_container.poll)"
    query_window = "5m"
    group = "concurrent_reqs"
    strategy "threshold" {
      lower_bound = 1.6
      delta = 10
    }
  }
  check "api_needs_downies" {
    source = "datadog"
    on_error = "ignore"
    query = "ewma_3(avg:api.concurrent_request_per_container.poll)"
    query_window = "10m"
    group = "concurrent_reqs"
    strategy "threshold" {
      upper_bound = 1.1
      delta = -4
      within_bounds_trigger = 10
    }
  }
}
```

Any interest in this feature? I will probably write this for our purposes over at @ClassDojo, but I suspect it would be a good mainline feature for nomad-autoscaler in general!

lgfa29:
Thanks for the suggestion and PR @erulabs!

Having an APM outage does sound like a huge pain....But I'm not sure what the safe option would be in this case 🤔 

Since this is (hopefully) an expected situation, I worry that it may not be immediately obvious what the consequence of using `on_error = "scale"` may be, causing an avalanche of scaling ups when your APM does go down.

It would also be good to have some way of distinguishing between a long-term failure vs. a blip. We may not want to fully scale on first error.

This all leads me to believer that we would need a more advanced way of configuring this, but I don't have any good ideas right now 😞 

The first thing that pops in my head would be a special `check` block that is used when all previous checks fail. So something like:

```hcl
policy {
  check "api_needs_uppies" {
    # ...
  }

  check "api_needs_downies" {
    # ...
  }

  error_handler {
    failures_before_triggering = 10
    strategy "fixed-value" {
[...]

PR #623, opened 2023-03-09
add ScalingPolicyOnErrorScale

erulabs:
Addresses feature request #622 

Adds a new `on_error` value for scaling checks which allows `on_error = "scale"`. This value means that when metrics fail to be collected (syntax error, no metrics, [datadog downtime](https://status.datadoghq.com/)), we can specify a check that should fail-safe - ie: scale towards max.

Some context: Datadog outage occurred right around our scale-up time - a lack of metrics meant nomad-autoscaler did nothing. We'd like it to fail open and automatically scale towards max when something has gone wrong.


lgfa29:
Just noting here that there are some discussion in https://github.com/hashicorp/nomad-autoscaler/issues/622. If anyone has more ideas or feedback feel free to include them there 🙂

Issue #519, opened 2021-08-20
Failed to get policy (EOF)


evandam:
Hi folks,

Seeing some more errors and Nomad jobs that should definitely be triggering scaling are not happening. 

https://gist.github.com/evandam/4956199b01cd4303873b69cf7d3779c1

My first thought is that the Nomad agent is having issues, but it seems okay and I'm able to hit the same endpoints in the logs directly on the host 🤷‍♂️ 

As a side note, is there a decent way to debug things? There's `policy_id` in logs but I'm not sure how that maps back to the actual policy or Nomad job in question. It seems like it would be very helpful to be able to trace these back.

We're running Nomad 1.1.2 for what it's worth.

Any thoughts?

lgfa29:
Hum...it's the [second time](https://github.com/hashicorp/nomad-autoscaler/issues/514#issuecomment-893028789) this is being reported, so there may be something wrong with our blocking query handling.

Do you happen to have anything sitting between Nomad and the Autoscaler? Like a load balancer or reverse proxy?

> There's policy_id in logs but I'm not sure how that maps back to the actual policy or Nomad job in question

Yeah, it's not ideal but you can query Nomad for the scaling policies, either using the [`nomad scaling policy list`](https://www.nomadproject.io/docs/commands/scaling/policy-list) command or the [GET `/v1/scaling/policies`](https://www.nomadproject.io/api-docs/scaling-policies#list-scaling-policies) endpoint, to get the policy ID.

evandam:
I'm running the autoscaler as a Nomad job itself, talking to the local Nomad agent. Here's the simplified HCL: https://gist.github.com/evandam/40430410b9bb18cca74aad3714e5f568

hntrmrrs:
We have experienced a similar issue where a failure to list scaling policies ends up with the autoscaler effectively hanging, periodically emitting logs like these (every ~3 minutes with our configuration):

```
2022-03-02T13:30:11.354Z [ERROR] policy_manager.policy_handler: timeout: failed to read policy in time: policy_id=c8449594-6c20-ca82-4bb5-869772a8aea1
2022-03-02T13:33:11.355Z [ERROR] policy_manager.policy_handler: timeout: failed to read policy in time: policy_id=c8449594-6c20-ca82-4bb5-869772a8aea1
[...]

PR #625, opened 2023-04-21
always return file policy for same id

Seros:
Addresses #519 

I tried to reproduce the issue locally by setting up the autoscale and a local nomad agent running and connection each other. When restarting nomad the error message "failed to call the Nomad list policies" was shown as described in the issue and during debugging I noticed that the handler has no current policy anymore. I digged deeper and found the reason. The file source does not return a policy because the one that gets monitored is already registered in the policyMap and so no policy update is triggered which would set the currentPolicy at https://github.com/hashicorp/nomad-autoscaler/blob/main/policy/handler.go#L153
With this change the policy will be returned no matter what happened to the handler/source.

### DIFF
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -7,7 +7,6 @@ import (
 	"context"
 	"crypto/md5"
 	"fmt"
-	"reflect"
 	"sync"
 
 	hclog "github.com/hashicorp/go-hclog"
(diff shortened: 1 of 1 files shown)


gulducat:
As mentioned in #733, this totally does fix the issue, but selfishly I went with my own PR instead, so I'll close this one out.

Thanks again for the submission @Seros! :heart:

PR #635, opened 2023-05-01
ci: fix nightly release

lgfa29:
The GHA version pinning bot upgraded all action versions, which pulled several breaking changes from `actions/github-script`.

The one that affected us here is that the `git` context was moved to `rest` (https://github.com/actions/github-script/issues/328).

Failed run: https://github.com/hashicorp/nomad-autoscaler/actions/runs/4848130638

Successful run: https://github.com/hashicorp/nomad-autoscaler/actions/runs/4853454638


jrasell (approved):
LGTM, thanks!

Issue #634, opened 2023-04-28
deps: update Nomad API dependency to v1.5.3


jrasell:
The Nomad API dependency should be updated to the latest released version, which at the time of writing this issue is v1.5.3. The Nomad API package is not versioned, therefore the command `git rev-list -n 1 v1.5.3` can be used to find the appropriate commit to use in the `go get` command.

PR #636, opened 2023-05-03
Update the dependency on the Nomad api to the latest release (1.5.5)

Juanadelacuesta:
Mentioned in issue [634](https://github.com/hashicorp/nomad-autoscaler/issues/634)

This PR bumps up the version of the nomad API used in the autoscaler


plugins/builtin/apm/nomad/nomad
lgfa29, line 1: Ops, I thought these were being `.gitignore`d, but apparently not 😅 

Could you remove this and the other test plugin binaries?

Normally you don't need to build the built-in plugins, but if you do happen to need them you can use [these make targets](https://github.com/hashicorp/nomad-autoscaler/blob/main/GNUmakefile#L152-L216) that will place them in an ignored folder.

I don't know if there's an easy way to ignore them other than listing them all in `.gitignore` 🤔 

Maybe add `.gitignore` files in each of these directories?

go.mod
```diff
@@ -3,54 +3,61 @@ module github.com/hashicorp/nomad-autoscaler
 go 1.19
```
lgfa29, line 3:
```suggestion
go 1.20
```

We missed this one in https://github.com/hashicorp/nomad-autoscaler/pull/631, and all other plugins `go.mod` as well (we really need a script to handle these updates 😬 ).

go.mod
```diff
@@ -3,54 +3,61 @@ module github.com/hashicorp/nomad-autoscaler
...
-	github.com/stretchr/testify v1.7.1
-	google.golang.org/api v0.80.0
-	google.golang.org/grpc v1.46.0
-	google.golang.org/protobuf v1.28.0
+	github.com/prometheus/client_golang v1.15.1
+	github.com/prometheus/common v0.42.0
+	github.com/stretchr/testify v1.8.2
+	google.golang.org/api v0.120.0
+	google.golang.org/grpc v1.54.0
+	google.golang.org/protobuf v1.30.0
```
lgfa29, line 32: Did all of these updates come from Nomad? While they look OK we've had some backwards incompatibility issues on dependency upgrade in the past, so it may be better to upgrade them in separate commits so we can bisect problems.

It would be also good to run a batch of [cluster scaling](https://developer.hashicorp.com/nomad/tutorials/autoscaler/horizontal-cluster-scaling) tests after the version bump.

lgfa29 (changes requested):

plugins/builtin/apm/nomad/nomad
Juanadelacuesta: lets find out! :)

go.mod
Juanadelacuesta: Done!

go.mod
Juanadelacuesta: Successfully run, without anything to report.

jrasell (approved):
LGTM!

Just missing a changelog entry, then we can merge.

PR #641, opened 2023-05-16
build: remove deprecated set-output command from GHA

tgross:
Fixes: #639

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ jobs:
         id: get-go-version
         run: |
           echo "Building with Go $(cat .go-version)"
-          echo "::set-output name=go-version::$(cat .go-version)"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -66,7 +66,7 @@ jobs:
         id: generate-ldflags
         run: |
           project="$(go list -m)"
-          echo "::set-output name=ldflags::"-X \'$project/version.GitDescribe=v${{ needs.get-product-version.outputs.product-version }}\'""
+          echo "ldflags="-X \'$project/version.GitDescribe=v${{ needs.get-product-version.outputs.product-version }}\'"" >> $GITHUB_OUTPUT
```
tgross, line 69: I have no idea if this will be broken with the extra `=` in there and have no way of testing it, so... :man_shrugging: I guess we'll find out whether this works the next time we do a release.

lgfa29 (approved):

PR #643, opened 2023-05-22
Migrate CRT configuration to 'prepare' workflow

smacfarlane:
This replaces the explicit configuration enumerating each step in the common release tooling process with a single step that encapsulates all organizational requirements (signing, notarization, etc.).


jrasell (commented):
Thanks @smacfarlane. I just wanted to check that it's expected that `build` no longer performs any action?

jrasell (approved):

PR #647, opened 2023-06-15
Noissue mcj cleanup

Juanadelacuesta:


policyeval/base_worker.go
```diff
@@ -120,13 +115,7 @@ func (w *BaseWorker) handlePolicy(ctx context.Context, eval *sdk.ScalingEvaluati
...
 	logger.Debug("fetching current count")
 
-	currentStatus, err := w.runTargetStatus(targetInst, eval.Policy)
-	if err != nil {
-		return fmt.Errorf("failed to fetch current count: %v", err)
-	}
-	if !currentStatus.Ready {
-		return errTargetNotReady
-	}
+	currentStatus := eval.TargetStatus
```
lgfa29, line 118: Hum...I don't quite understand why this was removed. The `eval.TargetStatus` here may contain stale data, since the eval enqueue and dequeue are async operations, so by the time this eval is handled by the worker it may not be ready anymore, so we must check again before attempting any scaling action.

lgfa29 (changes requested):
Looking good! I think we can also remove this bit of Consul config as well:
https://github.com/hashicorp/nomad-autoscaler/blob/516536afcecdc688cb6f5be4301dcf79a89c04c5/agent/config/config.go#L163-L210

Left a few comments, but I think the main blocker is a potential infinite loop in the `manager.Run` method, or maybe I'm missing something 😅

policyeval/base_worker.go
Juanadelacuesta: Does it makes sense to have it on both places?

agent/agent_test.go
```diff
@@ -33,7 +31,7 @@ func TestAgent_generateNomadClient(t *testing.T) {
 				},
 			},
 			expectedOutputEr: errors.New(`failed to instantiate Nomad client: invalid address '	': parse "\t": net/url: invalid control character in URL`),
-			name: "invalid input Nomad address", //nolint
+			name:             "invalid input Nomad address", //nolint
```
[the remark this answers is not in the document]
Juanadelacuesta: Why do you re run it instead of just letting nomad take care of starting a new autoscaler?

policyeval/base_worker.go
lgfa29: I would say so. We check on equeue to make sure the policy is still valid and avoid unnecessary work by processing evals for targets that are not ready:
https://github.com/hashicorp/nomad-autoscaler/blob/516536afcecdc688cb6f5be4301dcf79a89c04c5/policy/handler.go#L288-L309

And then recheck on dequeue to make sure the assumptions queue time still hold.

If we were to remove one of them, I would keep this one since it's where scaling actions actually happen, but the policy handler already needs to query the target, so we already have the opportunity to check the status early there, so there's no harm in doing it.

agent/agent_test.go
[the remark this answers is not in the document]
lgfa29: Hum...I didn't quite get the question. In this test case we do error and the command exists with return code `1`:
https://github.com/hashicorp/nomad-autoscaler/blob/516536afcecdc688cb6f5be4301dcf79a89c04c5/command/agent.go#L338-L341

But my comment was about removing the `//nolint` part that skips linter checks, I don't see anything there that would be worth skipping.

policy/manager.go
```diff
@@ -80,7 +79,25 @@ LOOP:
...
+				// Make sure we start the next iteration with an empty map of handlers.
+				m.lock.Lock()
+				m.handlers = make(map[PolicyID]*Handler)
+				m.lock.Unlock()
+
+				// Delay the next iteration of m.Run to avoid re-runs to start too often.
+				time.Sleep(10 * time.Second)
+				go m.Run(ctx, evalCh)
 			}
 			continue
```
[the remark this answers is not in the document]
Juanadelacuesta: The previous question was for here, not for the linter in the test jejejej ups

policyeval/base_worker.go
```diff
@@ -386,34 +380,20 @@ func newCheckHandler(l hclog.Logger, p *sdk.ScalingPolicy, c *sdk.ScalingCheckEv
...
 
-	strategyPlugin, err := h.pluginManager.Dispense(h.checkEval.Check.Strategy.Name, sdk.PluginTypeStrategy)
+	source, err := h.pluginManager.GetAPM(h.checkEval.Check.Source)
 	if err != nil {
-		return nil, fmt.Errorf(`strategy plugin "%s" not initialized: %v`, h.checkEval.Check.Strategy.Name, err)
-	}
-	strategyInst, ok = strategyPlugin.Plugin().(strategy.Strategy)
-	if !ok {
-		return nil, fmt.Errorf(`"%s" is not a strategy plugin`, h.checkEval.Check.Strategy.Name)
+		return nil, fmt.Errorf(`"%s" unable to get APM source`, h.checkEval.Check.Source)
```
lgfa29, line 388:
```suggestion
		return nil, fmt.Errorf("failed to dispense APM plugin: %v", err)
```

policyeval/base_worker.go
```diff
@@ -386,34 +380,20 @@ func newCheckHandler(l hclog.Logger, p *sdk.ScalingPolicy, c *sdk.ScalingCheckEv
...
 	}
 
 	// Query check's APM.
 	// Wrap call in a goroutine so we can listen for ctx as well.
 	apmQueryDoneCh := make(chan interface{})
```
lgfa29, line 393: Not part of your PR, but reading it again this should've been a context 😅

policyeval/base_worker.go
```diff
@@ -445,8 +425,13 @@ func (h *checkHandler) start(ctx context.Context, currentStatus *sdk.TargetStatu
 	}
 
 	// Calculate new count using check's Strategy.
+	strategy, err = h.pluginManager.GetStrategy(h.checkEval.Check.Strategy.Name)
+	if err != nil {
+		return nil, fmt.Errorf(`"%s" unable to get strategy`, h.checkEval.Check.Strategy.Name)
```
lgfa29, line 430:
```suggestion
		return nil, fmt.Errorf("failed to dispense strategy plugin: %v", err)
```

sdk/eval.go
```diff
@@ -22,13 +22,13 @@ type ScalingEvaluation struct {
...
 // level for safe usage throughout the scaling evaluation phase.
-func NewScalingEvaluation(p *ScalingPolicy, status *TargetStatus) *ScalingEvaluation {
+func NewScalingEvaluation(p *ScalingPolicy) *ScalingEvaluation {
 
 	// Create the base eval.
 	eval := ScalingEvaluation{
 		ID:           uuid.Generate(),
 		Policy:       p,
-		TargetStatus: status,
+		TargetStatus: &TargetStatus{},
```
lgfa29, line 31: Is this field even used anymore? If not we could just delete it.

policy/manager.go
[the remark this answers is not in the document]
lgfa29: Ahh I see 😄 

We could go that route, and I'm usually for failing loudly, but in this case it would assume the autoscaler to be running in an orchestrated environment (which is a fair assumption), but could also cause scaling actions to fail or be cancelled mid-action, and this actions can take a while to finish.

Since the workers and the policy manager work independently, safely exiting here would require all scaling actions to be paused and wait for those in-flight to continue. During this time any policy change would be ignored, and we would not execute new scaling actions.

lgfa29 (approved):
Looks great! I think that missing error return is the main thing to be fixed, but I don't want to block you.

policy/manager.go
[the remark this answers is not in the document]
Juanadelacuesta: ah, its about maintaining the current actions , got it

policy/handler.go
```diff
@@ -215,11 +214,42 @@ func (h *Handler) handleTick(ctx context.Context, policy *sdk.ScalingPolicy) (*s
...
+	if !policy.Enabled {
+		h.log.Debug("policy is not enabled")
+		return nil, nil
+	}
+
+	target, err := h.pluginManager.GetTarget(policy.Target)
 	if err != nil {
-		return nil, err
+		h.log.Warn("failed to get target", "error", err)
+		return nil, nil
```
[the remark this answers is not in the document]
Juanadelacuesta: There is a check for nil eval right after this, but there is no harm on visibility

policy/manager.go
```diff
@@ -33,58 +33,87 @@ type Manager struct {
...
+		m.handlers = make(map[PolicyID]*Handler)
+		m.lock.Unlock()
+
+		// Delay the next iteration of m.Run to avoid re-runs to start too often.
+		time.Sleep(10 * time.Second)
 	}
+}
 
-LOOP:
+func (m *Manager) followPolicies(ctx context.Context, evalCh chan<- *sdk.ScalingEvaluation) error {
```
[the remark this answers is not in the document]
Juanadelacuesta: make sense :)

policy/manager.go
```diff
@@ -33,58 +33,87 @@ type Manager struct {
...
-
 	// Start the metrics reporter.
-	go m.periodicMetricsReporter(monitorCtx, m.metricsInterval)
+	go m.periodicMetricsReporter(ctx, m.metricsInterval)
+
+	for {
+		// Create a separate context so we can stop the goroutine monitoring the
+		// list of policies independently from the parent context.
+		monitorCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
```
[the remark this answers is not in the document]
Juanadelacuesta: we can move the cancel at the end of every iteration

policy/manager.go
```diff
@@ -33,58 +33,87 @@ type Manager struct {
 	// metricsInterval is the interval at which the agent is configured to emit
 	// metrics. This is used when creating the periodicMetricsReporter.
 	metricsInterval time.Duration
+
+	policyIDsCh    chan IDMessage
+	policyIDsErrCh chan error
```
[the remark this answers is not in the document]
Juanadelacuesta: on it!

sdk/eval.go
Juanadelacuesta: I think it is used to pass the status when executing the evaluation, let me check

policyeval/base_worker.go
Juanadelacuesta: I tried to made it obvious, because otherwise it just looks like a duplication :)

sdk/eval.go
Juanadelacuesta: You are right, is not used anywhere, let me remove it

PR #649, opened 2023-06-28
Leader election using locks

Juanadelacuesta:
Add mechanism to run on high availability by using a lock to maintain  the synchrony among instances.  Currently is a free standing package that depends on the future implementation of the lock CLI on nomad.

It is still pending on exposing some metrics and error handling

For more information, refer to the RFC: [[NMD-179] Leader election using locking mechanism for the Autoscaler](https://go.hashi.co/rfc/nmd-179)


ha/lock.go
```diff
@@ -0,0 +1,132 @@
+package ha
```
tgross, line 1: The copywrite bot is going to require a copyright header on the new files in this PR.

ha/lock.go
```diff
@@ -0,0 +1,132 @@
...
+	hc.wait(ctx)
+
+	waitTimer := time.NewTimer(hc.waitPeriod)
+	defer waitTimer.Stop()
+
+	for {
+		lockID, err := hc.lock.Acquire(ctx, hc.ID)
+		if err != nil {
+			// TODO: What to do with fatal errors?
+			hc.logger.Error("unable to get lock", err)
```
tgross, line 63: Is failing to get the lock fatal, or should we fall thru and reset the timer so that we can pick up the lock if another autoscaler drops it?

ha/lock.go
```diff
@@ -0,0 +1,132 @@
...
+				// Give the protected function some time to return before potentially
+				// running it again.
+				hc.wait(ctx)
+			}
+		}
+
+		if !waitTimer.Stop() {
+			<-waitTimer.C
+		}
+		waitTimer.Reset(hc.waitPeriod)
```
tgross, line 90: If the shutdown context fires while we're waiting, I think this blocks shutdown for the wait period? 

In Nomad core we avoid this problem in a few places by starting with a very small wait period (see [`funcs.go#L369-L394`](https://github.com/hashicorp/nomad/blob/main/helper/funcs.go#L369-L394)) and then resetting it to the at the end of the loop as you've done here. Alternately, you could move the `select` up to the top of the loop and initialize the timer to `randomDelay`. That'd let you drop both this and `hc.wait` at the top of this method.

ha/lock.go
Juanadelacuesta: Failing to get the lock is not fatal and will not yield an error, just an empty lockID, and it will happen regularly on every instance that is not the leader, as they will attempt to get the lock periodically. Im thinking  more about "real" errors while talking to the nomad agent

ha/lock.go
Juanadelacuesta: For a timer to be successfully reseted, it needs to be stopped and drained first, the stop() function reports if draining is needed but the timer is stoped already:  https://pkg.go.dev/time#Timer.Stop.
I think a ticker is better suited here, Im making the change

ha/lock.go
```diff
@@ -0,0 +1,127 @@
...
+			if err != nil {
+				hc.logger.Debug("lease lost", err)
+				cancel()
+				// Give the protected function some time to return before potentially
+				// running it again.
+				hc.wait(ctx)
+			}
+		}
+
+		waitTicker = time.NewTicker(hc.waitPeriod)
```
tgross, line 87: This is creating a new ticker on each pass thru the loop, and I'm pretty sure it leaks the ticker's goroutine because it's not stopped when it goes out of scope like this. I think we can do:

```suggestion
		waitTicker.Reset(hc.waitPeriod)
```

Which handles the stop safely as you noted.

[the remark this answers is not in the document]
Juanadelacuesta: I tried that one, but the waitTicker.C channel is buffered and if the a tick has already gone before, we would have had to drain it, just like the timer, and in the case of an instance that was previously holding the lock, this is 100% the case, the options were to leave it as it was or to create a new ticker, like I did. Im not aware of any possible memory problems this can bring, but if you do, please tell me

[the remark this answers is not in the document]
Juanadelacuesta: I forgot to stop the ticker before rewriting it, updated it

ha/lock_test.go
```diff
@@ -0,0 +1,334 @@
...
+	time.Sleep(4 * time.Millisecond)
+	/*
+		After 4 ms more (4 ms total):
+		* hac2 should  not have tried to acquire the lock.
+		* hac1 should have the lock and the service should be running.
+		* The first lease is not over yet, no calls to renew should have been made.
+	*/
+	must.True(t, l.locked)
+	must.Eq(t, 1, l.acquiresCalls[hac1.ID])
+	must.Eq(t, 0, l.acquiresCalls[hac2.ID])
```
lgfa29, line 139: These calls that check `l` are by-passing the `mu` Mutex in the lock. It may be better to have safe getter methods.

ha/lock.go
```diff
@@ -0,0 +1,127 @@
...
+			hc.logger.Error("unable to get lock", err)
+		}
+
+		if lockID != "" {
+			hc.logger.Debug("lock acquired, ID", lockID)
+			funcCtx, cancel := context.WithCancel(ctx)
+			defer cancel()
+
+			// Start running the lock protected function
+			go protectedFunc(funcCtx)
```
lgfa29, line 73: I've been thinking about what would happen if this goroutine were to exit prematurely. I think we would end-up with an instance holding the lock forever but not actually doing anything? 🤔 

I'm not sure how to solve this, maybe some kind of health check mechanism so we only renew the lease if the function is still active? But maybe that's too much overhead for now, and even health checks can lie.

ha/lock_test.go
```diff
@@ -0,0 +1,334 @@
...
+	"errors"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/hashicorp/nomad/helper/testlog"
+	"github.com/shoenig/test/must"
+)
+
+var lease = 10 * time.Millisecond
```
lgfa29, line 17:
```suggestion
const defaultLockLease = 10 * time.Millisecond
```

This value is still global in the package, so a generic name could clash with another variable and cause some unexpected behaviours.

ha/lock.go
```diff
@@ -0,0 +1,127 @@
...
+
+		case <-renewTicker.C:
+			hc.logger.Debug("renewing lease")
+			err := hc.lock.Renew(ctx)
+			if err != nil {
+				return err
+			}
+		}
+	}
+
```
lgfa29, line 117:
```suggestion
```

ha/lock_test.go
```diff
@@ -0,0 +1,334 @@
...
+
+	must.Eq(t, 0, l.renewsCounter)
+
+	must.Eq(t, 1, s.startsCounter)
+	must.StrContains(t, hac1.ID, s.starterID)
+
+	time.Sleep(6 * time.Millisecond)
+	/*
+		After 6 ms more (10 ms total):
+		* hac2 should have tried to acquire the lock at least once.
```
lgfa29, line 149: This is because the random delay is over right?

lgfa29 (changes requested):
Looking great! Excited to see where this is going 😄 

I think the tests not running in CI is the biggest blocker, most of the comments are me trying to understand stuff 😬

ha/lock_test.go
Juanadelacuesta: yes!

ha/lock.go
```diff
@@ -0,0 +1,127 @@
...
+		waitPeriod:    time.Duration(float64(lease) * waitFactor),
+		ID:            uuid.Generate(),
+		randomDelay:   time.Duration(rn) * time.Millisecond,
+	}
+
+	return &hac
+}
+
+func (hc *HAController) Start(ctx context.Context, protectedFunc func(ctx context.Context)) error {
+	hc.logger.Named(hc.ID)
```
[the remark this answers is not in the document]
Juanadelacuesta: you are right! let me fix it

ha/lock_test.go
Juanadelacuesta: I thought about it, and even the mu feels like a bit of an overkill for the this simple test?

ha/lock.go
```diff
@@ -0,0 +1,127 @@
...
+func NewHAController(l lock, logger log.Logger, lease time.Duration) *HAController {
+	logger = logger.Named("ha_mode")
+
+	rn := rand.New(rand.NewSource(time.Now().Unix())).Intn(100)
+	hac := HAController{
+		lock:          l,
+		logger:        logger,
+		renewalPeriod: time.Duration(float64(lease) * renewalFactor),
+		waitPeriod:    time.Duration(float64(lease) * waitFactor),
+		ID:            uuid.Generate(),
```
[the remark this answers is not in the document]
Juanadelacuesta: No, but my thought was sometimes logs are aggregated and and ID can facilitate tracking one instance, and also for testing :s

ha/lock.go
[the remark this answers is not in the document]
tgross: LGTM!

I had to check for myself because I'm paranoid about this sort of thing and https://go.dev/play/p/kdMH0bva-5Z and https://go.dev/play/p/EV6J--iRBLf are equivalent in terms of the resulting stack traces; neither leaks.

ha/lock_test.go
Juanadelacuesta: It ended up being necessary to avoid trigger the race condition on the tests, so good thinking!

ha/lock.go
Juanadelacuesta: we can think about inverting the order and calling the lock as a routine and the main service as we always did?

ha/lock.go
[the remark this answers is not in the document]
Juanadelacuesta: Double checks are always welcomed!

ha/lock.go
Juanadelacuesta: Also, the autoscaler can cancel its context on error and we can catch that on the ha controller to return, what do you think? @lgfa29

agent/agent.go
```diff
@@ -180,14 +176,14 @@ func (a *Agent) stop() {
 }
 
 // generateNomadClient creates a Nomad client for use within the agent.
```
jrasell, line 178:
```suggestion
// GenerateNomadClient creates a Nomad client for use within the agent.
```

agent/config/config.go
```diff
@@ -355,6 +363,10 @@ const (
 	// defaultPolicyWorkerAckTimeout is the default time limit that a policy
 	// eval must be ACK'd.
 	defaultPolicyEvalAckTimeout = 5 * time.Minute
+
+	// defaultLockPath is the default path used for the lock that syncs the leader
+	// election.
+	defaultLockPath = "autoscaler/lock"
```
jrasell, line 369: nit: use the full name of the project:
```suggestion
	defaultLockPath = "nomad-autoscaler/lock"
```

agent/config/config.go
```diff
@@ -269,6 +272,11 @@ type Telemetry struct {
 	CirconusBrokerSelectTag string `hcl:"circonus_broker_select_tag,optional"`
 }
 
+type HighAvailability struct {
+	Enable   bool   `hcl:"high_availability"`
+	LockPath string `hcl:"lock_path,optional"  json:"-"`
+}
```
jrasell, line 278: Do we want to expose more options, such as the TTL and LockDelay? 

Checking with the Nomad config docs, we seem to prefer `enabled` which we might want to follow.

The HCL struct tag on the `enable` parameter seems wrong.

command/agent.go
```diff
@@ -259,6 +260,18 @@ Telemetry Options:
...
+  -high-availability
+	On cases when multiple instances of the autoscaler need to be run at the same
+	time, the high-availability option triggers a leader election using a lock 
+	for sync among the different lock instances. It defaults to false.
+  
+  -lock-path
+    When using the high-availability mode, the path to the lock to be used for the
+	leader election can be provided using the lock-path flag. The same path must 
+	be provided to every instance of the autoscaler in order to be included in the 
+	election.
```
jrasell, line 274: Other flags are suffixed with their component name and I wonder if we want to follow suit here? That would mean the flags become something like `high-availability-enabled` and `high-availability-lock-path`. I also think the flags names should mimic the config block names, to make it easier. for operators to understand.

command/agent.go
```diff
@@ -338,10 +351,43 @@ func (c *AgentCommand) Run(args []string) int {
...
 	ctx := context.Background()
 
+	// Generate the Nomad client.
+	if err := c.agent.GenerateNomadClient(); err != nil {
+		logger.Error("failed to start the nomad client", "error", err)
+		return 1
+	}
+
+	if parsedConfig.HighAvailability.Enable {
+		logger.Info("Running on ha mode")
```
jrasell, line 361: I wonder if we also want to log the configuration options passed?
```suggestion
		logger.Info("running in HA mode")
```

command/agent.go
```diff
@@ -338,10 +351,43 @@ func (c *AgentCommand) Run(args []string) int {
...
+		ttl := 5 * time.Minute
+		asLock := api.Variable{
+			Path: "autoscaler/lock/",
```
jrasell, line 364: Should this be using the parsed configuration option?

command/agent.go
```diff
@@ -338,10 +351,43 @@ func (c *AgentCommand) Run(args []string) int {
 
 	ctx := context.Background()
 
+	// Generate the Nomad client.
+	if err := c.agent.GenerateNomadClient(); err != nil {
+		logger.Error("failed to start the nomad client", "error", err)
+		return 1
+	}
+
+	if parsedConfig.HighAvailability.Enable {
```
jrasell, line 360: Just for readability, I wonder if we could use a switch statement to control this?
```
	switch parsedConfig.HighAvailability.Enable {
	case true:
	default:
	}
```

jrasell (changes requested):
Looking good and most of my comments are around UX and style, so we can catch in Slack if that's easier than GitHub :D

It would also be nice to add some tests to cover the new config parsing addition.

command/agent.go
Juanadelacuesta: Done!

agent/config/config.go
```diff
@@ -355,6 +380,16 @@ const (
...
+
+	// defaultLockPath is the default path used for the lock that syncs the leader
+	// election.
+	defaultLockPath = "nomad-autoscaler/lock"
+	// defaultLockTTL is the default ttl used for the lock that syncs the leader
+	// election.
+	defaultLockTTL = 5 * time.Minute
+	// defaultLockDelay is the default lockDelay used for the lock that syncs the leader
+	// election.
+	defaultLockDelay = 5 * time.Minute
```
jrasell, line 392: This means an autoscaler "cluster" could lose leadership for 10mins before another takes over, which seems a little long. What about something like 60s and 30s?

command/agent.go
```diff
@@ -259,6 +260,18 @@ Telemetry Options:
   -telemetry-circonus-broker-select-tag
     A tag which is used to select a broker ID when an explicit broker ID is not
     provided.
+  
+High Availability Options:
```
jrasell, line 264: Do we want to support the other config parameters on the CLI also?

jrasell (approved):
Approving, with a final few questions.

command/agent.go
Juanadelacuesta: They are defined, I forgot to update the docs, let me do it

PR #660, opened 2023-07-09
Add extra telemetry to monitor failures

the-nando:


policy/handler.go
```diff
@@ -228,6 +229,7 @@ func (h *Handler) handleTick(ctx context.Context, policy *sdk.ScalingPolicy) (*s
 
 	status, err := target.Status(policy.Target.Config)
 	if err != nil {
+		metrics.IncrCounter([]string{"policy", "target_status", "failure_count"}, 1)
```
lgfa29, line 232: We have a standard set of metrics already of these kind of plugin invocations:
https://developer.hashicorp.com/nomad/tools/autoscaling/telemetry#scaling-metrics

But it's only for a few actions, so it would good to expand to others as well.

Although having to remember to emit these metrics all the time is kind of annoying and easy to forget, so I think we can do something better.

Instead of leaving it to the caller to remember to handle this, the way I'm thinking about is to have the plugin manager return an instance of the requested plugin wrapped in a struct that handles the metric.

Here is a quick prototype of how this would look like for a target plugin: https://github.com/hashicorp/nomad-autoscaler/commit/4340235695b36ddb97456a376c16f9dda64d8ba5

This way, emitting the metrics is completely transparent to callers and we can ensure the metrics are always being handle.

What do you think of this approach? 

(cc @Juanadelacuesta and @jrasell as you've been digging into the autoscaler recently)

jrasell: @lgfa29 that approach makes sense to me and acts like a middleware/wrapper which should make plugin developer life easier.

lgfa29: Thanks for the extra input 😄 

@the-nando do you think you would be able to implement something like this?

lgfa29: Thank you! Feel free to reach out if you need any help 🙂

PR #693, opened 2023-08-07
deps: update golang.org/x/net to resolve security scan failure

jrasell:


lgfa29 (approved):
Oof..what an annoying process, thanks for taking care of it!

PR #724, opened 2023-10-02
Handle jobs being deleted and re-created

gulducat:
I noticed while working on another issue that `/v1/scaling/policy/{policy-id}` got started again after job purge/re-create, but `/v1/job/{job-id}/scale` did not.  I was surprised not to find an open issue for this, but maybe I missed it.

Instead of trying to un-stop the existing handler, I opted to just start a new one.  Let me know if this is horribly inefficient in some way that I'm not anticipating.


jrasell (approved):
LGTM.

I think we just need a changelog entry.

PR #733, opened 2023-10-09
Recover file policies after manager soft-restarts (from Nomad API errors)

### LINKED ISSUE #519: Failed to get policy (EOF)

gulducat:
Fixes #519 "failed to read policy in time" errors, at least those which occur due to Handlers getting re-started, but then not receiving file policies, because they already exist from the point of view of the file policy Source.

When the policy Manager creates and runs new Handlers, the new Handler needs to receive a policy on its h.ch, or h.handleTick() will have a nil currentPolicy, which results in "timeout: failed to read policy in time"

So, each time Handler asks a Source to MonitorPolicy(), the Source needs to send a policy back to the Handler (via MonitorPolicyReq.ResultCh: h.ch), even if the Source already knows the policy from a previous run.

---

I want to acknowledge that, after I banged my head against it for awhile, I found that @Seros totally had a handle (heh) on it already in #625, but here I am anyway, having addressed the bug in a slightly different way.

The only real difference in behavior here is to retain the changed-or-not distinction when a reload signal is issued, which is probably trivial performance-wise, but I think nice to have in the logs.

Regardless, I've added @Seros as a co-author, since they also fixed it, months ago. :yum:  Also feel free to verify this change, if you'd like, and let me know how it goes!

### DIFF
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -133,7 +133,8 @@ func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq)
 
 		p, _, err := s.handleIndividualPolicyRead(req.ID, file, name)
 		if err != nil {
-			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy %s", req.ID), req.ErrCh)
+			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy %s: %w", req.ID, err), req.ErrCh)
+			return
 		}
 		s.policyMap[req.ID] = &filePolicy{file: file, name: name, policy: p}
 		// We must send to ResultCh each time a Handler invokes this method,
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -104,8 +104,11 @@ func (s *Source) ReloadIDsMonitor() {
 	<-s.reloadCompleteCh
 }
 
-// MonitorPolicy satisfies the MonitorPolicy function of the policy.Source
-// interface.
+// MonitorPolicy reads policy from a file on disk and writes it to req.ResultCh.
+// Any error doing so will be sent to req.ErrCh.
+// If MonitorPolicy receives on s.ReloadCh, it will re-check the file on disk,
+// and write again to req.ResultCh if the policy has changed.
+// Note: MonitorPolicy should only return when ctx is done.
 func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq) {
 
 	// Close channels when done with the monitoring loop.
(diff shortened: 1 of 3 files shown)


policy/file/source.go
```diff
@@ -214,15 +215,15 @@ func (s *Source) handleIndividualPolicyRead(ID policy.PolicyID, path, name strin
...
 	}
 
 	// Check the new policy against the stored. If they are the same, and
 	// therefore the policy has not changed indicate that to the caller.
 	if reflect.DeepEqual(newPolicy, val.policy) {
-		return nil, nil
+		return newPolicy, false, nil
 	}
-	return newPolicy, nil
+	return newPolicy, true, nil
```
jrasell, line 226: Just spitballing, but this could potentially be written:
```
return newPolicy, !reflect.DeepEqual(newPolicy, val.policy), nil
```

jrasell (approved):
LGTM!

policy/file/source.go
```diff
@@ -131,14 +131,15 @@ func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq)
...
-		p, err := s.handleIndividualPolicyRead(req.ID, file, name)
+		p, _, err := s.handleIndividualPolicyRead(req.ID, file, name)
 		if err != nil {
-			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy %s", req.ID), req.ErrCh)
-		}
-		if p != nil {
-			req.ResultCh <- *p
-			s.policyMap[req.ID] = &filePolicy{file: file, name: name, policy: p}
+			policy.HandleSourceError(s.Name(), fmt.Errorf("failed to get policy %s: %w", req.ID, err), req.ErrCh)
+			return
```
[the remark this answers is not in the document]
gulducat: how's 30ee9bb3813cbe4c8e267df645ba4c19b7339887 look to ya?

policy/file/source_test.go
```diff
@@ -51,3 +55,158 @@ func TestSource_getFilePolicyID(t *testing.T) {
...
+
+	// running MonitorPolicy() twice should have the same result.
+	for _, n := range []string{"round one", "round two"} {
+
+		// not using sub-tests here, because if round one fails,
+		// there's no sense running round two.
+		// this little helper gives context for any failures.
+		fatal := func(msg string, args ...any) {
+			t.Helper()
+			t.Fatalf(n+": "+msg, args...)
```
lgfa29, line 94: Probably not too relevant for a test, but I think this can cause the `go s.MonitorPolicy(ctx, req)` goroutine to leak. You could register a `t.Cleanup(cancel)` below to make sure the context is cancelled when the test exits.

policy/file/source_test.go
lgfa29, line 1: Nice job with these test!

lgfa29 (approved):
Good work getting to the bottom of this. Also reiterating the big thanks to @Seros for fixing it a while ago!

policy/file/source.go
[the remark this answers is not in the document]
lgfa29: Looks good! The overall design of error handling feels questionable in hindsight, but that's my fault 😬

PR #807, opened 2023-12-21
aws: add `scale_in_protection` configuration

### LINKED ISSUE #650: aws-asg plugin: Respect scale-in protection
A common operation in our infrastructure is to replace our ec2 workers. We do this by doubling the asg size and scale protecting the new workers. We then migrate work to the new workers and scale back down to the original size. However, when we tried this while horizontal scaling was enabled it terminated the new scale protected instances.

Reading the aws docs on scale protection (https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-instance-protection.html) this makes sense. The plugin uses a terminate call that is documented as not honoring the flag and we have it set to chose the instance with the least work.

I propose that the plugin should honor this flag. I think this can be accomplished by skipping instances that have `ProtectedFromScaleIn` set to true, the same way it skips unhealthy or not in service instances.

lgfa29:
When `scale_in_protection` is enabled, the AWS target plugin skips instances that are protected from scale in operations[1].

Closes #650 

[1] https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-instance-protection.html

### DIFF
--- a/plugins/builtin/target/aws-asg/plugin/aws.go
+++ b/plugins/builtin/target/aws-asg/plugin/aws.go
@@ -96,8 +96,7 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, asg *types.AutoScalingGroup
 		DesiredCapacity:      aws.Int32(int32(count)),
 	}
 
-	// Ignore the response from Send() as its empty.
-
+	// Ignore the response from UpdateAutoScalingGroup() as its empty.
 	_, err := t.asg.UpdateAutoScalingGroup(ctx, &input)
 	if err != nil {
 		return fmt.Errorf("failed to update Autoscaling Group: %v", err)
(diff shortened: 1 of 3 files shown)


plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -116,15 +116,36 @@ func (t *TargetPlugin) scaleIn(ctx context.Context, asg *types.AutoScalingGroup,
...
+			log.Debug("skipping instance",
+				"instance_id", *inst.InstanceId,
+				"health_status", *inst.HealthStatus,
+				"lifecycle_state", inst.LifecycleState,
+				"protected_from_scale_in", *inst.ProtectedFromScaleIn,
+			)
+			continue
 		}
+
+		log.Debug("foud eligible instance", "instance_id", *inst.InstanceId)
```
jrasell, line 147:
```suggestion
		log.Debug("found eligible instance", "instance_id", *inst.InstanceId)
```

plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -116,15 +116,36 @@ func (t *TargetPlugin) scaleIn(ctx context.Context, asg *types.AutoScalingGroup,
...
+	// scale_in_protection.
+	scaleInProtection := t.scaleInProtectionEnabled
+	if str, ok := config[configKeyScaleInProtection]; ok {
+		b, err := strconv.ParseBool(str)
+		if err != nil {
+			return fmt.Errorf("failed to parse %s value from policy: %v", configKeyScaleInProtection, err)
+		}
+		scaleInProtection = b
+	}
+	log = log.With("scale_in_protection", scaleInProtection)
```
jrasell, line 129: Is it worth moving the sub-logger creation on L117 to here, to avoid setting up two sub-loggers?

plugins/builtin/target/aws-asg/plugin/aws.go
```diff
@@ -116,15 +116,36 @@ func (t *TargetPlugin) scaleIn(ctx context.Context, asg *types.AutoScalingGroup,
 	// would like on all log lines.
 	log := t.logger.With("action", "scale_in", "asg_name", *asg.AutoScalingGroupName)
 
+	// Check if policy overrides the plugin configuration for
+	// scale_in_protection.
+	scaleInProtection := t.scaleInProtectionEnabled
+	if str, ok := config[configKeyScaleInProtection]; ok {
+		b, err := strconv.ParseBool(str)
+		if err != nil {
+			return fmt.Errorf("failed to parse %s value from policy: %v", configKeyScaleInProtection, err)
```
jrasell, line 125:
```suggestion
			return fmt.Errorf("failed to parse %s value from policy: %w", configKeyScaleInProtection, err)
```

jrasell (approved):
LGTM with a couple of minor inline suggestions or queries.

plugins/builtin/target/aws-asg/plugin/aws.go
lgfa29: Ah, good call. I didn't notice it wasn't being used in between.

Issue #65, opened 2020-03-30
Support multiple namespaces


lgfa29:
Currently each agent can only monitor jobs and policies from a single namespace. For facilitate operating the Autoscaler in environments with multiple namespaces the agent configuration should accept a list of namespaces, potentially with support for wildcards.

pop:
I would really like this feature. I'm in a multi-namespace environment and this would simplify things quite a bit.

javier-diaz-herrera:
Currently, in our project, we need to specify several namespaces on the autoscaler.
Could you tell us if you are going to develop this functionality? It would be very useful to us.

javier-diaz-herrera:
Could you tell us if you will develop this feature?

jorgemarey:
Hi, 

We are using nomad-autoscaler v0.3.7 and using this configuration the autoscaler gets the scaling policies for all namespaces. I don't know if this works for some of you.

nomad {
....
  namespace = "*"
}

javier-diaz-herrera:
Thanks for the help @jorgemarey !! :)

We are using nomad-autoscaler v0.3.7 as well and use de apm-nomad.

The autoscaler shows the following logs:

The autoscaler is able to retrieve the autoscaling policies for each namespace:

`2023-07-04T11:14:18.871Z [DEBUG] internal_plugin.nomad-apm: expanded query: from=taskgroup_avg_memory-allocated/builds/builds to="&plugin.taskGroupQuery{metric:"memory-allocated", job:"builds", group:"builds", operation:"avg"}"`

but it is not able to retrieve the job metrics:

`2023-07-04T11:14:18.876Z [WARN]  policy_eval.worker: failed to run check: id=72abd79a-ed54-a6cb-a597-befb37f7ae7e policy_id=72e939b6-e602-b500-2176-99b09d636536 queue=horizontal target=nomad-target check=memory-usage on_error="" on_check_error="" error="failed to query source: failed to get total alloacted memory for taskgroup: failed to get info for job: Unexpected response code: 404 (job not found)"`

Any idea?

qk4l:
Hi could any one provide current status of support?

Based documentation autoscaling does not support several namespaces.

https://developer.hashicorp.com/nomad/tools/autoscaling/agent#nomad-namespaces

lgfa29:
Hi @javier-diaz-herrera 👋 

[...]

PR #808, opened 2023-12-21
nomad: specify namespace when querying group

lgfa29:
The Nomad Autoscaler agent can only target a single namespace, but when using the wildcard (`*`) value it is possible to monitor policies for all namespaces.

With this configuration, the Nomad APM plugin must specify the job namespace when querying for information as the agent is configured with the wildcard.

Ref: https://github.com/hashicorp/nomad-autoscaler/issues/65#issuecomment-1620071860

### DIFF
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -108,8 +108,14 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
 	// If the target is a Nomad job task group, format the query in the
 	// expected manner.
 	if t.IsJobTaskGroupTarget() {
-		c.Query = fmt.Sprintf("%s_%s/%s/%s",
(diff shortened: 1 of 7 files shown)


policy/policy.go
```diff
@@ -108,8 +108,14 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
-		c.Query = fmt.Sprintf("%s_%s/%s/%s",
-			nomadAPM.QueryTypeTaskGroup, c.Query, t.Config[sdk.TargetConfigKeyTaskGroup], t.Config[sdk.TargetConfigKeyJob])
+		c.Query = fmt.Sprintf(
+			"%s_%s/%s/%s@%s",
+			nomadAPM.QueryTypeTaskGroup,
+			c.Query,
+			t.Config[sdk.TargetConfigKeyTaskGroup],
+			t.Config[sdk.TargetConfigKeyJob],
+			t.Config[sdk.TargetConfigKeyNamespace],
+		)
```
lgfa29, line 118: This is an internal query format, so I wouldn't consider it a breaking change.

jrasell (approved):
LGTM

PR #809, opened 2023-12-21
build: copy binary to `$GOPATH` on `make dev`

lgfa29:
When developing, it's useful to have the `nomad-autoscaler` binary available in `$GOPATH` so it's easier run them.


jrasell (approved):
LGTM

PR #810, opened 2023-12-21
cluster: support `node_pool` value as filter

### LINKED ISSUE #749: cluster scaling: allow nodes to be grouped via node pools to form scaling pools
Nomad v1.6.0 added the concept of node pools as a way to logically group or separate nodes. It would be useful if the Nomad Autoscaler could perform horizontal cluster scaling on nodes, where the grouping is performed on a node pool basis.

lgfa29:
Add the `node_pool` node selector option to filter nodes by their `node_pool` value.

Closes #749


jrasell (approved):
LGTM

Issue #672, opened 2023-07-19
keep scaling when nodes are draining


janory:
Hi! 👋 

We recently started to use the Nomad Autoscaler agent and we really like it. 🚀 
We are using the Autoscaler with the `Nomad APM`, `aws-asg` target and `target-value` strategy plugins.

We have multiple long running (1-45 minutes) batch jobs on our nodes and when a scale in action happens the drain event won't finish until the last batch job completes on the node.

This leads to constant warning messages like this:
```
[WARN]  internal_plugin.aws-asg: node pool status readiness check failed: error="node 58952e93-5059-0625-23a5-61520e3cf126 is draining"
```
because the Autoscaler implicitly checks the ASG target's status for each tick (`handleTick -> generateEvaluation -> Status -> IsPoolReady -> FilterNodes -> if node.Drain`).

Based on the comment [here](https://github.com/hashicorp/nomad-autoscaler/blob/b33f10d465e8ec030e5bcfe4dfd2b5082b91b853/sdk/helper/scaleutils/node_identifier.go#L92) and also based on what we are experiencing the Autoscaler stops any further scaling actions until all draining activities are completed.

This is an issue for us, because in worst case scenario the long running batch jobs will prevent us scaling for 45 minutes.

Would it be possible to add a config for the `idFn` [function](https://github.com/hashicorp/nomad-autoscaler/blob/b33f10d465e8ec030e5bcfe4dfd2b5082b91b853/sdk/helper/scaleutils/node_identifier.go#L71) to filter out draining nodes and keep scaling?

We would also like to better understand what are the risks of scaling a cluster which has draining nodes and why such a cluster is considered unstable.

janory:
I was thinking about something like this: https://github.com/hashicorp/nomad-autoscaler/pull/679
Although this alone probably won't be enough, because even if this part passes, the [processLastActivity](https://github.com/hashicorp/nomad-autoscaler/blob/1d00a3c0dd5ab2b3a73cf5f6b4b18f270977fbaa/plugins/builtin/target/aws-asg/plugin/plugin.go#L230) call would set the `Ready` flag to `false`.

tgross:
Hi @janory!

> We would also like to better understand what are the risks of scaling a cluster which has draining nodes and why such a cluster is considered unstable.

[...]

PR #811, opened 2023-12-22
aws: add experimental configuration

lgfa29:
Add experimental configuration values to the AWS ASG plugin to allow cluster operators to bypass some of the safety checks enforced by the Autoscaler.

I'm still unsure about these changes, so I'm open to suggestions, but letting users experiment with them could generate useful information about the assumptions we make.

Ref: https://github.com/hashicorp/nomad-autoscaler/issues/672

### DIFF
--- a/sdk/helper/scaleutils/node_identifier.go
+++ b/sdk/helper/scaleutils/node_identifier.go
@@ -5,6 +5,7 @@ package scaleutils
 
 import (
 	"fmt"
+	"strconv"
 
 	multierror "github.com/hashicorp/go-multierror"
 	errHelper "github.com/hashicorp/nomad-autoscaler/sdk/helper/error"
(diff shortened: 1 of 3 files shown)


sdk/helper/scaleutils/node_identifier.go
```diff
@@ -125,3 +131,59 @@ func filterOutNodeID(n []*api.NodeListStub, id string) []*api.NodeListStub {
...
+			)
+		}
+
+		opts.ignoreDrain = ignoreDrain
+	}
+
+	return opts, nil
+}
+
+func (n *NodeFilterOptions) IgnoreInit() bool {
```
Juanadelacuesta, line 177: `NodeFilterOptions` is a structure only meant to store configurations values, we wont change them in the methods, is it necessary to make it a pointer? We can avoid nil pointers

Juanadelacuesta (approved):
Just one question but nothing major :)

sdk/helper/scaleutils/node_identifier.go
lgfa29: Since I'm still not sure how these options are going to behave the pointer gives us more flexibility to modify things for now. We can just pass and check for `nil` to disable any functionality we don't want to modify at this point.

PR #812, opened 2023-12-22
broker: prevent overlap of policy evaluations

lgfa29:
When a policy is sent for evaluation by a handler, it is placed into the eval broker queue. If the queue already has an eval for the policy, it discards one of them, keeping the most recent one. This is used to guarantee that a policy is never processed in parallel by different workers.

The code had the `return` statement in the wrong scope, causing a duplicate eval to be enqueued even when the policy was already been processed, resulting in unexpected concurrent policy evaluations.

Closes https://github.com/hashicorp/nomad-autoscaler/issues/610

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 IMPROVEMENTS:
  * agent: Add `BlockQueryWaitTime` config option for Nomad API connectivity [[GH-755](https://github.com/hashicorp/nomad-autoscaler/pull/755)]
 
+BUG FIXES:
+ * agent: Fixed a bug that could cause scaling policies to be evaluated concurrently [[GH-812](https://github.com/hashicorp/nomad-autoscaler/pull/812)]
+
 ## 0.4.0 (December 20, 2023)
 
 FEATURES:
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 IMPROVEMENTS:
  * agent: Add `BlockQueryWaitTime` config option for Nomad API connectivity [[GH-755](https://github.com/hashicorp/nomad-autoscaler/pull/755)]
 
+BUG FIXES:
+ * agent: Fixed a bug that could cause scaling policies to be evaluated concurrently [[GH-812](https://github.com/hashicorp/nomad-autoscaler/pull/812)]
```
jrasell, line 7:
```suggestion
 * agent: Fixed a bug that could cause the same scaling policy to be evaluated multiple times concurrently [[GH-812](https://github.com/hashicorp/nomad-autoscaler/pull/812)]
```

jrasell (approved):
LGTM.

Is it possible to add a test that covers this behaviour?

lgfa29:
> LGTM.
> 
> Is it possible to add a test that covers this behaviour?

Ah yeah, good point. Expanded `TestBroker` in https://github.com/hashicorp/nomad-autoscaler/pull/812/commits/ff181cdb9e6fb0786e49d541957b6b9a5aa3eb74 to cover this.

Issue #727, opened 2023-10-04
Threshold Strategy: lower_bound behavior does not match docs


jaltavilla:
The docs state that when lower_bound is specified and upper_bound omitted, that the lower bound is non-inclusive. However, this is not the case. The comparison is always inclusive (`>=`).  The specific piece of documentation I'm referencing can be found at the end of [https://developer.hashicorp.com/nomad/tools/autoscaling/plugins/strategy/threshold](https://developer.hashicorp.com/nomad/tools/autoscaling/plugins/strategy/threshold).
> Similarly, if upper_bound is not defined, any value above lower_bound will be considered within bounds.

We discovered this discrepancy when the below policy resulted in the asg scaling up to its max.

``` 
        # If nomad indicates that the cluster is missing cpu to schedule an allocation, then we need to add an ec2 instance
        check "missing-cpu" {
            source          = "datadog"
            query           = "default_zero(sum:nomad.nomad.blocked_evals.cpu{ {{template "byRoleTags" .asg.role}} }.rollup(sum).weighted())"
            query_window    = "3m"
            group           = "blocked-cpu"
            strategy "threshold" {
                lower_bound = 0
                delta       = 1
            }
        }
```

The relevant code can be found at https://github.com/hashicorp/nomad-autoscaler/blob/3a7e8c2ef1189c68b992060716bedf24cd2da25f/plugins/builtin/strategy/threshold/plugin/plugin.go#L262. The comparison doesn't change if there is no upper bound specified. Nor does `parseBound` above alter the lower bound with something like math.Nextafter (which is probably a bad idea anyway).

Semi-related, I wonder if `parseBound` should be using `math.Inf` instead of `math.MaxFloat64`. If a metric ever returned a max float64 it would never be within the bounds of a threshold, even one that was a lower bound only. It seems unlikely, but some poor engineer would have a very confusing experience. 😆

lgfa29:
Hi @jaltavilla 👋 

Nice catch! The docs and the code are actually quite vague on the expected behaviour here...

Reading the original issue (https://github.com/hashicorp/nomad-autoscaler/issues/438) it seems like `lower_bound` always being inclusive and `upper_bound` always being exclusive (as it's currently coded) was the original design, but I can't really remember why 😅 

I think the idea was to avoid confusion if you have multiple tiers. In the example provided, a value of exactly `70` would fall into `scale_out_high_threshold_zone` and not `scale_out_low_threshold_zone`.

[...]

PR #815, opened 2023-12-22
threhsold: clarify the behaviour of bounds

lgfa29:
The `lower_bound` value is always considered inclusive and the `upper_bound` value is always exclusive. This helps prevent overlapping or a dead zone when regions start or end right next to each other.

Also use `math.Inf` instead of `MaxFloat64` to allow a value that matches `MaxFloat64` to be considered within bounds.

No changelog since it doesn't change any user-facing behaviour.

Closes https://github.com/hashicorp/nomad-autoscaler/issues/727


jrasell (approved):
LGTM

Issue #828, opened 2024-01-13
Fail to start in high availability mode when using `namespace = "*"`


livioribeiro:
When running in high availability mode, the namespace used to acquire the lock is the same declared in the `nomad` stanza in the autoscaler configuration, so using `namespace = "*"` with high availability enabled causes an error when acquiring the lock with the api returning status 403.

One possible solution would be to use for the lock the same namespace where the job is running instead of the one defined in the `nomad` stanza. I do not know a lot of Go but I believe that it can be defined in the `api.WriteOptions{}` in this line:
 
https://github.com/hashicorp/nomad-autoscaler/blob/cbd0ccb444dc4f226357b4466bccc90183be4262/command/agent.go#L411

lgfa29:
Thank you for the report @livioribeiro!

You're absolutely right, we need to specify a single namespace for the variable. But using the autoscaler job namespace is not quite reliable, as it may not be running in Nomad (we even have people running it on Kubernetes 😄).

I opened #832 to add a new config for this. You use the [`NOMAD_NAMESAPCE` env var](https://developer.hashicorp.com/nomad/docs/runtime/environment#job-related-variables) in your job to get the behaviour your described.

PR #832, opened 2024-01-17
ha: add support for specifying the lock namespace

lgfa29:
When HA mode is enabled, the leader lock must be managed in a specific namespace, but when setting the `nomad.namespace` configuration to the wildcard `*` value the Nomad API client was trying to write to this namespace value, which is not allowed.

This commit adds a new `high_availability.lock_namespace` configuration to allow operators to specify the exact lock variable namespace.

Closes https://github.com/hashicorp/nomad-autoscaler/issues/828.


Juanadelacuesta (approved):
Wow you are fast :)

PR #845, opened 2024-02-01
Fix autoscale when dry-run in set

jorgemarey:
Autoscaler tries to scale even when dry-run is configured. This happens when the current count is out of the range of the configured min/max. Currently the dry-run check is performed after this checks are executed. https://github.com/hashicorp/nomad-autoscaler/blob/0d0ad24c2381f1c13995504a56354d0e9d269eeb/policyeval/base_worker.go#L146

This PR moves the dry-run check to inside the scaleTarget method so every scaling action goes through the check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+BUG FIXES:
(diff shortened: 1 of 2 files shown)


lgfa29 (approved):
Thanks @jorgemarey!

I have a vague memory of maybe doing this on purpose to always enforce the min/max values, but thinking about it again it doesn't make sense to bypass the dry-run mode.

PR #848, opened 2024-02-06
accept max_scale_up and max_scale_down for target-value

brmzkw:
Hey,

The two optional parameters max_scale_up and max_scale_down determine the maximum and minimum number of scaling differences between two evaluations.

It is useful to avoid scaling down or scaling up a job too quickly.

The code is not tested yet. For now, this is a draft pull request to see if you might be interested to include the feature in the target-value plugin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+IMPROVEMENTS:
+ * plugin/strategy/target-value: Add new configuration `max_scale_up` and `max_scale_down` to allow restricting how much change is applied on each scaling event [[GH-848](https://github.com/hashicorp/nomad-autoscaler/pull/848)]
+
 BUG FIXES:
  * agent: Fixed a bug that caused a target in dry-run mode to scale when outside of its min/max range [[GH-845](https://github.com/hashicorp/nomad-autoscaler/pull/845)]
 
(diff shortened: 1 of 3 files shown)


brmzkw:
I've force-pushed the unit tests. Feel free to reach out if there are any queries!

lgfa29 (approved):
Thank you for the contribution @brmzkw! I really like the idea 🙂 

It provides a stepped scaling instead of potentially abrupt changes. I was thinking if we could adopt this configuration more broadly such that it could be used with any strategy:

```hcl
policy {
  # ...
  max_scale_up   = 10
  max_scale_down = 3
  # ...
}
```

But looking at the other strategies we have so far they seem more controlled than `target-value` because they usually define static changes. We can always expand later if we find the need.

Would be able to write some docs for these new config in https://github.com/hashicorp/nomad/blob/main/website/content/tools/autoscaling/plugins/strategy/target-value.mdx?

brmzkw:
> But looking at the other strategies we have so far they seem more controlled than `target-value` because they usually define static changes. We can always expand later if we find the need.

This is why I implemented it in the target-value plugin. For example, I'm not sure it would make sense to have these parameters for the pass-through or for the fixed-value strategies.

Thanks for the merge, I'll update the documentation today.

Do you know when you intend to make a release?

PR #849, opened 2024-02-09
agent: reload Enterprise configuration

lgfa29:
When the Nomad agent configuration is reloaded the Enterprise license checker must also be updated in case the Nomad token used to make API calls changes.


GNUmakefile
lgfa29, line 1: These changes just make the `GNUmakefile` between OSS and ENT to be more alike to avoid merge conflicts.

jrasell (approved):
LGTM.

Issue #843, opened 2024-01-30
Datadog APM: Including `now` in time range causes incorrect scaling actions


jaltavilla:
We setup scaling policies using the datadog apm with both the aws-asg plugin for ec2 instances and the nomad-target for jobs. In both cases we used the target strategy to scale (policies below). When we did this, we discovered that scaling actions would occur seemingly randomly. The service or asg could be underutilized and still double or triple in scale then scale down to the expected amount after the policy cooldown expired.

After some investigation we discovered that the cause of this was that the datadog query was sometimes returning incorrect values for the most recent point(s), combined with the target strategy only using the last point in the series. This could also be seen on the dashboards; the point for the current minute would sometimes spike up or down for a few seconds before correcting itself. I spent some time working with datadog support to try and resolve this. Ultimately they said that including `now` in the query would always potentially lead to these results. Which makes sense to me; data is still coming in from the agents for the current time.

> I have reviewed this issue further with the Engineering team, and unfortunately the spike occurring seems to be a product of the live  timeframe query. This means that as long as the timeframe is querying a constant time, like past X minutes, the spike will keep  happening. 
 >
>The suggested workaround is to use a timeframe that doesn't have the live aspect and avoids the most recent minute. For example - if I'm querying at 01:00pm, to get complete data I'd query from 12:44pm-12:59pm (i.e. give 1min for the data to show up). It would take extra work on the front end to write a script that changes time intervals, but would ultimately end in similar data as a live timeframe. 

I searched the issues and hashicorp forums and was a bit surprised that no one else had experienced this. I modified policyeval/base_worker.go's generation of the range `to := time.Now().Add(-1 * time.Minute)` to test datadog's suggestion. So far this seems to have fixed it for us. (Although this would affect all apms)

It would be nice if the datadog apm either shifted the query window, or maybe better allowed it via config. The use case config would be nice for is to scale based on the prior weeks behavior. Our services tend to experience similar load on a weekly cadence. Datadog has a `week_before` function that can be used in a check. However, querying the week before at now isn't helpful because we are already experiencing the same load. But, if we used week_before with a timeshift of +15 minutes we could scale just ahead of time, 

Thanks!

aws-asg policy snippet
```
        check "cpu-allocated" {
            source          = "datadog"
[...]

PR #850, opened 2024-02-10
policy: add `query_window_offset` configuration

lgfa29:
The Nomad Autoscaler uses the current time as the end of the query window. This can cause instability, specially for strategies that only take the most recent data point into account, because the latest data may still be processing.

`query_window_offset` defines a time offset applied to the query window to push it back in time. A short `query_window_offset` value can help with the unstable readings, and a long value can be used for predictive scaling based on historical data (for example, from `7d` ago).

Closes https://github.com/hashicorp/nomad-autoscaler/issues/843.


lgfa29:
Tests fixed in https://github.com/hashicorp/nomad-autoscaler/pull/851

lgfa29:
> Tests fixed in #851

Rebased to pick-up test fixes

jrasell (approved):
LGTM!

Issue #711, opened 2023-09-19
Nomad Autoscaler Idle forever with context.canceled error


bernardoVale:
I couldn't find a way to reproduce this error, but sometimes, autoscaler gets stuck and returns `context canceled` to all policy evaluations:

```
2023-09-19T18:08:24.247Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=af2adcdd-ffc6-8c54-e71d-dbbf55543a77 error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:24.349Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=9418971a-8261-cbd7-7730-5f8748f455d0 error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:24.447Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=04db11b8-6242-c03d-cb93-010c35e35285 error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:24.551Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=3e2681aa-68d2-41ae-f4e0-cd1dcd0b8ba3 error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:24.848Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=3c10001b-0402-743b-5bfd-73577527d9ac error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:24.948Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=236fcef8-6a4f-be5d-ae61-fd56eabfe92f error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:25.148Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=fa573b85-9ab6-952c-8deb-611a59f64d32 error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:25.249Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=8bf225c1-5a57-7ce1-d13e-ce63f7de45de error="rpc error: code = Canceled desc = context canceled"
2023-09-19T18:08:25.345Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=8ea0ab80-bbaa-35b0-7a50-69bc38a719fb error="rpc error: code = Canceled desc = context canceled"
[...]

PR #931, opened 2024-07-02
Healthcheck plugins on dispense, re-instantiating broken ones

nick-kentik:
Disclaimer: I don't really think this is the correct solution. It's just a first pass at it, with the most important bit being the test that reliably reproduces the problem. Very happy to work on the solution in a different way.

At the moment, if an external plugin that nomad-autoscaler relies on dies for any reason (OOM, for instance, or someone coming along and running `kill` on it), the nomad-autoscaler process cannot recover. Instead, we see a steady stream of errors of the form `rpc error: code = Canceled desc = context canceled` for the remaining lifetime of the process.

Attempt to solve this by:

- Running the `PluginInfo` RPC against plugins when they are being dispensed
- Killing the plugin if that fails, removing it from the plugin instances map
- Re-running `dispensePlugins()` if a request is made for a plugin not in the instance map

This seems like the minimal approach in terms of moving code around, but I suspect it has a significant performance impact, and probably spams the logs heavily in a range of scenarios. Still, it's a starting point.

Closes https://github.com/hashicorp/nomad-autoscaler/issues/711

### DIFF
--- a/plugins/manager/manager.go
+++ b/plugins/manager/manager.go
@@ -187,25 +187,45 @@ func (pm *PluginManager) killPluginLocked(pID plugins.PluginID) {
 // Dispense returns a PluginInstance for use by safely obtaining the
 // PluginInstance from storage if we have it.
 func (pm *PluginManager) Dispense(name, pluginType string) (PluginInstance, error) {
-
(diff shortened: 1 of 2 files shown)


nick-kentik:
Ping?

PR #945, opened 2024-08-07
Change nomad.Config to autoscaler.Nomad.Config when providing configuration for plugins.

### LINKED ISSUE #944: Nomad unix socket doesn't work.
When setting the address to the nomad unix api socket, the autoscaler stops working correctly.

The following errors are shown constantly

```
2024-08-07T10:15:32.127Z [WARN]  policy_manager.policy_handler: failed to get target status: policy_id=798ce9d8-6074-3295-692b-281d2fa52fd6 error="Get \"http://127.0.0.1/v1/job/test-job/scale?namespace=test&region=mx&wait=300000ms\": dial tcp 127.0.0.1:80: connect: connection refused"
2024-08-07T10:15:32.127Z [ERROR] policy_manager.policy_handler: Get "http://127.0.0.1/v1/job/test-job/scale?namespace=test&region=mx&wait=300000ms": dial tcp 127.0.0.1:80: connect: connection refused: policy_id=798ce9d8-6074-3295-692b-281d2fa52fd6
```

It seems that the problem is that when setting the unix socket in the adress nomad mutates that Address and that is later used on the nomad target plugin instead of the unix address.

jorgemarey:
Fixes #944

The config.Address is mutated when using unix sockets so when that address is provided for the nomad target plugin, the client created there doesn't use the unix address. This fixes that by using the autoscaler nomad config instead of the nomad api config.


tgross (commented):
Hi @jorgemarey! This passes tests but I'm concerned about dropping the use of `a.nomadCfg` here specifically because the documentation for that field says:

> // nomadCfg is the merged Nomad API configuration that should be used when
> // setting up all clients. It is the result of the Nomad api.DefaultConfig
> // merged with the user-specified Nomad config.Nomad.

That implies we could be missing default field values if we're using `a.config.Nomad` directly. Where is it that you're seeing the mutation that's breaking things?

jorgemarey:
Hi @tgross 

[Here](https://github.com/hashicorp/nomad/blob/2e5d6192a7d56abaa662de7e8eabaf8daf868885/api/api.go#L519-L525)  nomad api mutates de the Address field when it's an unix:// 

So first when the autoscaler calls api.NewClient [here](https://github.com/hashicorp/nomad-autoscaler/blob/d15da50bfe36f83de984b9b91235837424294df8/agent/agent.go#L187) that Address changes, so later when using that configuration to generate a new client for the [target plugin](https://github.com/hashicorp/nomad-autoscaler/blob/d15da50bfe36f83de984b9b91235837424294df8/plugins/builtin/target/nomad/plugin/plugin.go#L102), that client has an http://127.0.0.1 as the address for the client.

I changed the api.Config to a nomad autoscaler config because I thought made sense that both clients we're creating from the same configuration (both from the nomad autoscaler configuration).

tgross:
> Here nomad api mutates de the Address field when it's an unix://

Oh, I see... this looks like it's actually a bug in `api.NewClient`! The reason that mutation breaks reuse is that the `Config`'s private `url` field is updated here: https://github.com/hashicorp/nomad/blob/v1.8.2/api/api.go#L513 from the address. It should only ever update that if the value is unset (as it will be if you haven't constructed a client from the config before). Let me do a little bit of testing of that and I think we'll want to fix this in the `api` package and then rev that here. That'll also fix the problem for all consumers and not just the autoscaler.

tgross:
I've fixed that here: https://github.com/hashicorp/nomad/pull/23785

I'm going to close this PR, and once #23785 has been merged, we can get the autoscaler's version of the Nomad `api` package updated when Nomad 1.8.3 gets released (next week).

jorgemarey:
Great. Thanks @tgross!

PR #947, opened 2024-08-08
test: run CI on pull requests

tgross:
The GitHub Actions configuration doesn't run CI on pull requests, only on pushes which doesn't cover external PRs (Ex https://github.com/hashicorp/nomad-autoscaler/pull/945). Update the behavior to match what we do in Nomad core, where PRs get CI as well as merges to main.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,9 @@
 name: build
 
 on:
+  pull_request:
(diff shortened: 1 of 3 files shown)


dduzgun-security (approved):
LGTM! 👍

gulducat (approved):

Issue #955, opened 2024-08-20
Autoscaler can't connect to UNIX domain socket.


nwmqpa:
Nomad server version: 1.8.3
Nomad client version: 1.8.3
Nomad autoscaler version: 0.4.5 [(with custom plugin)](https://hub.docker.com/r/nwmqpa/nomad-autoscaler-plugins)

I've started experimenting with nomad workload identity and nomad autoscaler, and it would seems that the patch for  #944 isn't working for me.

<details>
  <summary>Relevant log</summary>
  
  ```
  2024-08-20T13:54:56.590Z [WARN]  internal_plugin.nomad-target: failed to read job scale status, retrying in 10 seconds: Job=service2 error="Get \"http://127.0.0.1/v1/job/service2/scale?namespace=XXXX&region=playground&wait=300000ms\": dial tcp 127.0.0.1:80: connect: connection refused"
  2024-08-20T13:54:56.084Z [WARN]  internal_plugin.nomad-target: failed to read job scale status, retrying in 10 seconds: Job=service1 error="Get \"http://127.0.0.1/v1/job/service1/scale?namespace=XXXX&region=playground&wait=300000ms\": dial tcp 127.0.0.1:80: connect: connection refused"
  ```

</details>

<details>
  <summary>Full job HCL</summary>

  ```
  job "nomad-autoscaler" {
    namespace = "infrastructure"
  
    priority = 98
  
    group "nomad-autoscaler" {
  
      network {
        mode = "bridge"
        port "http" {
          to = 8080
        }
      }
  
      service {
        name = "nomad-autoscaler"
        port = 8080
  
        connect {
          sidecar_service {
            proxy {
              transparent_proxy {}
            }
          }
  
          sidecar_task {
            resources {
              cpu    = 20
              memory = 50
            }
          }
        }
  
        check {
          expose   = true
          type     = "http"
          port     = "http"
          path     = "/v1/health"
          interval = "3s"
          timeout  = "1s"
        }
      }
  
      task "nomad-autoscaler" {
        driver = "docker"
  
        leader = true
  
        vault {
          role        = "infrastructure-nomad-autoscaler"
          change_mode = "restart"
        }
  
        identity {
          env = true
        }
  
        config {
          image   = "nwmqpa/nomad-autoscaler-plugins:v0.2.2"
          command = "nomad-autoscaler"
          ports   = ["http"]
          args = [
            "agent",
            "-config=${NOMAD_TASK_DIR}/config",
[...]

PR #966, opened 2024-09-05
config: use `URL` method to get Nomad address for plugins

tgross:
For #944 we fixed the Nomad API package so that it no longer mutated the private `url` field if previously set, which allowed reusing an `api.Config` object between clients when a unix domain socket was in use.

However, the autoscaler plugins for Nomad strategy and target don't use the `api.Config` object we parse directly and instead get a map of string->string derived from that config so it can be passed over the go-plugin interface. This mapping did not account for the `Address` field being mutated when unix domain sockets are in use, so the bug was not actually fixed.

Update the mapping to use the safe `URL()` method on the config, rather than reading the `Address` field.

Fixes: https://github.com/hashicorp/nomad-autoscaler/issues/955
Ref: https://github.com/hashicorp/nomad/pull/23785

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## UNRELEASED
 
+BUG FIXES:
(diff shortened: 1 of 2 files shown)


tgross:
Needs a test fix.

Juanadelacuesta (approved):
LGTM

PR #1001, opened 2024-11-20
setting default codeowners file

mmcquillan:
Setting a default for codeowners file for PLAT-3161

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,6 +1,2 @@
 # Blanket ownership of all PRs.
-* @jrasell @Juanadelacuesta
-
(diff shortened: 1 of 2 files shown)


mdeggies:
👍  thank you!

jrasell (commented):
@mmcquillan we already have a codeowners file [here](https://github.com/hashicorp/nomad-autoscaler/blob/main/.github/CODEOWNERS). Is it possible to update this one instead of adding a second file, or does it need to be within the root dir?

jrasell (approved):

pkazmierczak (approved):

PR #1060, opened 2025-03-24
build: update toolchain components to 1.24

tgross:
We updated to Go 1.24 in #1045 but it looks like this missed a few build components that are now causing issues with dependabot PRs.


jrasell (approved):
LGTM, thanks @tgross!

PR #1066, opened 2025-04-08
Upgrade armon/go-metrics to hashicorp/go-metrics

mkeeler:
## Description

This PR updates the metric emission to use hashicorp/go-metrics instead of armon/go-metrics. This is generally a non-functional change. If any clients were importing the `agent`, `plugins/manager`, `policy` or `policyeval` packages then this could potentially break their own metrics export. From looking at the codes purpose and talking with a few others, I don't believe this should be the case.

## Testing

TBD - still figuring this out.


chickenandpork:
Hey @mkeeler I think I've solved this CI error you've run into before; Mind if I take a look, maybe slide a commit onto your PR ?

tgross:
@mkeeler if you rebase on main you should be good-to-go.

tgross (approved):

tehut (approved):

PR #1070, opened 2025-04-16
Do not return negative newCount

lukaszo:
When the number of instances is 0 and the threshold plugin is used with `delta=  -1` it could return a `newCount = -1`.
This would cause creating a new scaling event. 

To avoid such behavior, I'm adding a check and set `newCount` to 0.


pkazmierczak:
@lukaszo That makes sense to me, thanks a lot for the contribution! Mind adding a line to the changelog, under `UNRELEASED/BUG FIXES`? Something like ` * plugin/strategy/threshold: do not return negative newCount`? dzięki 🙇‍♂️

pkazmierczak (approved):
LGTM! Thanks again for the contribution, @lukaszo!

PR #1082, opened 2025-05-14
build: update alpine to address busybox vulnerability

pkazmierczak:
Alpine 3.21 ships with busybox 1.37 which addresses CVE-2024-58251 and CVE-2025-46394.


pkazmierczak:
ugh, 1.37.0-r12 which is the latest version of busybox in Alpine is still affected by the CVE. I'll keep an eye and revisit once the package is patched.

PR #1088, opened 2025-05-23
Update policy and target handlers to stop monitoring policies on stopped jobs

Juanadelacuesta:
Whenever a workload is stopped, the autoscaler is not notified and keeps monitoring the job until it is garbage collected. This PR updates the handlers to stop monitoring stopped jobs.
The rate limiter is used to track metrics and will be removed on a following PR once all the updates on the nomad connections is done.


policy/handler.go
```diff
@@ -159,6 +159,11 @@ func (h *Handler) Run(ctx context.Context, evalCh chan<- *sdk.ScalingEvaluation)
 			continue
 
 		case p := <-h.ch:
+			if !p.Enabled {
+				monitorCtx.Done()
+				break
+			}
```
gulducat, line 165: what's this little guy doin here?  is this PR somehow causing `p.Enabled = false`, or some other reason?

also a `Context.Done()` call doesn't cause anything to happen, it just returns a channel that gets closed when the context is done.

I would say you want `cancel()` here, but with this `break`, the `defer cancel()` from line 115 kicks in.

policy/nomad/source.go
```diff
@@ -102,41 +104,26 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
 		// of scaling policies. The call is done in a goroutine so we can
 		// still listen for the context closing or a reload request.
-		blockingQueryCompleteCh := make(chan struct{})
-		go func() {
-			// Obtain a handler now so we can release the lock before starting
-			// the blocking query.
-			s.nomadLock.RLock()
-			scaling := s.nomad.Scaling()
-			s.nomadLock.RUnlock()
+		//blockingQueryCompleteCh := make(chan struct{})
```
gulducat, line 108: this comment no longer applies without the goroutine, but why remove the goroutine?

since the API call is a blocking query (right?), which might hold open for a while, we would want to be able to drop it on quit/reload...

policy/nomad/source.go
```diff
@@ -102,41 +104,26 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
 
-		select {
-		case <-ctx.Done():
-			s.log.Trace("stopping ID subscription")
-			return
-		case <-s.reloadCh:
-			s.log.Trace("reloading policies")
-			continue
-		case <-blockingQueryCompleteCh:
-		}
```
gulducat, line 125: if we want to clean this up just a little bit, this channel could instead be a `make(chan error)`, and send the `ListPolicies` `err` down it. that way the next `err != nil` check wouldn't be referring to something set directly in the goroutine.  something like

```go
case err = <-listPoliciesError:
```

policy/nomad/source.go
```diff
@@ -102,41 +104,26 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
-			case <-s.reloadCh:
-				s.log.Trace("reloading policies")
-				continue
-			case <-time.After(10 * time.Second):
-				continue
-			}
+			policy.HandleSourceError(s.Name(),
+				fmt.Errorf("failed to call the Nomad list policies API: %v", err), req.ErrCh)
+			time.Sleep(10 * time.Second)
+
```
gulducat, line 125: we also don't want to sleep a full 10 seconds, or bother error-handling something that may be stale from a reload, or irrelevant from a shutdown, eh?

policy/nomad/source.go
```diff
@@ -164,6 +151,16 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
...
+		select {
+		case <-ctx.Done():
+			s.log.Trace("stopping ID subscription")
+			return
+		case <-s.reloadCh:
+			s.log.Trace("reloading policies")
+			continue
+		case <-ticker.C:
+		}
```
gulducat, line 163: hm, is this new ticker here to account for cases where the blocking query unblocks very rapidly?  wouldn't this add an extra half-second to all calls, even those that properly block waiting for changes?

policy/nomad/source.go
```diff
@@ -213,12 +210,6 @@ func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq)
 		case <-blockingQueryCompleteCh:
 		}
 
-		// Return immediately if context is closed.
-		if ctx.Err() != nil {
-			log.Trace("done with policy monitoring")
-			return
-		}
```
gulducat, line 220: is this no longer needed?

gulducat (commented):
I had some difficulty following these changes, I think because there's a lot of goroutine/channel action in this repo, only tangential to your changes!

similarly, my commentary is maybe a little scattered. let me know if you'd like to go over it synchronously (heh) on a video call or something!

plugins/builtin/target/nomad/plugin/plugin.go
```diff
@@ -186,18 +188,30 @@ func (t *TargetPlugin) Status(config map[string]string) (*sdk.TargetStatus, erro
...
 
 		t.statusHandlers[nsID] = jsh
+
+		go func() {
+			// This is a blocking function that will only return if there is
+			// an error or the job has stopped.
+			jsh.start()
+
+			t.statusHandlersLock.Lock()
+			delete(t.statusHandlers, nsID)
```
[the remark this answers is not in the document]
Juanadelacuesta: Yes sir!

plugins/builtin/target/nomad/plugin/state_test.go
```diff
@@ -179,12 +176,11 @@ func Test_jobStateHandler_stop(t *testing.T) {
 	assert.Equal(t, int64(0), jsh.lastUpdated)
 
 	// Set some data that will be overwritten by stop().
-	jsh.isRunning = true
+	jsh.jobRunning = true
 	jsh.scaleStatus = &api.JobScaleStatusResponse{JobID: "test"}
 
 	// Call stop and make assertions.
 	jsh.setStopState()
-	assert.False(t, jsh.isRunning)
```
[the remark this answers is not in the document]
Juanadelacuesta: I didn't add anything because Im not changing the way it works, but you are right, that behaviour was not there before, so I need a new test

rate_limiter/rate_limiter.go
```diff
@@ -0,0 +1,78 @@
...
+type CustomRoundTRipper struct {
+	rateLimiter *rate.Limiter
+	source      string
+	rt          http.RoundTripper
+}
+
+func (irt *CustomRoundTRipper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if irt.rateLimiter != nil {
+		ctx := context.Background()
+		err := irt.rateLimiter.Wait(ctx)
```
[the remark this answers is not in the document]
Juanadelacuesta: Yeah, it makes sense to return if the request is canceled, thanks!

plugins/builtin/target/nomad/plugin/state.go
```diff
@@ -158,14 +162,11 @@ func (jsh *jobScaleStatusHandler) status(group string) (*sdk.TargetStatus, error
 // start runs the blocking query loop that processes changes from the API and
 // reflects the status internally.
 func (jsh *jobScaleStatusHandler) start() {
+	defer jsh.setStopState()
```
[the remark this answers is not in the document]
Juanadelacuesta: You are right, and that would only happen when a job was garbage collected.

policy/nomad/source.go
Juanadelacuesta: This part of the code will have a lot of changes, so I didn't want to do much refactoring just yet.

policy/handler.go
Juanadelacuesta: "what's this little guy doin here? is this PR somehow causing p.Enabled = false, or some other reason?" - On line 119, a second loop is triggered that monitors any changes to the policy definition, which are fed into this loop using the `p` channel. If the policy is disabled, we need to return, that is what that check is doing. This connects to the Nomad PR where a policy is automatically disabled when a job is stopped.

 "also a Context.Done() call doesn't cause anything to happen, it just returns a channel that gets closed when the context is done." If returns from the handler forcing the `defer cancel()` which causes the policy monitor to return as well.

Anything that causes this function to return need to also stop the policy monitor, hence the `defer cancel()` instead of multiple calls to cancel.

policy/nomad/source.go
Juanadelacuesta: Before this was a weird logic that the first time the query returned, the `blockingQueryCompleteCh` was closed or a time out and error happend, the intention behind it  seemed to be to ensure queries were possible and the connection was working, so I split the logic, moving the initial call to initialisation, making the  `blockingQueryCompleteCh` unnecessary. The test I removed was checking that logic.

plugins/builtin/target/nomad/plugin/state.go
```diff
@@ -68,34 +63,43 @@ type jobScaleStatusHandler struct {
...
+			Namespace: jsh.namespace,
+		}
+
+		status, _, err := jsh.client.Jobs().ScaleStatus(jsh.jobID, q)
+
+		resultCh <- struct {
+			status *api.JobScaleStatusResponse
+			err    error
+		}{status, err}
+	}()
```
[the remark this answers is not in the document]
Juanadelacuesta: It was done like that, which I didn't really like, having the extra channel an all that just to check that calls were possible, I just rearranged and separated initialising calls from main loop calls.

policy/nomad/source.go
Juanadelacuesta: It didnt make a lot of sense to begin with

gulducat:
Hey sorry I was rude and didn't get back to you! <3

PR #1094, opened 2025-05-26
Switch from blocking query to periodical polling for policy monitoring

Juanadelacuesta:
The policy monitor is constantly running a blocking query on the `/v1/scaling/policies`, this added to one blocking query per active policy can end up starving the nomad servers of available connections. This PR changes the mechanism to a polling query every 500ms, allowing the connection to be reused in the meantime.

### DIFF
--- a/policy/nomad/source.go
+++ b/policy/nomad/source.go
@@ -97,32 +97,36 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
 	ticker := time.NewTicker(policiesIDsPollingInterval)
 	defer ticker.Stop()
 
+	var (
+		policies []*api.ScalingPolicyListStub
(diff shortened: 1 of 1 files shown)


Juanadelacuesta:
> I'm not sure I understand the reasoning behind fix. The problem seems to be that we're making a blocking query per active policy, not that we're making a single blocking query for the List Scaling Policies API. Shouldn't we be making a change to ensure we're not making the blocking query per active policy instead? i.e. we could keep this efficient blocking query to List Scaling Policies and then make Read Scaling Policy API calls only as needed.

The idea is to move away from blocking queries altogether, currently there is one blocking query for all policies and one per policy in the agent, plus one per job in the target. Im switching them all for polling, but one by one, there is another PR in draft and Im working on the third.

Juanadelacuesta:
Taking a different approach

PR #1101, opened 2025-06-06
Update the source monitor to account for policy changes 

Juanadelacuesta:
Currently the autoscaler does policy monitoring in two steps:

One function is constantly looking for the overall number of policies
One function per policy is constantly looking for changes on the policy.
This approach is very sensitive to race conditions because we have basically two functions monitoring the same thing.
This PR is the first of a series that will change the approach into having one function monitoring all the policies and reporting to the individual handler if a particular policy was changed and leaving it to the handler to fetch the changes.

To do so we need to add one more function to the source interface to make a single call for updates and change the monitoring IDs function to report changes on each particular policy.

### DIFF
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -276,7 +276,7 @@ func (s *Source) handleDir() (map[policy.PolicyID]bool, error) {
 		for name, scalingPolicy := range policies {
 			// Get the policyID for the file.
 			policyID := s.getFilePolicyID(file, name)
-			scalingPolicy.ID = string(policyID)
+			scalingPolicy.ID = policyID
 
 			if !scalingPolicy.Enabled {
 				s.log.Trace("policy is disabled",
@@ -341,7 +341,7 @@ func (s *Source) getFilePolicyID(file, name string) policy.PolicyID {
 	// and store this.
 	policyID, ok := s.idMap[md5Sum]
 	if !ok {
-		policyID = policy.PolicyID(uuid.Generate())
+		policyID = uuid.Generate()
 		s.idMap[md5Sum] = policyID
 	}
 
--- a/policy/file/source.go
+++ b/policy/file/source.go
@@ -108,7 +108,9 @@ func (s *Source) ReloadIDsMonitor() {
 // Any error doing so will be sent to req.ErrCh.
 // If MonitorPolicy receives on s.ReloadCh, it will re-check the file on disk,
 // and write again to req.ResultCh if the policy has changed.
-// Note: MonitorPolicy should only return when ctx is done.
+// Note: MonitorPolicy should only return when ctx is done and will not be
+// monitoring changes in policies, a call to Reload is necessary for changes
+// in policies to become effective.
 func (s *Source) MonitorPolicy(ctx context.Context, req policy.MonitorPolicyReq) {
 
 	// Close channels when done with the monitoring loop.
@@ -245,9 +247,8 @@ func (s *Source) identifyPolicyIDs(resultCh chan<- policy.IDMessage, errCh chan<
 }
 
 // handleDir iterates through the configured directory, attempting to decode
-// and store all HCL and JSON files as scaling policies. If the policy is not
-// enabled it will be ignored.
-func (s *Source) handleDir() ([]policy.PolicyID, error) {
+// and store all HCL and JSON files as scaling policies.
+func (s *Source) handleDir() (map[policy.PolicyID]policy.PolicyUpdate, error) {
 
 	// Obtain a list of all files in the directory which have the suffixes we
 	// can handle as scaling policies.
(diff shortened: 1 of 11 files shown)


policy/source.go
```diff
@@ -15,6 +15,10 @@ import (
 // a policy check.
 const DefaultQueryWindow = time.Minute
 
+// PolicyID contains identifying information about a policy, as returned by
+// policy.Source.MonitorIDs()
+type PolicyID = string
```
tgross, line 20: Non-blocking: seems a shame to reduce type safety by switching to a type alias here instead of the original typedef. We've got a bunch of string IDs flying around this code and having the typedef prevents bugs pretty much for free. Is this necessary for the refactoring?

(That being said, we have string IDs all over the Nomad core codebase where we should probably have structured IDs and it doesn't seem to be a problem in practice, and that's a much larger codebase. So it probably doesn't matter?)

policy/file/source.go
```diff
@@ -348,3 +351,23 @@ func (s *Source) getFilePolicyID(file, name string) policy.PolicyID {
...
 	return md5.Sum([]byte(fmt.Sprintf("%v", i)))
 }
+
+func (s *Source) GetLatestVersion(_ context.Context, policyID policy.PolicyID) (*sdk.ScalingPolicy, error) {
+	s.policyMapLock.Lock()
+	defer s.policyMapLock.Unlock()
+
+	val, ok := s.policyMap[policyID]
+	if !ok {
+		return nil, fmt.Errorf("failed to get policy %s", policyID)
```
tgross, line 361: Nitpick on this error message: this seems like a case where the policy is no longer being watched, right?

policy/file/source_test.go
```diff
@@ -81,7 +81,12 @@ func testFileSource(t *testing.T, dir string) (*Source, []policy.PolicyID) {
 
 func TestSource_MonitorPolicy(t *testing.T) {
 	s, ids := testFileSource(t, "./test-fixtures")
-	pid := ids[0]
+	var pid policy.PolicyID
+
+	for id := range ids {
+		pid = id
+		break
+	}
```
tgross: Given that `testFileSource` is only ever used by these two tests, maybe it should return the `PolicyID` instead of a map that the callers have to iterate?

policy/nomad/source.go
```diff
@@ -90,29 +135,44 @@ func (s *Source) ReloadIDsMonitor() {
...
-			scaling := s.nomad.Scaling()
-			s.nomadLock.RUnlock()
 
-			policies, meta, err = scaling.ListPolicies(q)
-			close(blockingQueryCompleteCh)
+			q.WaitIndex = s.latestIndex
+			ps, meta, err := s.policiesGetter.ListPolicies(q)
+
+			// There is no access to the call context, but to avoid writing to
+			// a closed channel first check if the context was cancelled.
```
tgross: This is racy: the context could close after we check but before we send, and that would panic. It's not even safe to do this:

```go
			select {
			case <-ctx.Done():
				return
			case blockingQueryCompleteCh <- results:
				policies: ps,
				meta:     meta,
				err:      err,
			}
```

because the ordering of case evaluation is semi-random. I'm pretty sure this is why the original design uses a mutex. But I think we can solve this by creating a new channel on each loop iteration and then have _this_ goroutine responsible for closing it. (We usually call these "futures" in the Nomad core codebase.)

tgross (changes requested):
This looks great @Juanadelacuesta! The only blocking thing here is the race condition / panic in the `MonitorIDs` loop.

policy/nomad/source.go
Juanadelacuesta: :0

policy/file/source.go
Juanadelacuesta: Not exactly, according to the code on the `MonitoringIDs`, the file source only loads the policies on start and on relaod,  so if the policy is not present on that map, it was not found to begin with. We could make it change aware later, but I for the time being, I want to keep all the same behaviours. The one thing I should change is that we already have the policy loaded. No need to read it again, since changes are only picked up on reload.

policy/file/source_test.go
Juanadelacuesta: Good point!

policy/nomad/source.go
Juanadelacuesta: Placed the create and close channel mechanism back in place

policy/source.go
Juanadelacuesta: I didnt really see the advantage because it just lead to a bunch of typecasting in different places done in different ways, and  as we are not checking in any way that the PolicyID is an ID and not a random string, I saw no difference, should I put it back?

tgross: Makes sense... what you've got here is fine. I think we'd only really see an advantage if the API serializer returned the typedef type and not a string.

tgross (approved):
LGTM!

PR #1102, opened 2025-06-06
Modify policy handler mechanism

Juanadelacuesta:
This PR is a continuation of [PR-1100](https://github.com/hashicorp/nomad-autoscaler/pull/1100)

In that one, the source interface and all its implementations where updated to include a new non blocking query to get the latest version of a policy.

In this PR, we change the mechanism for monitoring and executing policies and  move from using 2 blocking queries, one in the monitor for the overall number of policies and on in the policy handler to monitor changes on each specific policy, to having  the manager monitor the total number of policies with a blocking query, and when it signals any changes,  use the newly added function to get the latest version of any updated policy and feed id to the corresponding handler, that way the policy monitor is in charge of reading and updating any policy changes, while the policy handler just executes it.

### DIFF
--- a/policy/manager.go
+++ b/policy/manager.go
@@ -69,8 +69,8 @@ func NewManager(log hclog.Logger, ps map[SourceName]Source, pm *manager.PluginMa
 }
 
 func (m *Manager) getHandlersNum() int {
-	m.handlersLock.Lock()
-	defer m.handlersLock.Unlock()
+	m.handlersLock.RLock()
+	defer m.handlersLock.RUnlock()
 
 	return len(m.handlers)
 }
@@ -178,19 +178,18 @@ func (m *Manager) monitorPolicies(ctx context.Context, evalCh chan<- *sdk.Scalin
 					}
 				}
 
-				m.handlersLock.Lock()
+				m.handlersLock.RLock()
 				pht := m.handlers[policyID]
+				m.handlersLock.RUnlock()
 				if pht != nil {
 					// If the handler already exists, send the updated policy to it.
 					m.log.Trace("sending updated policy to existing handler",
 						"policy_id", policyID, "policy_source", message.Source)
 
 					pht.updates <- updatedPolicy
-					m.handlersLock.Unlock()
+
 					continue
 				}
-
-				m.handlersLock.Unlock()
 
 				// If we reach this point it means it is a new policy and we need
 				// a new handler for it.
@@ -285,8 +284,6 @@ func (m *Manager) EnforceCooldown(id string, t time.Duration) {
 
 // ReloadSources triggers a reload of all the policy sources.
 func (m *Manager) ReloadSources() {
-	m.handlersLock.Lock()
-	defer m.handlersLock.Unlock()
 
 	// Tell the ID monitors to reload.
 	for _, policySource := range m.policySources {


policy/manager.go
```diff
@@ -5,30 +5,39 @@ package policy
...
 type Manager struct {
 	log           hclog.Logger
-	policySource  map[SourceName]Source
-	pluginManager *manager.PluginManager
+	policySources map[SourceName]Source
+	targetGetter  targetMonitorGetter
 
 	// lock is used to synchronize parallel access to the maps below.
-	lock sync.RWMutex
+	hanldersLock sync.RWMutex
```
tgross:
```suggestion
	handlersLock sync.RWMutex
```

policy/manager.go
```diff
@@ -48,16 +58,23 @@ func NewManager(log hclog.Logger, ps map[SourceName]Source, pm *manager.PluginMa
 
 	return &Manager{
 		log:             log.ResetNamed("policy_manager"),
-		policySource:    ps,
-		pluginManager:   pm,
-		handlers:        make(map[PolicyID]*Handler),
-		keep:            make(map[PolicyID]bool),
+		policySources:   ps,
+		targetGetter:    pm,
+		hanldersLock:    sync.RWMutex{},
```
tgross: It doesn't matter, but you don't actually need to initialize this; the default value for a mutex is initialized.

policy/handler.go
```diff
@@ -5,173 +5,97 @@ package policy
...
-		case p := <-h.ch:
-			h.applyMutators(&p)
-			h.updateHandler(currentPolicy, &p)
-			currentPolicy = &p
+			h.applyMutators(updatedPolicy)
+			h.updateHandler(updatedPolicy)
 
 		case <-h.ticker.C:
-			eval, err := h.handleTick(ctx, currentPolicy)
+			h.log.Trace("handling new tick")
```
tgross: Nitpick: you've got this log line in the `handleTick` function as well.

policy/handler.go
```diff
@@ -5,173 +5,97 @@ package policy
...
-	monitorCtx, cancel := context.WithCancel(ctx)
-	defer cancel()
-
-	// Start monitoring the policy for changes.
-	req := MonitorPolicyReq{ID: h.policyID, ErrCh: h.errCh, ReloadCh: h.reloadCh, ResultCh: h.ch}
-	go h.policySource.MonitorPolicy(monitorCtx, req)
+	h.ticker = time.NewTicker(h.policy.EvaluationInterval)
+	defer h.ticker.Stop()
 
+	h.log.Error("starting policy handler")
```
tgross: Probably shouldn't be an error :grin: 

```suggestion
	h.log.Trace("starting policy handler")
```

policy/manager.go
```diff
@@ -121,88 +140,127 @@ func (m *Manager) monitorPolicies(ctx context.Context, evalCh chan<- *sdk.Scalin
...
+				s := m.policySources[message.Source]
+				updatedPolicy, err = s.GetLatestVersion(ctx, policyID)
+				if err != nil {
+					m.log.Error("encountered an error getting the latest version for policy", "policyID", policyID, "error", err)
+					if isUnrecoverableError(err) {
+						return err
+					}
+				}
+
+				if pht := m.handlers[policyID]; pht != nil {
```
tgross: I think we need to hold a read lock on the `m.handlers` here in case of a concurrent `ReloadSources` call.

tgross (commented):
This looks great @Juanadelacuesta! I've left a few comments but only the one about the read lock is really blocking.

policy/manager.go
Juanadelacuesta: I was thinking, do we even need the lock? Only the monitor will be modifying the handlers.

tgross: The `EnforceCooldown` and `ReloadSources` methods (which I think get called from outside the monitor loop?) reads from the handlers. Mutating the map isn't atomic, so if you read it without the lock you can get weird stuff like a valid key but no value (which the runtime avoids by panicking with the message "fatal error: concurrent map read and map write").

policy/manager.go
```diff
@@ -121,88 +140,133 @@ func (m *Manager) monitorPolicies(ctx context.Context, evalCh chan<- *sdk.Scalin
...
+				if err != nil {
+					m.log.Error("encountered an error getting the latest version for policy", "policyID", policyID, "error", err)
+					if isUnrecoverableError(err) {
+						return err
+					}
+				}
 
-				// Mark policy as must-keep so it doesn't get removed.
-				m.keep[policyID] = true
+				m.handlersLock.Lock()
```
tgross: I think this can be a `Rlock()`?

policy/manager.go
```diff
@@ -121,88 +140,133 @@ func (m *Manager) monitorPolicies(ctx context.Context, evalCh chan<- *sdk.Scalin
...
+					// If the handler already exists, send the updated policy to it.
+					m.log.Trace("sending updated policy to existing handler",
+						"policy_id", policyID, "policy_source", message.Source)
 
-				// Check if we already have a handler for this policy.
-				if _, ok := m.handlers[policyID]; ok {
-					m.log.Trace("handler already exists",
-						"policy_id", policyID, "policy_source", policyIDs.Source)
+					pht.updates <- updatedPolicy
+					m.handlersLock.Unlock()
```
tgross: Do we need to wait until we've sent the update to unlock, or can we unlock immediately after the read? If we can unlock right after `pht := m.handlers[policyID]` there'd be less contention.

policy/manager.go
```diff
@@ -221,18 +285,13 @@ func (m *Manager) EnforceCooldown(id string, t time.Duration) {
 
 // ReloadSources triggers a reload of all the policy sources.
 func (m *Manager) ReloadSources() {
-	m.lock.Lock()
-	defer m.lock.Unlock()
+	m.handlersLock.Lock()
+	defer m.handlersLock.Unlock()
```
tgross: Now that we're not sending the reload to the handlers in this method, do we still need this lock? 

It looks like `policySources` never gets mutated once we create the `Manager` so we can use it without the lock. But if I've missed it somewhere then we'd probably need the lock on all the other `policySources` reads too.

tgross (dismissed):
LGTM. The remaining comments I have are performance improvements rather than correctness improvements. Would be nice to get those in.

policy/manager.go
Juanadelacuesta: I think this lock was a left over of the previous implementation, lets remove it

pkazmierczak (approved):
:shipit:

Issue #1096, opened 2025-05-30
The azure-vmss target plugin fails when used with Azure scale sets in flexible orchestration mode


presse1:
Nomad version: Nomad v1.10.1+ent
Nomad autoscaler version: Nomad Autoscaler v0.4.6+ent

The [azure-vmss target plugin](https://github.com/hashicorp/nomad-autoscaler/tree/main/plugins/builtin/target/azure-vmss) fails when used with an Azure scale set in flexible orchestration mode with the following error:
`Operation 'VirtualMachineScaleSets.instanceView.GET' is not allowed on Virtual Machine Scale Set`.

Per [Azure documentation](https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-orchestration-modes#autoscaling-and-instance-orchestration) flexible scale sets do not support `VMSS Get Instance View`.

Can the azure-vmss plugin be updated to support flexible scale sets?

tgross:
Will ship in the upcoming release.

PR #1106, opened 2025-06-09
Updating Azure Plugin to Support Flexible VMSS

benjamin-lykins:
Addresses #1096 

- Reviewed contributor guide. 
I will note, this might not be a  _small_ first PR to put in. Ended up being a little bit more noisy since I did migrate off the legacy Azure SDK.

### Changes 

1. Updated from deprecated to current Azure Compute SDK. 

"github.com/Azure/azure-sdk-for-go/services/compute/mgmt/2020-06-01/compute"
"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute"

2. Updated status checks

Supports both flexible and uniform scale sets. Queries the VMSS to get the `type` and then conditionally runs the respective process status/scale in and out processes. 

- Flexible scale sets are a shift from uniform and were unable to query the instance view of the instance. Essentially, grab the list of instances from the VMSS, then use that against the VM (not VMSS) API to get instance status. 

- When getting instance states, since Flexible VMSS can have up to 2000 instances, I used a goroutine to iterate through them more effectively. Hardcoded at 5 concurrent, might be worth allowing the end user to provide a limit on this, but from my testing with 5 concurrent against 100 instances, it took ~1 second. When I was sequentially pulling instance state, it took ~30 seconds for the same count. I did test limiting to 10 concurrent, but it took it from ~1 second to a little less than a second, so not much gained. 

3. The only existing function specifically  with Uniform I updated was `processInstanceView`. 

I can double check the why, but I believe one of the attributes was not available in the current api. 

4. Testing

Admittedly `Test_processInstanceView` used copilot to refactor for the new api. 

### Testing

Flexible VMSS

- [ ] Scale In
- [ ] Scale Out

Uniform VMSS

- [ ] Scale In
- [ ]  Scale Out


go.mod
```diff
@@ -34,10 +34,21 @@ require (
 	google.golang.org/protobuf v1.36.6
 )
 
+require (
+	github.com/AzureAD/microsoft-authentication-library-for-go v1.4.2 // indirect
+	github.com/golang-jwt/jwt/v5 v5.2.2 // indirect
+	github.com/kylelemons/godebug v1.1.0 // indirect
+	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
+)
```
tgross: https://github.com/hashicorp/nomad-autoscaler/actions/runs/15525245328/job/43703925781?pr=1106 is telling me that we need to run `go mod tidy` here. You should end up with a single `require` section for direct dependencies and a single `require` section for indirect dependencies.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
-	vmss.Authorizer = authorizer
+	// Create a new Azure client secret credential using the provided configuration.
+	cred, err := azidentity.NewClientSecretCredential(tenantID, clientID, secretKey, nil)
+	if err != nil {
+		return fmt.Errorf("failed to create Azure client secret credential: %w", err)
+	}
 
-	t.vmss = vmss
+	// Needed for flexible VMSS
+	// As of 4 June 2024, VMSS Get Instance View is not supported for flexible orchestration mode.
```
tgross: This is useful context for the PR but not a useful comment for future developers who will read it. We can remove this.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
 		return fmt.Errorf("failed to get the vmss update response: %v", err)
 	}
 
-	err = future.WaitForCompletionRef(ctx, t.vmss.Client)
+	resp, err := future.Poll(ctx)
 	if err != nil {
 		return fmt.Errorf("cannot get the vmss update future response: %v", err)
 	}
 
+	log.Debug("vmss update response", "response", resp.Body)
```
tgross: The response body is kinda huge and has a lot of deeply-nested structs which we'd need to audit continuously for sensitive data. Better to leave this log line out.

```suggestion
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
-	for pager.NotDone() {
-		for _, vm := range pager.Values() {
-			for _, s := range *vm.VirtualMachineScaleSetVMProperties.InstanceView.Statuses {
-				if strings.HasPrefix(*s.Code, "PowerState/") {
-					if *s.Code == "PowerState/running" {
-						log.Debug("found healthy instance", "id", *vm.ID, "instance_id", *vm.InstanceID)
-						remoteIDs = append(remoteIDs, fmt.Sprintf("%s_%s", vmScaleSet, *vm.InstanceID))
-					} else {
-						log.Debug("skipping instance", "id", *vm.ID, "instance_id", *vm.InstanceID, "code", *s.Code)
+		log.Debug("iterating over Azure ScaleSet instances", "vmss_name", vmScaleSet)
```
tgross: We already have a debug line for the `vmss_name` near the top; we can probably leave this one out too.

```suggestion
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
 	log := t.logger.With("action", "scale_in", "resource_group", resourceGroup, "vmss_name", vmScaleSet)
 
-	// Find instance IDs in the target VMSS and perform pre-scale tasks.
-	pager, err := t.vmssVMs.List(ctx, resourceGroup, vmScaleSet,
-		"startswith(instanceView/statuses/code, 'PowerState') eq true",
-		"instanceView/statuses", "instanceView")
-	if err != nil {
-		return fmt.Errorf("failed to query VMSS instances: %v", err)
-	}
+	log.Debug("listing Azure ScaleSet instances", "vmss_name", vmScaleSet)
```
tgross: The `logger.With` call above has already added this field to the logger:

```suggestion
	log.Debug("listing Azure ScaleSet instances")
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+		for pager.More() {
+			page, err := pager.NextPage(ctx)
+			if err != nil {
+				return fmt.Errorf("failed to list instances in VMSS: %v", err)
+			}
+			for _, vm := range page.Value {
+				if vm.Properties != nil && vm.Properties.InstanceView != nil && vm.Properties.InstanceView.Statuses != nil {
+					for _, s := range vm.Properties.InstanceView.Statuses {
+						if s.Code != nil && strings.HasPrefix(*s.Code, "PowerState/") {
+							if *s.Code == "PowerState/running" {
```
tgross: Given that we're comparing `s.Code` against  the full string `PowerState/running` anyways, I  think we can skip the `HasPrefix` conditional here.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
-	// Find instance IDs in the target VMSS and perform pre-scale tasks.
-	pager, err := t.vmssVMs.List(ctx, resourceGroup, vmScaleSet,
-		"startswith(instanceView/statuses/code, 'PowerState') eq true",
-		"instanceView/statuses", "instanceView")
-	if err != nil {
-		return fmt.Errorf("failed to query VMSS instances: %v", err)
-	}
+	log.Debug("listing Azure ScaleSet instances", "vmss_name", vmScaleSet)
+
+	if vmssMode == "Uniform" {
```
tgross: The scope of this conditional and the one for `"Flexible"` are quite large and have a _lot_ of shared code. We should extract the queries for the two modes out into their own methods and then reuse as much of that shared code as we can either here or in functions we call from those two methods.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +357,21 @@ func azureNodeIDMap(n *api.Node) (string, error) {
 
 	return "", fmt.Errorf("attribute %q not found", nodeAttrAzureInstanceID)
 }
+
+// Ended up making this a method to simplify.
+func (t *TargetPlugin) getVMSSFlexibleVMs(ctx context.Context, resourceGroup string, vmScaleSet string) (err error, vms []armcompute.VirtualMachineScaleSetVM) {
```
tgross: The method signature here is backwards from idiomatic Go; the error should always be returned last. We probably want to return a slice of pointers here -- the caller ends up needing them as pointers and these are large objects we don't want to copy around if we don't have to.

Also, this method is ultimately turning a paged response into a slice. You could probably make it generic function over the contents of the pager and the type of the response and then you could share it between both vmss modes:

`func vmssPagerToSlice[T](ctx context.Context, pager *runtime.Pager[VirtualMachineScaleSetVMsClientListResponse]) ([]T, error)`

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+
+				mu.Lock()
+				remoteIDs = append(remoteIDs, *vm.InstanceID)
+				mu.Unlock()
+
+			}(vm)
+		}
+		wg.Wait()
+
+		// Business as usually, run the pre-scale tasks to prepare the nodes for scale in.
```
tgross: We haven't actually checked that we have non-zero amount of `remoteIDs`... we could have gotten errors or they weren't in `provisioningState/succeeded`. (This looks like it might be an existing bug but we shouldn't perpetuate it! :grin: )

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+
+			} else {
+				return errors.New("failed to validate remoteid format")
+			}
+		}
+
+		log.Debug("found instance IDs", "ids", instanceIDs)
+
+		// Not sure if there is a better way to do this, but we need to convert
+		// the instanceIDs to a slice of pointers for the Azure SDK.
```
tgross: This is fine... in the core Nomad codebase we have some helpers for this kind of slice conversion that makes it a little nicer, but it doesn't look like we have 'em here. I'd remove the "not sure if" part of the comment though.

Also, can't we do this conversion in the loop above? We don't use `instanceIDs` for anything else, right?

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+		instanceIDPtrs := make([]*string, 0, len(instanceIDs))
+
+		for _, id := range instanceIDs {
+			idCopy := id // ensure a new variable for correct reference
+			instanceIDPtrs = append(instanceIDPtrs, &idCopy)
```
tgross: As of Go 1.22 you don't need to do this kind of thing anymore ([ref](https://go.dev/blog/loopvar-preview)). 

```suggestion
			instanceIDPtrs = append(instanceIDPtrs, &id)
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+
+		// Terminate the detached instances.
+		log.Debug("deleting Azure ScaleSet instances", "instances", instanceIDs)
+
+		future, err := t.vmss.BeginDeleteInstances(ctx, resourceGroup, vmScaleSet, armcompute.VirtualMachineScaleSetVMInstanceRequiredIDs{
+			InstanceIDs: instanceIDPtrs,
+		}, nil)
+
+		if err != nil {
+			return fmt.Errorf("failed to scale in Azure ScaleSet: %v", err)
```
tgross:
```suggestion
			return fmt.Errorf("failed to scale in Azure ScaleSet: %w", err)
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
+
+		// Mainly for debugging purposes, log the number of instances to scale in.
+		log.Debug("found instances to scale in", "count", len(ids))
+
+		// Grabbing the instance IDs from the Nomad Nodes to scale in.
+		// You'll notice that the logic here is different than Uniform mode. This is mainly due to Azure's consistency of being inconsistent.
+		// https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-names
+
+		var instanceIDs []string
+		for _, node := range ids {
```
tgross: Just a general note on naming... we've got several slices floating around like `instanceIDs`, `ids`, and `remoteIDs`. But it looks like some of these IDs are for VM instances and some of them are for VMSS (that's like an autoscaling group, right?). We should probably differentiate between them.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -30,147 +30,314 @@ func argsOrEnv(args map[string]string, key, env string) string {
...
 
-	// Run any post scale in tasks that are desired.
-	if err := t.clusterUtils.RunPostScaleInTasks(ctx, config, ids); err != nil {
-		return fmt.Errorf("failed to perform post-scale Nomad scale in tasks: %v", err)
+				select {
+				case <-ctx.Done():
+					return
+				default:
+					// Continue processing
+				}
```
tgross: This is a nice little design!

You can get a little crazy with this kind of thing by having the outer goroutine looping over a channel of remote IDs (and handling empty strings as an error value), which would let us use two channels instead of a mutex, a wait group, and a channel. But this is probably easier to read this way, so :+1:

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -49,10 +51,16 @@ var _ target.Target = (*TargetPlugin)(nil)
...
-	config  map[string]string
-	logger  hclog.Logger
-	vmss    compute.VirtualMachineScaleSetsClient
-	vmssVMs compute.VirtualMachineScaleSetVMsClient
+	config map[string]string
+	logger hclog.Logger
+
+	// Using deprecated package.
+	// vmss    compute.VirtualMachineScaleSetsClient
+	// vmssVMs compute.VirtualMachineScaleSetVMsClient
```
tgross: Leftover from writing:

```suggestion
```

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -70,6 +78,8 @@ func NewAzureVMSSPlugin(log hclog.Logger) *TargetPlugin {
 // SetConfig satisfies the SetConfig function on the base.Base interface.
 func (t *TargetPlugin) SetConfig(config map[string]string) error {
 
+	t.logger.Info("setting Azure VMSS target plugin config", "config", config)
```
tgross: I'd leave this out... config especially feels like it's going to be easy to accidentally log someone's keys in the future.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -168,24 +180,56 @@ func (t *TargetPlugin) Status(config map[string]string) (*sdk.TargetStatus, erro
...
+	// https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute@v1.0.0#OrchestrationMode
+	// Flexible is not compatible with GetInstanceView which is used in Uniform logic.
+	// Get all VMs in the VMSS, so we can process them individually later.
+	vms := &[]armcompute.VirtualMachineScaleSetVM{}
+	if *vmss.Properties.OrchestrationMode == "Flexible" {
+		t.logger.Debug("VMSS Orchestration Mode", "mode", *vmss.Properties.OrchestrationMode)
+
+		err, flexVMs := t.getVMSSFlexibleVMs(ctx, resourceGroup, vmScaleSet)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get VMSS flexible VMs: %v", err)
```
tgross: Try to do this everywhere:

```suggestion
			return nil, fmt.Errorf("failed to get VMSS flexible VMs: %w", err)
```

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -168,24 +180,56 @@ func (t *TargetPlugin) Status(config map[string]string) (*sdk.TargetStatus, erro
...
 		Ready: true,
-		Count: *vmss.Sku.Capacity,
+		Count: *vmss.SKU.Capacity,
 		Meta:  make(map[string]string),
 	}
 
-	processInstanceView(instanceView, &resp)
+	// If flexible, it takes the VMSS VMs and processes them individually
+	// outside of the scope of the VMSS.
+	if *vmss.Properties.OrchestrationMode == "Flexible" {
```
tgross: Can we define these modes as constants? (If they're not already in the Azure SDK) It'd be easy for someone to come along later and accidentally drop the capitalization.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
 
 // processInstanceView updates the status object based on the details within
 // the vmss instances.
-func processInstanceView(instanceView compute.VirtualMachineScaleSetInstanceView, status *sdk.TargetStatus) {
+func processInstanceView(instanceView *armcompute.VirtualMachineScaleSetsClientGetInstanceViewResponse, status *sdk.TargetStatus) {
+
+	// Old logic did not work with the new Azure SDK. This is a bit more simplified.
+	// Updated to only see if any Statuses is getting reported back.
```
tgross: This comment looks like it's referring to the PR change and not a comment we want to carry forward?

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+	if instanceView == nil || len(instanceView.VirtualMachineScaleSetInstanceView.Statuses) == 0 {
+		return
+	}
 
-	for _, instanceStatus := range *instanceView.VirtualMachine.StatusesSummary {
-		if *instanceStatus.Code != "ProvisioningState/succeeded" {
-			status.Ready = false
+	if instanceView.VirtualMachineScaleSetInstanceView.Statuses != nil {
```
tgross: This can't be nil here, right? We just checked that.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+		for _, instanceStatus := range instanceView.VirtualMachineScaleSetInstanceView.Statuses {
+			if instanceStatus.Code != nil && *instanceStatus.Code != "ProvisioningState/succeeded" {
+				status.Ready = false
+			}
```
tgross: This loop seems redundant with the loop below, isn't it?

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
 		}
 	}
 }
+
+// Unfortunately, this requires us to iterate over each VM in the VMSS.
+// This can be upwards of 2000 VMs in a single VMSS, which Uniform is limited to 200.
+// Added a timer for tracking how long it takes to process the instance views.
+// Ran with a concurrency limit of 10 --- avoided api rate limits.
+// 100 VMs took about ~1 second to process.
+// Previously when running each VM sequentially, it took about about 20-30 seconds.
```
tgross: It looks like this section of the docstring is describing steps you took to validate the design? That's fine but let's call that out specifically.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -168,24 +180,56 @@ func (t *TargetPlugin) Status(config map[string]string) (*sdk.TargetStatus, erro
...
 	}
 
-	instanceView, err := t.vmss.GetInstanceView(ctx, resourceGroup, vmScaleSet)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get Azure ScaleSet Instance View: %v", err)
+	// Currently only two orchestration modes are supported - Flexible and Uniform.
+	// https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute@v1.0.0#OrchestrationMode
+	// Flexible is not compatible with GetInstanceView which is used in Uniform logic.
+	// Get all VMs in the VMSS, so we can process them individually later.
+	vms := &[]armcompute.VirtualMachineScaleSetVM{}
```
tgross: A slice is already mutable, so we don't need for this to be a pointer (which we'd then have to nil check and dereference everywhere it's used).

Also, it looks like at this point we're bifurcating into the two modes. So why have two conditional blocks for each instead of having them together? (We might also want to make sure we're gracefully handling any _new_ orchestration mode that gets introduced in future versions of Azure's API.)

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+	defer cancel()
+
+	// Using a WaitGroup to wait for all goroutines to finish.
+	// Using a channel to limit the number of concurrent requests to the Azure API.
+	// Mutex is used to protect the status.Ready field from concurrent writes.
+	// sync.Once is used to ensure that we only set the status to not ready once, context will be cancelled after that.
+	var wg sync.WaitGroup
+	requests := make(chan struct{}, 5)
+	mu := sync.Mutex{}
+	once := sync.Once{}
```
tgross: The zero value for these are initialized, so you can use this more idiomatic definition:

```suggestion
	var mu sync.Mutex
	var once sync.Once
```

Also, this code is _very_ similar to the code you've introduced in `plugins/builtin/target/azure-vmss/plugin/azure.go`. Is it possible to pull out a reusable "rate limited instance views" method that you can return a response from? Maybe it takes a function as a parameter that closes over the response?

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+			if len(instanceView.Statuses) == 0 || *instanceView.Statuses[0].Code != "ProvisioningState/succeeded" {
+
+				for _, statusCode := range instanceView.Statuses {
+					if *statusCode.Code != "ProvisioningState/succeeded" {
+						t.logger.Debug("VM instance view not ready", "vm_name", *vm.Name, "status_code", *statusCode.Code)
+
+						once.Do(func() {
+							t.logger.Debug("Setting status to not ready", "vm_name", *vm.Name)
+							mu.Lock()
+							status.Ready = false
```
tgross: Do we actually need the mutex if we have a `sync.Once` and the caller is going to be waiting before reading status? Alternately (and this might be better for reuse) if we set a bool variable protected by the mutex to false and then set `status.Ready` to that value after the `wg.Wait`, could we avoid the `sync.Once` instead?

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +247,125 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+							mu.Unlock()
+							cancel()
```
tgross: Because we're canceling early here, we should probably put a context check in the outer loop as well (before `wg.Add(1)`. Otherwise we end up spawning potentially 1000s of goroutines that we don't need.

tgross (changes requested):
This is great work @benjamin-lykins. I've left a bunch of comments -- some on small details and some on design. For the smaller comments like contents of debug logging, try to take a pass over the whole PR with those in mind.

We've added a lot of new code here and it doesn't seem like there's a lot of test coverage. Unfortunately we don't have integration tests so that's challenging, but maybe as you break up some of those large methods I pointed out there'd be a way to extract some of the logic into testable chunks.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: cleaning up

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: updated with constants

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: removing comment.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: correct, removing that condition

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: Removed the loop, it was redundant.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: Ended up cleaning this up. More or less notes dropped when I was testing.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: added this to the start of the loop. this seems like the simplest way (at least in my mind to handle it). I saw another example using a select statement, but not sure if one is better practice than the other. 

```go
		if ctx.Err() != nil {
			t.logger.Debug("Context cancelled, stopping further processing of VMs")
			break
		}
```

go.mod
benjamin-lykins: Reran go mod tidy and it cleared this error up.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Removed comment

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Removed redundant conditional.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: Cleaned up the pointer and the downstream uses of it. Tested locally and no issues after my changes. 

Starting to look at consolidating code and generalizing.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: updated to consolidate the conditionals

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: fixed and put the error last. 

updated and no longer returning as a pointer, but as a slice.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: updated a couple of spots to check that any functions or processes actually return something back to give an early exit.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: this is the last thing I need to clean up. still used in two spots.

plugins/test/noop-apm/go.mod
```diff
@@ -10,14 +10,14 @@ require (
 
 require (
```
tgross, line 11: I'm not sure why all these plugins are getting dep updates? Do you need to rebase on `main`?

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,109 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+	for pager.More() {
+		page, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list VMSS instances: %w", err)
+		}
+
+		for _, vm := range page.Value {
+			if vm.Name == nil {
+				continue
+			}
```
tgross: The code below assumes that `vm.InstanceID` is also non-nil. I can't imagine why the Azure SDK has is as a pointer to a string or how it could possibly ever be nil, but because they did we should probably assume it could be nil and treat that safely to avoid crashes.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,109 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+
+			// Check for status codes on the instance view, only update remoteIDs with instances that have provisioned.
+			if len(instanceView.Statuses) == 0 {
+				for _, status := range instanceView.Statuses {
+					if *status.Code != "provisioningState/succeeded" {
+						t.logger.Debug("instance view status not ready", "vm_name", vm, "status_code", *status.Code)
+						return
+					}
+				}
+			}
```
tgross: The guard at the top of this block is backwards. It should be checking `!=0` (but also you don't need that for non-zero lists, because ranging over it already takes care of that for you).

We're also returning early if _any_ status we see is not `provisioningState/succeeded`. Can we guarantee that if the instance view has that state that it's always going to be the first one we see?

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,109 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+			// Unable to get flexible instanceView from the VMSS instance.
+			if vm.Properties != nil && vm.Properties.InstanceView != nil && vm.Properties.InstanceView.Statuses != nil {
+				for _, s := range vm.Properties.InstanceView.Statuses {
+					if s.Code != nil && *s.Code == "PowerState/running" {
+						t.logger.Debug("found healthy instance", "name", *vm.Name, "instance_id", *vm.InstanceID)
+						vmNames = append(vmNames, *vm.Name)
+						break
+					} else {
+						t.logger.Debug("skipping instance", "name", *vm.Name, "instance_id", *vm.InstanceID, "code", *s.Code)
+					}
```
tgross: This log line seems wrong... we're not skipping the instance here, we're iterating to the next `InstanceView.Status` and will print this log line for each status we find.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,109 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+				t.logger.Debug("skipping instance", "id", *vm.ID, "instance_id", *vm.InstanceID)
+			}
+
+		}
+	}
+
+	return vmNames, nil
+}
+
+func (t TargetPlugin) getFlexibleReadyRemoteIDs(ctx context.Context, resourceGroup string, vms []string) (ids []string, err error) {
```
tgross: We never return an error from this function. Which means the consumer is potentially getting a different length of `ids` then the `vms` input. Is that safe? Expected?

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -96,73 +103,122 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, resourceGroup string, vmSca
...
-			return fmt.Errorf("failed to list instances in VMSS: %v", err)
+			return fmt.Errorf("failed to get remote IDs: %w", err)
 		}
+	default:
+		return fmt.Errorf("unsupported VMSS mode: %s", vmssMode)
+	}
+
+	// early exit if no remote IDs found
+	if len(remoteIDs) == 0 {
+		return fmt.Errorf("no remoteIDs filtered")
```
tgross: See my note on `getFlexibleReadyRemoteIDs` about the error condition.

tgross (changes requested):
Making good progress!

plugins/test/noop-apm/go.mod
benjamin-lykins: not 100% sure, did a merge from main to resolve merge conflicts. in transparency, I hit the GitHub automagic-merge button. 

https://github.com/hashicorp/nomad-autoscaler/commit/349670efc0970939b4978c3e0053d1dafd391db3

then the ci checks started failing on the additional plugins wanting go mod tidy run. this could be ignorance on my part of doing something I probably didn't need to do in the first place.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Adding check for instanceID as well : 

```go
	if vm.Name == nil || vm.InstanceID == nil {
				continue
			}
```

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Ran through this a bit more in my head. This was carry over from what it was previously. I had tried to simplify this since it had multiple nested if statement. I ended up reverting back to what Luiz had previously. 

```go
if vm.Properties != nil && vm.Properties.InstanceView != nil && vm.Properties.InstanceView.Statuses != nil {
		for _, s := range vm.Properties.InstanceView.Statuses {
			if s.Code != nil && strings.HasPrefix(*s.Code, "PowerState/") {
				if *s.Code == "PowerState/running" {
					t.logger.Debug("found healthy instance", "name", *vm.Name, "instance_id", *vm.InstanceID)
					vmNames = append(vmNames, *vm.Name)
					break
				} else {
					t.logger.Debug("skipping instance - power state is not running", "name", *vm.Name, "instance_id", *vm.InstanceID, "code", *s.Code)
				}
			}
		}
```

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Cleaned this up a bit more and thinking back to the existing code and plugin, it checks for both provisioned and powered on VMs. 

I updated the function a bit to do that and setting todos for the other status and health checks. 


```go
func (t TargetPlugin) getFlexibleReadyRemoteIDs(ctx context.Context, resourceGroup string, vms []string) (ids []string, err error) {

	remoteIDs := []string{}

	var wg sync.WaitGroup
	// Manually limiting the number of concurrent requests to avoid hitting API rate limits.
	requests := make(chan struct{}, 5)
	mu := sync.Mutex{}

	for _, vm := range vms {
		wg.Add(1)
		requests <- struct{}{}

		go func(vm string) {
			defer wg.Done()
			defer func() { <-requests }()

			select {
			case <-ctx.Done():
				return
			default:
				// Continue processing
			}

			// Get the instance view for the VM in the Flexible VMSS.
			instanceView, err := t.vm.InstanceView(ctx, resourceGroup, vm, nil)
			if err != nil {
				t.logger.Debug("failed to get instance view for VM", "vm_name", vm, "error", err)
				return
			}

			// Checks if instance is running and successfully provisioned.
			var provisioningReady, powerRunning bool
			for _, s := range instanceView.Statuses {
				if s.Code == nil {
					continue
				}

				switch *s.Code {
				case "ProvisioningState/succeeded":
					provisioningReady = true
				case "PowerState/running":
					powerRunning = true
				default:
					t.logger.Debug("unexpected status code", "vm_name", vm, "status_code", *s.Code)
				}
			}

			if !provisioningReady || !powerRunning {
				t.logger.Debug("instance not ready", "vm_name", vm, "provisioning_ready", provisioningReady, "power_ready", powerRunning)
				return
			}

			t.logger.Debug("instance candidate for scale in action", "vm_name", vm)

			mu.Lock()
			remoteIDs = append(remoteIDs, vm)
			mu.Unlock()

		}(vm)
	}
	wg.Wait()

	return remoteIDs, nil
}
```

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: So a different length _should_ be fine. 

Essentially this is doing what Uniform is able to do with InstanceView from the VMSS api. 

In this case, with Flexible, I can only pull if a VMSS instance provisioned successfully. To get its power state I have to iterate using the Azure VM API. So I could input 200 VMs to check, but only 50 are provisioned and powered on. 

I will probably create an error in this function, if there are no VMs found to be candidates to scale in. Instead of here:

```go
	// early exit if no remote IDs found
	if len(remoteIDs) == 0 {
		return fmt.Errorf("no remoteIDs filtered")
```

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Will move this into the function and output as an error and check for an error where nothing eligible found for scale in actions.

benjamin-lykins: updated and removed this.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Updated, function outputs an error and gets handled

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: @tgross Tried to take your advice, but went a different route. I hope this works out, it makes a bit more sense for my smooth brain. Regarding the bool variable, I added it in and put it at the end after wg.Wait. 

Instead of sync.Once or mutex, I dug into it a bit more and used the atomic package based on some [articles I read](https://www.codingexplorations.com/blog/understanding-golangs-atomic-package-and-mutexes). Seems like this is a use case where using it with CompareAndSwap would be simpler? 

Updated function: 

```go
func (t *TargetPlugin) processInstanceViewFlexible(vms []string, resourceGroup string, status *sdk.TargetStatus) {
	start := time.Now()

	if len(vms) == 0 {
		t.logger.Debug("No VMs found in the VMSS, skipping instance view processing.")
		return
	}

	t.logger.Debug("Total VMs found in the Flexible VMSS", "count", len(vms))

	ctx, cancel := context.WithCancel(context.Background())
	defer cancel()

	var wg sync.WaitGroup
	requests := make(chan struct{}, 5)

	var notReady atomic.Bool
	var statusReady = true

	for _, vmName := range vms {
		if ctx.Err() != nil {
			t.logger.Debug("Context cancelled, stopping further processing of VMs")
			break
		}

		wg.Add(1)
		requests <- struct{}{}

		go func(vm string) {
			defer wg.Done()
			defer func() { <-requests }()

			select {
			case <-ctx.Done():
				return
			default:
			}

			instanceView, err := t.vm.InstanceView(ctx, resourceGroup, vm, nil)
			if err != nil {
				t.logger.Debug("Failed to get instance view for VM", "vm_name", vm, "error", err)
				return
			}

			if !isFlexibleVMReady(instanceView.Statuses) {
				t.logger.Debug("VM not ready", "vm_name", vm, "statuses", instanceView.Statuses)

				if notReady.CompareAndSwap(false, true) {
					t.logger.Debug("Setting status to not ready and cancelling context", "vm_name", vm)
					statusReady = false
					cancel()
				}
				return
			}

			t.logger.Debug("VM instance view is ready", "vm_name", vm)
		}(vmName)
	}
	wg.Wait()

	t.logger.Debug("Setting status", statusReady)
	status.Ready = statusReady
	t.logger.Debug("Finished processing VM instance views", "duration_seconds", time.Since(start).Seconds())
}
```

Added a helper function too which is used by both the status check processes and scale in process

```go
// isFlexibleVMReady checks whether the Flexible VMSS VM has both ProvisioningState/succeeded and PowerState/running.
func isFlexibleVMReady(statuses []*armcompute.InstanceViewStatus) bool {
	var provisioned, poweredOn bool

	for _, s := range statuses {
		if s.Code == nil {
			continue
		}

		switch *s.Code {
		case "ProvisioningState/succeeded":
			provisioned = true
		case "PowerState/running":
			poweredOn = true
		}
	}

	return provisioned && poweredOn
}
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,126 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+			mu.Lock()
+			remoteIDs = append(remoteIDs, vmName)
+			mu.Unlock()
+		}(vm)
+	}
+
+	wg.Wait()
+
+	if len(remoteIDs) == 0 {
+		t.logger.Error("No instances found to be provisioned and running.")
```
tgross: Won't the error bubble up somewhere to get logged?

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -190,3 +246,126 @@ func azureNodeIDMap(n *api.Node) (string, error) {
...
+			if vm.Name == nil || vm.InstanceID == nil {
+				continue
+			}
+
+			// Check for PowerState/running if instanceView is available.
+			// Unable to get flexible instanceView from the VMSS instance.
+			if vm.Properties != nil && vm.Properties.InstanceView != nil && vm.Properties.InstanceView.Statuses != nil {
+				for _, s := range vm.Properties.InstanceView.Statuses {
+					if s.Code != nil && strings.HasPrefix(*s.Code, "PowerState/") {
+						if *s.Code == "PowerState/running" {
```
tgross: We don't handle any other power states, so I think we can just do this:

```suggestion
					if s.Code != nil && *s.Code == "PowerState/running" {
```

plugins/test/noop-apm/go.mod
tgross: Probably getting pulled in because of the Nomad API update you've got, which seems orthogonal to this PR.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -96,73 +103,122 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, resourceGroup string, vmSca
...
+				instanceIDs = append(instanceIDs, node.RemoteResourceID[idx+1:])
+			} else {
+				return errors.New("failed to get instance-id from remoteid")
+			}
+		}
+	case orchestrationModeFlexible:
+		for _, node := range ids {
+			log.Debug("processing node for scale in", "node_id", node, "remote_id", node.RemoteResourceID)
+			// You'll notice that the logic here is different than Uniform mode. This is mainly due to Azure's consistency of being inconsistent.
+			// https://learn.microsoft.com/en-us/azure/virtual-machine-scale-sets/virtual-machine-scale-sets-instance-ids#scale-set-vm-names
```
tgross: The only difference is how we extract the ID from the `RemoteResourceID`, right? Having only a 2 character difference between the two repeated cases makes this a little hard to read. Can we wrap the extractor into a little function so we can just loop over the IDs? Ex. 

```go
for _, node := ids {
	id, err := idFromRemoteID(vmssMode, node)
	if err != nil {
		return err
	} 
	instanceIDs =  append(instanceIDs, node.RemoteResourceID)
}
```

We can drop all the debug logging, as anything that goes wrong is an error anyways. Then you've got a helper function something like (maybe could use a better name, error messages):

```go
func idFromRemoteID(mode string, remoteResourceID string) (string, error) {
	switch mode {
	case orchestrationModeUniform:
		if idx := strings.LastIndex(remoteResourceID, "_"); idx != -1 &&
			strings.EqualFold(remoteResourceID[0:idx], vmScaleSet) &&
			len(remoteResourceID) >= idx+1 {
			return remoteResourceID[idx+1:], nil
		}

	case orchestrationModeFlexible:
		if strings.HasPrefix(remoteResourceID, vmScaleSet) {
			return remoteResourceID, nil
		}
	}
	return "", errors.New("lolwut")
}
```

There are plenty of times to "pull ifs up and fors down" but I don't think this is going to be one of them.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +246,93 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			instanceView, err := t.vm.InstanceView(ctx, resourceGroup, vm, nil)
+			if err != nil {
+				t.logger.Debug("Failed to get instance view for VM", "vm_name", vm, "error", err)
```
tgross: This seems like it's more important than `Debug`?

benjamin-lykins: updated from debug to error. most likely was a bad copy and paste that should have been updated.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: updated both sections. cleaned up the helper function to be a bit more descriptive and add in the input for vmScaleSet. 

This should be an easy function to add for testing as well. I'll see what else I can break off and test. 

```go
func idFromRemoteID(mode string, remoteResourceID string, vmScaleSet string) (string, error) {
	switch mode {
	case orchestrationModeUniform:
		// For Uniform mode, remoteResourceID is expected to be in the format "<vmss_name>_<instance_id>"
		if idx := strings.LastIndex(remoteResourceID, "_"); idx != -1 &&
			strings.EqualFold(remoteResourceID[:idx], vmScaleSet) &&
			len(remoteResourceID) > idx+1 {
			return remoteResourceID[idx+1:], nil
		}
		return "", fmt.Errorf("invalid remoteResourceID format for Uniform mode: %q", remoteResourceID)

	case orchestrationModeFlexible:
		// For Flexible mode, remoteResourceID is the VM name, which should start with the vmScaleSet name
		if strings.HasPrefix(remoteResourceID, vmScaleSet) {
			return remoteResourceID, nil
		}
		return "", fmt.Errorf("invalid remoteResourceID format for Flexible mode: %q", remoteResourceID)
	}
	// This should be caught during the initial plugin initialization
	// Adding for now as a placeholder
	return "", fmt.Errorf("unsupported orchestration mode: %q", mode)
}
```

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: ahh it should. I'll fix that.

benjamin-lykins:
Azure API rate limiting no longer an issue, now bumping into Nomad rate limiting issue. The other plugins have a retry function which I'll look to incorporate. Probably not in this PR, but in a future I will add better handling if the node drain occurs on some nodes, but not all to re-enable them for allocations.

benjamin-lykins:
> > now bumping into Nomad rate limiting issue.
> 
> Once #1134 has been merged, this should reduce the impact of this and make it easier to wrap this up.

Ahh great! I'll hold off, since I had been trying to work at what was available to handle that. Thanks for the heads up @tgross.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -49,14 +55,20 @@ var _ target.Target = (*TargetPlugin)(nil)
...
+	vmss    *armcompute.VirtualMachineScaleSetsClient
+	vmssVMs *armcompute.VirtualMachineScaleSetVMsClient
 
 	// clusterUtils provides general cluster scaling utilities for querying the
 	// state of nodes pools and performing scaling tasks.
 	clusterUtils *scaleutils.ClusterScaleUtils
+
+	// readyInstances is a list of instances that are ready for scale in actions.
+	// used with the flexible orchestration mode to prevent the need for multiple calls to the Azure API to check.
+	readyFlexibleInstances []string
```
tgross: It looks like this is getting updated by the `Status` RPC (in `plugin.go`) which could be called concurrently with the `Scale` RPC (which ultimately calls `scaleIn`/`scaleOut`), so we're going to need to put a mutex around this to avoid race conditions between setting it and getting it:

```suggestion
	// readyInstances is a list of instances that are ready for scale in actions.
	// used with the flexible orchestration mode to prevent the need for multiple calls to the Azure API to check.
	readyFlexibleInstances []string
	readyFlexibleInstancesLock sync.Mutex
```

Then wherever you read/write to it, you'll do:

```go
t.readyFlexibleInstancesLock.Lock()
defer t.readyFlexibleInstancesLock.Unlock()
t.readyFlexibleInstances = newValue
```

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -96,82 +102,112 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, resourceGroup string, vmSca
...
 	if err != nil {
-		return fmt.Errorf("failed to perform pre-scale Nomad scale in tasks: %v", err)
+
+		// Possible to be hit rate limit if trying to scale too many nodes at once.
+		if strings.Contains(err.Error(), "Unexpected response code: 429") {
+			return fmt.Errorf("rate limit exceeded while performing pre-scale Nomad scale in tasks: %w", err)
+		}
+
+		// TODO: Need to handle the case where the pre-scale tasks fail
+		// for nodes which were made ineligibe, we would need to revert.
```
tgross: We still have this open TODO. Should we try to close that out in this PR or should that land in a follow-up? I'm not sure I know what the consequences of not having that would be here.

plugins/builtin/target/azure-vmss/plugin/azure.go
tgross: I think this still has the same problem. Suppose I have the following set of [`InstanceViewStatus`](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/compute/armcompute/v7#InstanceViewStatus):

```go
 []InstanceViewStatus{
	{ Code: pointer.Of("ProvisioningState/succeeded"), ...}
	{ Code: pointer.Of("Whatever/foo"), ...}
	{ Code: pointer.Of("PowerState/running"), ...}
}
```

In this case, we'll log that we're skipping the instance twice and then not skip it, because we eventually find a `"PowerState/running"` code. The number of times we log ends up being dependent on the order that the statuses get returned (which I can't find anywhere in the Azure docs). So we only want to log if we _never_ find that `"PowerState/running"` code. Similar to how you've got `isFlexibleVMReady` working.


Or we can drop this `else` block entirely, which I think would also be fine.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +251,99 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
 		}
 	}
 }
+
+func (t *TargetPlugin) processInstanceViewFlexible(vms []string, resourceGroup string, status *sdk.TargetStatus) {
+	readyInstances := []string{}
+	var mu sync.Mutex
+
+	t.logger.Debug("Processing VM instance views for Flexible VMSS", "vm_count", len(vms))
+	t.readyFlexibleInstances = []string{}
```
tgross, line 292: When we add the mutex guard on this, we'll want to avoid holding the lock open for this entire function.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +251,99 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			instanceView, err := t.vm.InstanceView(ctx, resourceGroup, vm, nil)
+			if err != nil {
+				if strings.Contains(err.Error(), "RESPONSE 404") {
+					t.logger.Debug("VM not found", "vm_name", vm)
+				} else if strings.Contains(err.Error(), "context canceled") || err == context.Canceled {
```
tgross: We shouldn't need this `strings.Contains` clause. Try `errors.Is(err, context.Canceled)`

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +251,99 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	var wg sync.WaitGroup
+	requests := make(chan struct{}, concurrentInstanceViewRequestLimit)
+	var notReady atomic.Bool
+
+	for _, vmName := range vms {
+		if ctx.Err() != nil {
+			t.logger.Debug("Context cancelled, stopping further processing of VMs")
```
tgross: We'll already have a log line for the cancelation, this just adds a pile of identical debug lines for every remaining VM:

```suggestion
```

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -203,28 +251,99 @@ func (t *TargetPlugin) calculateDirection(vmssDesired, strategyDesired int64) (i
...
+			select {
+			case <-ctx.Done():
+				return
+			default:
+			}
+
+			instanceView, err := t.vm.InstanceView(ctx, resourceGroup, vm, nil)
+			if err != nil {
+				if strings.Contains(err.Error(), "RESPONSE 404") {
+					t.logger.Debug("VM not found", "vm_name", vm)
```
tgross: Should this be treated as "not ready" too? What about a non-404/non-context error?

tgross (changes requested):
@benjamin-lykins this is really getting there. I've left a few small comments but nothing that's a major design issue at this point. We should also add a changelog entry to the Unreleased section, so that when we release this we have it noted.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: This should be good now. When I tested last, it looks to be fixed in #1134. 

Essentially, the prescale in tasks would get rate limited. The scale in would fail, and the node pool would b in a state where x amount of clients would be marked ineligible for scheduling. 

Looking at this now, I can probably clean this up a bit more.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: Good call. I'll add this in.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: removed.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: There is a bit more to this, and I think I can handle it a little bit better thinking about it now. There is an odd bug I've found with flexible vmss, where if you request x amount of instances, but it is only partially fulfilled - azure will still return names of instances which do not exist. I'll get back to this one soon after I knock out some of the others.

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: updated.

plugins/builtin/target/azure-vmss/plugin/azure.go
benjamin-lykins: Updated and simplified

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: I believe I have a better way to fix this. I just need to rework how I pull in the list of flexible vmss vms.

benjamin-lykins: Just some context in the issue. Requested 200 VMs. Only 174 provision. 

image

In Nomad, I see only 174 as well. 

image

When running the Nomad Autoscaler it sees 200 when running in the `getVMSSVMs` function. 

```
2025-11-03T14:48:20.968-0500 [DEBUG] external_plugin.azure-vmss: Processing VM instance views for Flexible VMSS: vm_count=200 timestamp=2025-11-03T14:48:20.966-0500
```

When it runs this function `processInstanceViewFlexible`, that is where it is getting the not found errors. 

```
_c7f44d95 timestamp=2025-11-03T14:46:24.782-0500
2025-11-03T14:46:24.797-0500 [DEBUG] external_plugin.azure-vmss: VM not found: vm_name=flexible-vmss_c8793a17 timestamp=2025-11-03T14:46:24.797-0500
```

I'm going to update the `getVMSSVMs` for flexible and do additional verification that the vm exists.

benjamin-lykins: Reworked this a bit, added additional verification that the flexible vm can be found. Output below has shows it is unable to find 26 vms, which lines up that it could only find 174 of 200 vms. 

```go
func (t *TargetPlugin) verifyFlexibleVM(ctx context.Context, resourceGroupName string, vmName string, options *armcompute.VirtualMachinesClientGetOptions) bool {
	_, err := t.vm.Get(ctx, resourceGroupName, vmName, options)
	if err != nil {
		t.logger.Debug("failed to get VM", "name", vmName)
		return false
	}
	return true
}
```

This provides a cleaner list into the `processInstanceViewFlexible` function later on. 

```
2025-11-03T20:54:23.113-0500 [DEBUG] external_plugin.azure-vmss: getting Azure ScaleSet status: resource_group=flexible-vmss vmss_name=flexible-vmss timestamp=2025-11-03T20:54:23.113-0500
2025-11-03T20:54:24.191-0500 [DEBUG] external_plugin.azure-vmss: VMSS Orchestration Mode: mode=Flexible timestamp=2025-11-03T20:54:24.191-0500
2025-11-03T20:54:26.107-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_06179162 timestamp=2025-11-03T20:54:26.107-0500
2025-11-03T20:54:26.301-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_0673a511 timestamp=2025-11-03T20:54:26.301-0500
2025-11-03T20:54:27.027-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_0bd6e175 timestamp=2025-11-03T20:54:27.027-0500
2025-11-03T20:54:28.024-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_0f14e5cc timestamp=2025-11-03T20:54:28.024-0500
2025-11-03T20:54:28.524-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_13bab681 timestamp=2025-11-03T20:54:28.524-0500
2025-11-03T20:54:31.644-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_37a5aa1f timestamp=2025-11-03T20:54:31.643-0500
2025-11-03T20:54:32.685-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_4073fee5 timestamp=2025-11-03T20:54:32.684-0500
2025-11-03T20:54:33.393-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_4dd3491e timestamp=2025-11-03T20:54:33.393-0500
2025-11-03T20:54:33.878-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_5506d4b9 timestamp=2025-11-03T20:54:33.878-0500
2025-11-03T20:54:33.962-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_56bcd5c8 timestamp=2025-11-03T20:54:33.962-0500
2025-11-03T20:54:35.601-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_6ad64883 timestamp=2025-11-03T20:54:35.601-0500
2025-11-03T20:54:36.844-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_7d847942 timestamp=2025-11-03T20:54:36.844-0500
2025-11-03T20:54:36.910-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_7dadd211 timestamp=2025-11-03T20:54:36.909-0500
2025-11-03T20:54:38.371-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_8a9215dc timestamp=2025-11-03T20:54:38.371-0500
2025-11-03T20:54:38.568-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_8ed6466a timestamp=2025-11-03T20:54:38.568-0500
2025-11-03T20:54:39.251-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_97e5d6eb timestamp=2025-11-03T20:54:39.251-0500
2025-11-03T20:54:39.403-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_9919c483 timestamp=2025-11-03T20:54:39.403-0500
2025-11-03T20:54:39.467-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_9a03243d timestamp=2025-11-03T20:54:39.467-0500
2025-11-03T20:54:41.104-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_ae9d2eeb timestamp=2025-11-03T20:54:41.103-0500
2025-11-03T20:54:41.841-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_b4e8c8c9 timestamp=2025-11-03T20:54:41.841-0500
2025-11-03T20:54:42.874-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_be2bf076 timestamp=2025-11-03T20:54:42.874-0500
2025-11-03T20:54:46.156-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_e8ebf2d8 timestamp=2025-11-03T20:54:46.156-0500
2025-11-03T20:54:46.589-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_f0f21b17 timestamp=2025-11-03T20:54:46.589-0500
2025-11-03T20:54:46.731-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_f1329632 timestamp=2025-11-03T20:54:46.731-0500
2025-11-03T20:54:47.465-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_fc36a2e8 timestamp=2025-11-03T20:54:47.465-0500
2025-11-03T20:54:48.008-0500 [DEBUG] external_plugin.azure-vmss: failed to get VM: name=flexible-vmss_ff30d069 timestamp=2025-11-03T20:54:48.008-0500
2025-11-03T20:54:48.082-0500 [DEBUG] external_plugin.azure-vmss: Processing VM instance views for Flexible VMSS: vm_count=174 timestamp=2025-11-03T20:54:48.082-0500
2025-11-03T20:54:51.771-0500 [INFO]  external_plugin.azure-vmss: VM not ready: statuses=["map[code:ProvisioningState/failed/AllocationFailed displayStatus:Provisioning failed level:Error message:The VM allocation failed due to an internal error. Please retry later or try deploying to a different location. time:2025-11-04T01:23:13.5835359Z]"] vm_name=flexible-vmss_cd9c78b0 timestamp=2025-11-03T20:54:51.770-0500
2025-11-03T20:54:51.771-0500 [INFO]  external_plugin.azure-vmss: Setting status to not ready and cancelling context: vm_name=flexible-vmss_cd9c78b0 timestamp=2025-11-03T20:54:51.770-0500
2025-11-03T20:54:51.771-0500 [DEBUG] external_plugin.azure-vmss: Context cancelled during instance view retrieval: vm_name=flexible-vmss_d1ba788e timestamp=2025-11-03T20:54:51.771-0500
2025-11-03T20:54:51.771-0500 [DEBUG] external_plugin.azure-vmss: Context cancelled during instance view retrieval: vm_name=flexible-vmss_d0cbbb1a timestamp=2025-11-03T20:54:51.771-0500
2025-11-03T20:54:51.771-0500 [DEBUG] external_plugin.azure-vmss: Context cancelled during instance view retrieval: vm_name=flexible-vmss_cc1302f4 timestamp=2025-11-03T20:54:51.771-0500
2025-11-03T20:54:51.771-0500 [DEBUG] external_plugin.azure-vmss: Context cancelled during instance view retrieval: vm_name=flexible-vmss_beea14cb timestamp=2025-11-03T20:54:51.771-0500
```

plugins/builtin/target/azure-vmss/plugin/plugin.go
benjamin-lykins: updated in both locations

benjamin-lykins:
@tgross made a series of changes off of your last set of reviews and ready for rereview.

plugins/builtin/target/azure-vmss/plugin/plugin.go
```diff
@@ -101,31 +114,34 @@ func (t *TargetPlugin) Scale(action sdk.ScalingAction, config map[string]string)
...
 
-	currVMSS, err := t.vmss.Get(ctx, resourceGroup, vmScaleSet)
+	currVMSS, err := t.vmss.Get(ctx, resourceGroup, vmScaleSet, nil)
 	if err != nil {
-		return fmt.Errorf("failed to get Azure vmss: %v", err)
+		return fmt.Errorf("failed to get Azure vmss: %w", err)
 	}
 
-	capacity := *currVMSS.Sku.Capacity
+	capacity := *currVMSS.SKU.Capacity
```
benjamin-lykins, line 145: @tgross I'm going to rework this section as well. It has always pulled what the total number of instances _can be_ in either an orchestrated or flexible vmss. I think it would be better to grab the actual count.

benjamin-lykins: Updated and tested with the following successfully: 

```go
	// Recommended removing this since it only gets the requested capacity and not the actual count.
	// capacity := *currVMSS.SKU.Capacity

	vms, err := t.getVMSSVMs(ctx, resourceGroup, orchestrationMode, vmScaleSet)

	if err != nil {
		return fmt.Errorf("failed to get VMSS VMs: %w", err)
	}

	capacity := int64(len(vms))
```

I'll find some time to run this against a uniform set and confirm scaling actions perform without issue as well for all the changes. Last time I did, there were no issues, but wanting to double check since it has been a bit.

plugins/builtin/target/azure-vmss/plugin/azure.go
```diff
@@ -96,81 +102,108 @@ func (t *TargetPlugin) scaleOut(ctx context.Context, resourceGroup string, vmSca
...
+
+	// additional processing needed for flexible mode to get status of VMs
+	// for uniform, populate remoteIDs with VMs
+	var remoteIDs []string
+	log.Debug("getting ready remote IDs for Azure ScaleSet instances")
+	switch vmssMode {
+	case orchestrationModeUniform:
+		remoteIDs = vms
+	case orchestrationModeFlexible:
+		remoteIDs = t.readyFlexibleInstances
```
tgross: We need the lock here as well because we're reading the value. And because `t.readyFlexibleInstances` is a reference type (a slice, see [an example of that behavior](https://go.dev/play/p/HZVUTBBiBAB)), you'll want to shallow-copy it in the process:


```suggestion
		t.readyFlexibleInstancesLock.Lock()
		remoteIDs = slices.Clone(t.readyFlexibleInstances)
		t.readyFlexibleInstancesLock.Unlock()
```

benjamin-lykins: thank you! I got this committed and pushed up.

tgross (approved):
LGTM! Nice work!

Issue #1097, opened 2025-05-30
Latest nightly build not available at hashicorppreview docker hub


caio-vinicius:
Hello. I'm trying to deploy the latest build to confirm the bug related to the policy discussed in https://github.com/hashicorp/nomad-autoscaler/pull/1071. Is it possible to trigger a new build to preview it?

tgross:
Hi @caio-vinicius I'm not sure I know how that was getting published to begin with, as nightlies are only supposed to be getting released via GitHub releases. I'll mark this issue for investigation.

But you can grab a nightly image from https://github.com/hashicorp/nomad-autoscaler/releases/tag/nightly

tgross:
It looks like a security scanner is tripping over the base image and this didn't create an alert. I'll update the base image and hopefully that'll get the pipeline unstuck.

PR #1108, opened 2025-06-09
build: bump alpine base image to 3.22

tgross:
The build pipeline's security scanner is picking up a (not meaningful) vuln in the Alpine base image we're using, which is preventing nightlies from being published. Rev the alpine base image.

Ref: https://github.com/hashicorp/crt-workflows-common/actions/runs/15543056919/job/43758419778
Fixes: https://github.com/hashicorp/nomad-autoscaler/issues/1097


gulducat (approved):
:two: :two:

PR #1115, opened 2025-06-23
F rate limiter

Juanadelacuesta:
The Nomad Autoscaler will periodically query the APM defined in the policy for metrics. This can lead to [a problem in hosted/SaaS offerings](https://discuss.hashicorp.com/t/best-practices-for-using-the-nomad-autoscaler-with-saas-apm-tools/18652) that have strict rate-limits, where checks will fail due to limits being hit.

This PR introduces a general purpose rate limiter and adds a configuration option to the Data Dog APM configuration and the corresponding implementation to be able to use it.

### DIFF
--- a/rate_limiter/rate_limiter.go
+++ b/rate_limiter/rate_limiter.go
@@ -21,31 +21,27 @@ type CustomRoundTripper struct {
 }
 
 func (crt *CustomRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
-	if irt.rateLimiter != nil {
-		err := irt.rateLimiter.Wait(req.Context())
-		if err != nil {
-			return nil, fmt.Errorf("transport: unable to ratelimit: %w", err)
-		}
-	}
-
 	labels := []metrics.Label{
 		{
 			Name:  "method",
 			Value: req.Method,
 		},
 		{
-			Name:  "path",
-			Value: req.URL.Path,
-		},
-		{
 			Name:  "source",
-			Value: irt.source,
+			Value: crt.source,
 		},
 	}
 
 	defer metrics.MeasureSinceWithLabels([]string{"http", "dur"}, time.Now(), labels)
 
-	resp, err := irt.rt.RoundTrip(req)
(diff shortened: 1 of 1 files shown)


rate_limiter/rate_limiter.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// CustomRoundTripper wraps http.RoundTripper to observe metrics and rate limit if necessary
+type CustomRoundTripper struct {
+	rateLimiter *rate.Limiter
+	source      string
+	rt          http.RoundTripper
+}
+
+func (irt *CustomRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```
tgross: Extremely nitpicky :grin: , feel free to ignore:

```suggestion
func (crt *CustomRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
```

rate_limiter/rate_limiter.go
```diff
@@ -0,0 +1,78 @@
...
+			Name:  "path",
+			Value: req.URL.Path,
+		},
+		{
+			Name:  "source",
+			Value: irt.source,
+		},
+	}
+
+	defer metrics.MeasureSinceWithLabels([]string{"http", "dur"}, time.Now(), labels)
```
tgross, line 35: This makes it so that the time we measure doesn't include the wait time for rate limiting. Should it be at the beginning to make sure that metrics can show when a plugin is getting rate limited?

tgross (commented):
Looks great @Juanadelacuesta. Should we include documentation for this configuration option?

Juanadelacuesta:
@tgross the docs are in the main nomad repo: https://github.com/hashicorp/nomad/tree/docs-autoscaler-rate-limiter Im waiting for this one to open the other one.

rate_limiter/rate_limiter.go
Juanadelacuesta: Makes sense, I will move it.

Issue #272, opened 2020-09-25
Add ability to have different CoolUp vs CoolDown timeouts


urjitbhatia:
Hi guys,

In our home-grown autoscaler, we have different cooldown policies for scaling up v/s down allowing our services to respond to rapid traffic spikes by scaling up much more aggressively than scaling down.

By having a single value, an autoscaler will be forced to flip-flop more when the resource consumption is right around the edge since users are forced to use a shorter cooldown to account for fast scale-ups. What do you guys think? Happy to send a quick PR if there is any appetite for this.

cgbaker:
hi @urjitbhatia , this sounds good to me! we'll put this on the list of features, and we would definitely consider a PR on this. thanks for your suggestion!

lgfa29:
Hi @urjitbhatia,

That's a good point and something we have plans to fix. The biggest issue right now is that when a cooldown is triggered the goroutine that tracks the policy is [blocked](https://github.com/hashicorp/nomad-autoscaler/blob/v0.1.1/policy/handler.go#L334), so even if a load increase happens there's nothing looking for it. There's also no way right to "wake up" the policy handler apart from shutting it down.

We will refactor the entire policy evaluation flow in a way that allows for this (and other) capabilities to be easily added, including different strategies to avoid count flapping. 

But as @cgbaker mentioned, we would gladly consider a PR if you find a quick fix. I would just not try too hard right now as better things are coming 😄

elcomtik:
any news?

lgfa29:
> any news?

No updates on this yet unfortunately. The improvement that I mentioned before turned out to be a bit more complex than we expected so it hasn't been implemented yet.

the-maldridge:
This bit me earlier today.  Is there a recommended way to do horizontal cluster scaling until this is fixed?  Without this you wind up with either a cluster that has no capacity for an extended period of time, or one that thrashes machines.

lgfa29:
> This bit me earlier today. Is there a recommended way to do horizontal cluster scaling until this is fixed? Without this you wind up with either a cluster that has no capacity for an extended period of time, or one that thrashes machines.

@the-maldridge I can't think of anything in particular that would help with this 😞 

I will try to revisit this in the coming days. The first problem is that the cooldown is checked inadvertently in a few places, so there's an unfortunate dynamic on how the sub-components interact with it.

Eyald6:
[...]

PR #1116, opened 2025-06-24
Add a different cooldown on scale up option

Juanadelacuesta:
Having a different cooldown period for scaling up allows for a faster and more aggressive response in times of spikes. This PR introduces the option to add this new configuration, that defaults to the normal cooldown period if not provided.

The cool down period is set and enforced in 2 different places in the code:

1. After a regular [scale event](https://github.com/hashicorp/nomad-autoscaler/blob/5e0cf9f9a8ea08ede6f1b070239578d5a32f1f48/policyeval/base_worker.go#L337)
2. At a policy handler evaluation time [after a disaster recovery or failover](https://github.com/hashicorp/nomad-autoscaler/blob/5e0cf9f9a8ea08ede6f1b070239578d5a32f1f48/policy/handler.go#L148), what is referred to as "out of band events".

This PR only handles the first place, the second will be handled later because given the fact that the autoscaler is stateless, it relies on the providers to look up for ongoing scaling events, and the current APIs don't have a straight forward way to extract the direction of the current or past scaling events, making it impossible to respect the coldown on scale up configuration.

### DIFF
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -205,7 +205,7 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
 // enforceCooldown blocks until the cooldown period has been reached, or the
 // handler has been instructed to exit. The boolean return details whether or
 // not the cooldown period passed without being interrupted.
-func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) (complete bool) {
+func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) bool {
 
 	// Log that cooldown is being enforced. This is very useful as cooldown
 	// blocks the ticker making this the only indication of cooldown to
@@ -221,12 +221,12 @@ func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) (complet
 	// Cooldown should not mean we miss other handler control signals. So wait
 	// on all the channels desired here.
 	select {
-	case <-timer.C:
-		complete = true
-		return
 	case <-ctx.Done():
-		return
+		return false
+	case <-timer.C:
 	}
+
+	return true
 }
 
 // calculateRemainingCooldown calculates the remaining cooldown based on the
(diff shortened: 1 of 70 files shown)


policy/handler.go
```diff
@@ -205,7 +206,19 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
 // enforceCooldown blocks until the cooldown period has been reached, or the
 // handler has been instructed to exit. The boolean return details whether or
 // not the cooldown period passed without being interrupted.
-func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) (complete bool) {
+func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) bool {
+	metrics.SetGaugeWithLabels([]string{"cooldown", "test"}, 1, []metrics.Label{
```
tgross: `"test"`? Is this left over from debugging?

policy/nomad/test-fixtures/src/invalid-cooldown-on-scale-up.hcl
```diff
@@ -0,0 +1,27 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+job "invalid-cooldown-on-scale-up" {
+  datacenters = ["dc1"]
```
tgross: Nitpicky: we don't need the `datacenters` field anymore. It defaults to `["*"]`

tgross (commented):
Code-wise this looks great @Juanadelacuesta but I'm not sure I understand some of the design intent. Whichever action happened last is setting the cooldown timeout, which strictly speaking fulfills the feature request. But doesn't that mean that if there's a longer timeout for one action that it blocks the other action from happening?

My understanding of how the policy enforcement happens is that in `policy/handler.go` is that the cooldown prevents the policy from being evaluated again entirely (see also [this comment](https://github.com/hashicorp/nomad-autoscaler/issues/272#issuecomment-699113157)). So if I have `cooldown = "5m"` and `cooldown_for_scaleup = "2m"` and I have a scaledown, I now have to wait 5m for the policy to get evaluated again even if that evaluation will result in a scaleup. It feels like as a user I'd want the policy evaluation to happen but for the cooldown to impact the scaling operation rather than the evaluation. 

But maybe I'm misunderstanding how this works in practice?

policy/handler.go
Juanadelacuesta: a yes, it was a "temp" commit I forgot to remove, sorry

tgross (approved):
LGTM!

Issue #1124, opened 2025-07-10
Segmentation Violation when starting policy manager when nil policy is encountered.


tristanmorgan:
### Nomad Autoscaler version
Nomad Autoscaler v0.4.7 (2b9cf6f)
Nomad v1.10.3

### Operating system and Environment details
Linux HashiCarp 6.12.27-v8+ #1876 SMP PREEMPT Tue May  6 15:12:28 BST 2025 aarch64 GNU/Linux
Docker Version: 28.2.2

### Issue
A segmentation Violation is occurring during program startup when trying to parse the policies from Nomad. Some policies are defined as just a min and max and without any scaling policies and checks. I think its theses that are triggering the exception.

### Reproduction steps
Run a nomad job with the following scaling block and Nomad Autoscaler will fail to start with an error.
```
    scaling {
      enabled = true
      min     = 1
      max     = 10
    }
```

#### Expected Result
Expected to ignore the scaling policy block during evaluations because it has no policy to evaluate. (as previous versions did.)

#### Actual Result
```
panic: runtime error: invalid memory address or nil pointer dereference
```

### Job file (if appropriate)
```
job "webapp" {
  datacenters = ["dc1"]

  group "demo" {
    network {
      mode = "host"
      port "http" {
      }
    }

    service {
      port = "http"
      name = "webapp"
      tags = ["urlprefix-webapp.service.consul/"]
      check {
        port     = "http"
        type     = "http"
        path     = "/index.html"
        method   = "HEAD"
        interval = "10s"
        timeout  = "2s"
      }
    }

    task "app" {
      driver = "docker"

      config {
        image = "hashicorp/demo-webapp:v3"
        ports = ["http"]
      }

      env {
        PORT = "${NOMAD_PORT_http}"
      }
      resources {
        cpu        = 64
        memory     = 32
        memory_max = 64
      }
    }
    scaling {
      enabled = true
      min     = 1
      max     = 10
    }
  }
}
```

### Nomad Autoscaler logs
```
2025-07-10T03:45:43.723Z [INFO]  agent: starting Nomad Autoscaler agent
2025-07-10T03:45:43.723Z [INFO]  agent: Nomad Autoscaler agent configuration:
2025-07-10T03:45:43.723Z [INFO]  agent
2025-07-10T03:45:43.723Z [INFO]  agent:         Bind Addrs: 0.0.0.0
2025-07-10T03:45:43.723Z [INFO]  agent:  High Availability: true
[...]

PR #1125, opened 2025-07-11
recover from validation errors in policy monitoring

tgross:
The `monitorPolicies` method of the manager can receive Nomad API scaling policies with a nil inner policy (for example, if the policy is intended not to be used for autoscaling). In this case validation of the policy fails but the monitor is continuing as though it hadn't because it's not continuing to the next ID when hitting a recoverable error.

Fixes: https://github.com/hashicorp/nomad-autoscaler/issues/1124
Ref: https://hashicorp.atlassian.net/browse/NMD-890

### DIFF
--- a/policy/nomad/validate.go
+++ b/policy/nomad/validate.go
@@ -28,7 +28,7 @@ func validateScalingPolicy(policy *api.ScalingPolicy) error {
 	var result *multierror.Error
 
 	if policy == nil {
-		return multierror.Append(result, errors.New("ScalingPolicy c"))
+		return multierror.Append(result, errors.New("ScalingPolicy is empty, this policy won't execute any verification or scaling and should have enabled set to false"))
 	}
 
 	// Validate ID.


Juanadelacuesta (dismissed):

policy/manager.go
```diff
@@ -160,75 +161,87 @@ func (m *Manager) monitorPolicies(ctx context.Context, evalCh chan<- *sdk.Scalin
...
-					m.log.Trace("sending updated policy to existing handler",
-						"policy_id", policyID, "policy_source", message.Source)
+		s := m.policySources[message.Source]
+		updatedPolicy, err = s.GetLatestVersion(ctx, policyID)
+		if err != nil {
+			m.log.Error("encountered an error getting the latest version for policy", "policyID", policyID, "error", err)
+			if isUnrecoverableError(err) {
+				return fmt.Errorf("failed to get latest version for policy %s: %w", policyID, err)
+			}
+			continue
```
tgross, line 190: We're preventing the panic, but isn't this going to repeatedly log `encountered an error getting the latest version for policy` every time we have a policy ID for a policy without a [`policy`](https://developer.hashicorp.com/nomad/docs/job-specification/scaling#policy) map? Should we really be reporting that as an error at all?

Juanadelacuesta: This error should not be happening on regular basis, and might cause some other things to fail, we need to show it somehow. A policy that is enabled should not be empty

Juanadelacuesta: Or what am I not seeing?

tgross: The issue described here https://github.com/hashicorp/nomad-autoscaler/issues/1124 was with a `scaling` block like the following:

```hcl
    scaling {
      enabled = true
      min     = 1
      max     = 10
    }
```

That's a scaling block without a `policy` map, which is valid for scaling without the Nomad Autoscaler and you'll get this error message (as the did the person who reported the issue). If there's no `policy` field should we always treat it as `enabled = false`?

Juanadelacuesta: ok, so this is a different thing not directly related to the panic.

Juanadelacuesta: After running the exact example of the ticket, I still believe we need to show the error, because Nomad does not verify the job policy on submission, this is the only place were the policy is validated.

tgross: > After running the exact example of the ticket, I still believe we need to show the error, because Nomad does not verify the job policy on submission, this is the only place were the policy is validated.

Well it's a valid policy, just not one the Autoscaler should be looking at. But I get your point here. Do we want to just tell users "hey if you _don't_ have a `policy` field then you should leave `enabled = false`" (the default)? That seems pretty reasonable to me, and we can then land this fix and a little docs adjustment to close this out.

Juanadelacuesta (approved):

policy/nomad/validate.go
```diff
@@ -28,7 +28,7 @@ func validateScalingPolicy(policy *api.ScalingPolicy) error {
 	var result *multierror.Error
 
 	if policy == nil {
-		return multierror.Append(result, errors.New("ScalingPolicy is nil"))
+		return multierror.Append(result, errors.New("ScalingPolicy c"))
```
tgross: I think we meant to finish with some other value here?

PR #1134, opened 2025-07-22
Limit concurrent operations

Juanadelacuesta:
This PR introduces a new mechanism to limit the number of concurrent operations, getting rid of the old eval broker/eval worker system and leaving the policy evaluations and executions to the policy handler.

For more information, refer to the [RFC](https://go.hashi.co/rfc/nmd-217)

Ref https://hashicorp.atlassian.net/browse/NMD-839

### DIFF
--- a/policy/manager.go
+++ b/policy/manager.go
@@ -319,8 +319,8 @@ func (m *Manager) periodicMetricsReporter(ctx context.Context, interval time.Dur
 			h := m.getHandlersNum()
 
 			metrics.SetGauge([]string{"policy", "total_num"}, float32(h))
-			metrics.SetGauge([]string{"queue", "horizontal"}, float32(m.Limiter.QueueSize("horizontal")))
-			metrics.SetGauge([]string{"queue", "cluster"}, float32(m.Limiter.QueueSize("cluster")))
+			metrics.SetGauge([]string{"policy", "queue", "horizontal"}, float32(m.Limiter.QueueSize("horizontal")))
+			metrics.SetGauge([]string{"policy", "queue", "cluster"}, float32(m.Limiter.QueueSize("cluster")))
 		}
 	}
 }


plugins/manager/manager.go
```diff
@@ -371,7 +368,11 @@ func (pm *PluginManager) GetTarget(target *sdk.ScalingPolicyTarget) (targetpkg.T
 	return targetInst, nil
 }
 
-func (pm *PluginManager) GetAPM(source string) (apm.APM, error) {
+func (pm *PluginManager) GetAPMLooker(source string) (apm.Looker, error) {
+	return pm.GetAPM(source)
+}
```
tgross, line 373: What's this second method for? It seems entirely redundant with `GetAPM` because they have the same signatures. The `dependencyGetter` interface could just use `GetAPM` I think.

agent/agent.go
```diff
@@ -75,64 +69,29 @@ func (a *Agent) Run(ctx context.Context) error {
...
-		a.logger.ResetNamed("policy_eval"),
-		a.config.PolicyEval.AckTimeout,
-		a.config.PolicyEval.DeliveryLimit)
-	a.initWorkers(ctx)
-
 	a.initEnt(ctx, a.entReload)
 
 	// Launch the eval handler.
-	go a.runEvalHandler(ctx, policyEvalCh)
+	//	go a.runEvalHandler(ctx, policyEvalCh, limiter)
```
tgross: Leftover commented-out code

plugins/manager/manager.go
```diff
@@ -384,6 +385,10 @@ func (pm *PluginManager) GetAPM(source string) (apm.APM, error) {
 	return apmInst, nil
 }
 
+func (pm *PluginManager) GetStrategyRunner(name string) (strategy.Runner, error) {
+	return pm.GetStrategy(name)
+}
+
```
tgross, line 391: There's only one caller of `GetStrategy` and it's this new `GetStrategyRunner`. I get that you're breaking out the two interfaces so that we don't have to mock up the plugin interface in tests. But do we actually need this indirection to ensure that `base.Base` is implemented?

If there's no other way to do it, I'd suggest we at least make the unused `Get*` functions unexported.

policy/handler.go
```diff
@@ -5,21 +5,61 @@ package policy
...
+//     │                         │                    │
+//     │         timeout error   │  scaling error     │
+//     └─────────────────────────┴────────────────────┘
+//
+
+type handlerState int
+
 const (
-	cooldownIgnoreTime = 3 * time.Minute
+	StateScaling = iota
```
tgross: Expresses the intent a little better:

```suggestion
	StateScaling handlerState = iota
```

policy/handler.go
```diff
@@ -28,160 +68,283 @@ type Handler struct {
...
+	// A nil status indicates the target doesn't exist or is not ready, log and
+	// continue
+	if currentStatus == nil {
+		h.log.Warn("target not found", "target", h.policy.Target.Config)
+	}
+
+	lastEventTS, err := checkForOutOfBandEvents(currentStatus)
+	if err != nil {
+		return nil, fmt.Errorf("unable to get out of band event: %w", err)
+	}
```
tgross: If we get an error from `runTargetStatus` we send on the error channel and don't return, which means we hit `checkForOutOfBandEvents` line with `currentStatus == nil` and we'll panic in the first line when we call `currentStatus.Meta`.

Should we be returning an error instead? And/or should we be returning early if `currentStatus == nil` below?

policy/handler.go
```diff
@@ -28,160 +68,283 @@ type Handler struct {
...
 	h.ticker = time.NewTicker(h.policy.EvaluationInterval)
 	defer h.ticker.Stop()
 
 	h.log.Trace("starting policy handler")
+
 	for {
 		select {
 		case <-ctx.Done():
 			h.log.Error("stopping policy handler due to context done")
+			h.errChn <- ctx.Err()
```
tgross: This will send `context.Canceled` when the parent context is canceled during shutdown, won't it? And it looks like we're logging everything we send on this channel with the `encountered an error monitoring policy IDs` message (even nil errors!)

policy/handler.go
```diff
@@ -28,160 +68,283 @@ type Handler struct {
...
 			h.updateHandler(updatedPolicy)
 
 		case <-h.ticker.C:
-			eval, err := h.handleTick(ctx, h.policy)
+
+			currentStatus, err := h.runTargetStatus()
 			if err != nil {
-				h.errChn <- fmt.Errorf("handler: unable to process policy %v", err)
+				h.errChn <- fmt.Errorf("handler: failed to get target status for target: %s, %w",
+					h.policy.Target.Name, err)
```
tgross, line 222: Should we `continue` here?

policy/handler.go
```diff
@@ -199,49 +362,245 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
...
-// enforceCooldown blocks until the cooldown period has been reached, or the
-// handler has been instructed to exit. The boolean return details whether or
-// not the cooldown period passed without being interrupted.
-func (h *Handler) enforceCooldown(ctx context.Context, t time.Duration) bool {
+// applyMutators applies the mutators registered with the handler in order and
+// log any modification that was performed.
+func (h *Handler) applyMutators(p *sdk.ScalingPolicy) {
+	for _, m := range h.mutators {
+		for _, mutation := range m.MutatePolicy(p) {
+			h.log.Info("policy modified", "modification", mutation)
```
tgross, line 391: Nitpick: logging these at info seems a little noisy?

policy/handler.go
```diff
@@ -199,49 +362,245 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
...
+		})
+	}
+
+	winner := pickWinnerActionFromGroups(checkGroups)
+	if winner.handler == nil || winner.action == nil || winner.action.Direction == sdk.ScaleDirectionNone {
+		return sdk.ScalingAction{}, nil
+	}
+
+	h.log.Debug(fmt.Sprintf("check %s selected", winner.handler.check.Name),
+		"direction", winner.action.Direction, "count", winner.action.Count)
```
tgross: This means we don't have to allocate for `fmt` until we actually log (ex. if you're logging at info) and lets you filter structured logs better:

```suggestion
	h.log.Debug("check selected", "name", winner.handler.check.Name,
		"direction", winner.action.Direction, "count", winner.action.Count)
```

policy/handler.go
```diff
@@ -28,160 +68,283 @@ type Handler struct {
...
+					time.Until(h.OutOfCooldownOn()))
+
+			case StateWaitingTurn:
+				h.log.Debug("updating action, waiting to execute")
+
+			case StateScaling:
+				h.log.Info("skipping scaling, target still scaling")
+
+			case StateIdle:
+				go func() {
```
tgross, line 280: We might want to think about moving the `h.UpdateState(StateWaitingTurn)` call out of the `WaitAndScale` method and before starting this goroutine. Consider the following scenario:

* ticker fires
* this `select` receives on `<-h.ticker.C` 
* `runTargetStatus` takes a long time, enough that the evaluation period passes
* ticker fires
* we start the `WaitAndScale` goroutine
* we go back to the top of the loop
* this `select` receives on `<-h.ticker.C` 
* we start another `WaitAndScale` goroutine
* the first `WaitAndScale` gets scheduled and grabs the lock

This is incredibly unlikely because the second `runTargetStatus` would have to return almost immediately to hit this race, but it's much like calling `sync.WaitGroup.Add(1)` before starting a goroutine.

policy/handler.go
```diff
@@ -28,160 +68,283 @@ type Handler struct {
...
+			if !scalingNeeded(action, currentCount) {
+				h.log.Info("skipping scaling, no action needed")
+				continue
+			}
+
+			h.UpdateNextAction(action)
+
+			if nowFunc().After(h.OutOfCooldownOn()) {
+				h.UpdateState(StateIdle)
+			}
```
tgross: I can't find where we're resetting the `outOfCooldownOn` field. Should we make sure we're also checking that we're in `StateCooldown` here and/or clearing `outOfCooldownOn` once we have?

policy/handler.go
```diff
@@ -199,49 +362,245 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
...
+	if a.Direction == sdk.ScaleDirectionUp {
+		return p.CooldownOnScaleUp
+	}
+
+	return p.Cooldown
+}
+
+// pickWinnerActionFromGroups decide which action wins in the group.
+// The decision processes still picks the safest choice, but it handles `none`
+//
```
tgross, line 508:
```suggestion
```

tgross (commented):
Overall this looks great @Juanadelacuesta!

plugins/apm/apm.go
```diff
@@ -17,6 +17,10 @@ type APM interface {
 	// interface.
 	base.Base
 
+	Looker
+}
+
+type Looker interface {
```
jrasell, line 23: Can we add a comment here to explain what the interface does and potentially why its separate?

agent/agent.go
```diff
@@ -164,13 +123,14 @@ func (a *Agent) setupPolicyManager() (chan *sdk.ScalingEvaluation, error) {
 	// TODO: Once full policy source reload is implemented this should probably
 	// be just a warning.
 	if len(sources) == 0 {
-		return nil, fmt.Errorf("no policy source available")
+		return fmt.Errorf("no policy source available")
```
jrasell: nit: the problem was in the original code too, but we can use "errors.New" rather than the formatter.

plugins/manager/manager.go
jrasell: I think the same question could also apply to the `GetTargetController` and `GetTarget` functions.

plugins/target/target.go
```diff
@@ -17,21 +17,16 @@ type Target interface {
...
 // is in a position to be scaled as well as the current running count which
 // will be used when performing the strategy calculation.
-type TargetStatusGetter interface {
-	Status(config map[string]string) (*sdk.TargetStatus, error)
-}
-
 // Scale triggers a scaling action against the remote target as specified
 // by the config func argument.
-type TargetScaler interface {
+type TargetController interface {
```
jrasell: The name stutters, so we can probably remove the `Target` prefix and the comment is now incorrect.

policy/check.go
```diff
@@ -0,0 +1,161 @@
...
+			if ch.policy.OnCheckError == sdk.ScalingPolicyOnErrorFail {
+				return sdk.ScalingAction{}, err
+			}
+		}
+	}
+
+	return a, nil
+}
+
+func (ch *checkRunner) runStrategy(ctx context.Context, currentCount int64, ms sdk.TimestampedMetrics) (sdk.ScalingAction, error) {
```
jrasell, line 86: The context isn't used here; should we remove it from the function signature or be using it?

policy/check.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	if runResp == nil {
+		return sdk.ScalingAction{}, nil
+	}
+
+	return *runResp.Action, nil
+}
+
+// QueryMetricsy wraps the apm.Query call to provide operational functionality.
```
jrasell:
```suggestion
// QueryMetrics wraps the apm.Query call to provide operational functionality.
```

policy/check.go
```diff
@@ -0,0 +1,161 @@
...
+	defer metrics.MeasureSinceWithLabels([]string{"plugin", "apm", "query", "invoke_ms"},
+		time.Now(), labels)
+
+	// Calculate query range from the query window defined in the ch.check.
+	to := nowFunc().Add(-ch.check.QueryWindowOffset)
+	from := to.Add(-ch.check.QueryWindow)
+	r := sdk.TimeRange{From: from, To: to}
+
+	var err error
+	metrics := sdk.TimestampedMetrics{}
```
jrasell: nit: Variable 'metrics' collides with imported package name

plugins/manager/manager.go
Juanadelacuesta: That is a mistake I made, I changed the return value for both functions, the second one `GetAPM` should return `apm.APM`, not just the looker. Let's fix this

policy/manager.go
```diff
@@ -341,3 +342,42 @@ func isUnrecoverableError(err error) bool {
 	}
 	return false
 }
+
+type Limiter struct {
```
jrasell, line 348: Curious if we looked at https://pkg.go.dev/golang.org/x/time/rate for the implementation? If so, might be worth adding some comments and why we chose this approach.

plugins/manager/manager.go
Juanadelacuesta: The plugin manager uses the full `Strategy` interface. I wanted to split functionality and configuration by creating the 2 interfaces:
https://github.com/hashicorp/nomad-autoscaler/blob/a0f5dda1bf1bfb7c8a2f3fce28a8e5baf3aca13a/plugins/shared/util.go#L213

policy/check.go
```diff
@@ -0,0 +1,161 @@
...
+	Log hclog.Logger
+
+	// StrategyRunner is the strategy runner for the check.
+	StrategyRunner strategy.Runner
+
+	// MetricsGetter is the metrics getter for the check.
+	MetricsGetter apm.Looker
+
+	Policy *sdk.ScalingPolicy
+}
```
jrasell, line 32: It seems like we only set this configuration within the same package, so we could make it private. I think the same could apply to all the exported functions.

policy/handler.go
```diff
@@ -199,49 +362,244 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
...
+	defer h.cooldownLock.RUnlock()
+
+	return h.outOfCooldownOn
+}
+
+func (h *Handler) UpdateOutOfCooldownOn(ooc time.Time) {
+	h.cooldownLock.RLock()
+	defer h.cooldownLock.RUnlock()
+
+	h.outOfCooldownOn = ooc
```
jrasell: This is writing, so should we be using a write lock and not a read lock?

policy/handler.go
Juanadelacuesta: I was thinking of times when the target takes a little to be up and running.  Better to check if the `status.Meta` is not nil? If we return at starting the handler, we wont track the policy and we wont retry to start the handler until there is apolicy change and we pick up the policy again, which might end up in really weird behaviour. I rather keep it up and throwing errors, ay least it is visible.

policy/handler.go
```diff
@@ -199,49 +362,244 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
...
+			continue
+		}
+
+		winner = winner.preempt(groupWinner)
+	}
+
+	return winner
+}
+
+func (h *Handler) State() handlerState {
```
jrasell: This function and below all seems to be only used within this file/package, so do we need/want them to be exported?

plugins/manager/manager.go
Juanadelacuesta: Im trying to change as little as possible, but I would 100% move the interfaces where they are used and get rid of `strategy.Strategy`, `apm.APM` and `target.Target`

policy/handler.go
Juanadelacuesta: Good point, lets just log locally and return.

policy/handler.go
Juanadelacuesta: yes we should

policy/handler.go
Juanadelacuesta: The `OutOfCooldownOn` is updated once the scaling is successful: https://github.com/hashicorp/nomad-autoscaler/blob/0941fb2b135a1f5fddc04801ccdc5cbad7e04f34/policy/handler.go#L323

policy/handler.go
Juanadelacuesta: Makes sense

policy/handler.go
Juanadelacuesta: Not sure, this is old code I moved around but Im not really sure of its usage

plugins/target/target.go
Juanadelacuesta: lets fix the comments

plugins/apm/apm.go
Juanadelacuesta: I separated the interfaces because they were needed by different things, the part that implemented the plugin "scaffolding" and the part that did the actual plugin functionality. It was a "mix of concerns" and in practical terms, when using mocks it would force you to mock `PluginInfo() (*PluginInfo, error)` if you wanted to use `strategy.Run`, useless and confusing. This division was the way I found to keep most of the code of what is not directly related to what Im changing untouched.  It is also a little bit of a philosophical decision: I like to have each package to be "self contained" and only "informed" of what it needs to know. The fact that the strategy, target and apm are targets makes 0 difference for the policy handler, why bring an interface with plugin details if not necessary? Like I mentioned, if I was just writing from scratch, I would have placed the scaffolding interface where is needed and the implementation interface in the policy handler, but every line changed is a possibility to introduce problems and the plugins are working so I dont want to touch them :S

policy/check.go
Juanadelacuesta: We should be using it!

policy/check.go
Juanadelacuesta: fixed!

policy/manager.go
Juanadelacuesta: That is a time limiter, that makes calculations of how many calls can be made on a period of time. It is actually used in the `rate_lmiter` package. For this case, we don't need to limit per time, we want to limit the number of concurrent executions at any given moment.

policy/handler.go
Juanadelacuesta: you are right, copy paste error :S

policy/check.go
Juanadelacuesta: If we ever separate the Checks from the handler, I would like to keep this public, for the rest of the functions, lets revisit.

policy/handler.go
tgross: Oh right. For some reason I was thinking we needed to set it to 0 at some point, but that doesn't make sense. LGTM

policy/handler.go
tgross: The change you've made doesn't address this because the `h.updateState(StateWaitingTurn)` is still happening inside the goroutine.

policy/handler.go
tgross: That works.

tgross (approved):
LGTM

pkazmierczak (approved):

PR #1140, opened 2025-08-08
Check for nil policies after an update

Juanadelacuesta:
Sometimes the Nomad policy source sends empty policies and it was causing the autoscaler to crash. This PR introduces a check for those cases.

Ref: https://github.com/hashicorp/nomad-autoscaler/issues/1124

### DIFF
--- a/policy/manager.go
+++ b/policy/manager.go
@@ -93,11 +93,12 @@ func (m *Manager) Run(ctx context.Context, evalCh chan<- *sdk.ScalingEvaluation)
 	// Start the metrics reporter.
 	go m.periodicMetricsReporter(ctx, m.metricsInterval)
 
+	// Create a separate context so we can stop the goroutine monitoring the
+	// list of policies independently from the parent context.
+	monitorCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
 	for {
-		// Create a separate context so we can stop the goroutine monitoring the
-		// list of policies independently from the parent context.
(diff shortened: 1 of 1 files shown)


policy/manager.go
```diff
@@ -192,16 +195,25 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
...
+				"policyID", policyID, "error", err)
 			if isUnrecoverableError(err) {
-				return fmt.Errorf("failed to get latest version for policy %s: %w", policyID, err)
+				return fmt.Errorf("failed to get latest version for policy %s: %w",
+					policyID, err)
 			}
 			continue
 		}
 
+		if updatedPolicy == nil {
```
tgross, line 207: It looks like the only time that `updatedPolicy == nil && err == nil` is if the context closes. Do we really want to log that as an error or should we just `continue` so we can exit on the next pass thru the loop?

Juanadelacuesta: I just run into the issue, if there is an empty policy on file

tgross: Is that an error though, rather than just being a case where there's no [`policy`](https://developer.hashicorp.com/nomad/docs/job-specification/scaling#policy) block and it's just a policy not intended for use with the autoscaler? Someone who has a policy like that (as described in the original #1124 issue) will keep getting repeated errors in their logs, won't they?

Juanadelacuesta: When the policy is updated, or reloaded

PR #1145, opened 2025-08-11
Tracking policies from different sources

Juanadelacuesta:
This PR addresses 2 big bugs present on the policy tracking:

- When updating the file policy source, the internal map that holds the policies was only being populated with the policy name and file, but the content was missing, resulting on every file policy resolving to nil.
- The policy update messages where compared to all running handlers disregarding the source, which led to updates on one source deleting handlers for the others.


tgross (approved):
LGTM!

PR #1146, opened 2025-08-12
Condition the OSS code to migrate the changes to ent

Juanadelacuesta:
In order to migrate the policy worker redesign into the ent version, some parts of the code needed to be rearranged. This PR does it, plus addressing a missing deletion clean up when stoping the handlers.

### DIFF
--- a/policy/check.go
+++ b/policy/check.go
@@ -41,7 +41,7 @@ type checkRunner struct {
 }
 
 // NewCheckHandler returns a new checkHandler instance.
-func NewCheckRunner(config *CheckRunnerConfig, c *sdk.ScalingPolicyCheck) *checkRunner {
+func newCheckRunner(config *CheckRunnerConfig, c *sdk.ScalingPolicyCheck) *checkRunner {
 	return &checkRunner{
 		log:            config.Log,
 		check:          c,
@@ -131,7 +131,7 @@ func (ch *checkRunner) runStrategy(ctx context.Context, currentCount int64, ms s
 	return *runResp.Action, nil
 }
 
-// QueryMetrics wraps the apm.Query call to provide operational functionality.
+// queryMetrics wraps the apm.Query call to provide operational functionality.
 func (ch *checkRunner) queryMetrics(ctx context.Context) (sdk.TimestampedMetrics, error) {
 	ch.log.Debug("querying source", "query", ch.check.Query, "source", ch.check.Source)
 
@@ -177,11 +177,11 @@ func (ch *checkRunner) queryMetrics(ctx context.Context) (sdk.TimestampedMetrics
 	return ms, nil
 }
 
-func (ch *checkRunner) Group() string {
+func (ch *checkRunner) group() string {
 	return ch.check.Group
 }
 
-func (ch *checkRunner) RunCheckAndCapCount(ctx context.Context, currentCount int64) (sdk.ScalingAction, error) {
+func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int64) (sdk.ScalingAction, error) {
 	ch.log.Debug("received policy check for evaluation")
 
 	metrics, err := ch.queryMetrics(ctx)
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -61,8 +61,8 @@ type limiter interface {
 }
 
 type checker interface {
-	RunCheckAndCapCount(ctx context.Context, currentCount int64) (sdk.ScalingAction, error)
-	Group() string
+	runCheckAndCapCount(ctx context.Context, currentCount int64) (sdk.ScalingAction, error)
+	group() string
 }
 
 // Handler monitors a policy for changes and controls when them are sent for
@@ -107,27 +107,26 @@ type Handler struct {
 }
 
 type HandlerConfig struct {
-	UpdatesChan         chan *sdk.ScalingPolicy
-	ErrChan             chan<- error
-	Policy              *sdk.ScalingPolicy
-	Log                 hclog.Logger
-	TargetController    targetpkg.Controller
-	Limiter             *Limiter
-	DependencyGetter    dependencyGetter
+	UpdatesChan      chan *sdk.ScalingPolicy
+	ErrChan          chan<- error
+	Policy           *sdk.ScalingPolicy
+	Log              hclog.Logger
+	TargetController targetpkg.Controller
+	Limiter          *Limiter
+	DependencyGetter dependencyGetter
+
+	// Ent only field
 	HistoricalAPMGetter HistoricalAPMGetter
 	EvaluateAfter       time.Duration
 }
 
 func NewPolicyHandler(config HandlerConfig) (*Handler, error) {
-
 	h := &Handler{
 		log: config.Log,
 		mutators: []Mutator{
 			NomadAPMMutator{},
 		},
 		pm:                  config.DependencyGetter,
-		historicalAPMGetter: config.HistoricalAPMGetter,
-		evaluateAfter:       config.EvaluateAfter,
 		targetController:    config.TargetController,
 		updatesCh:           config.UpdatesChan,
 		policy:              config.Policy,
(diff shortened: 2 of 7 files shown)


policy/handler_test.go
```diff
@@ -290,7 +290,7 @@ func Test_pickWinnerActionFromGroups(t *testing.T) {
 				must.Eq(t, tt.wantAction.Direction, result.action.Direction)
 				must.Eq(t, tt.wantAction.Count, result.action.Count)
 				must.NotNil(t, result.handler)
-				must.Eq(t, tt.wantHandler.check.Name, result.handler.check.Name)
+				//must.Eq(t, tt.wantHandler.check.Name, result.handler.check.Name)
```
tgross:
```suggestion
```

agent/agent.go
```diff
@@ -189,6 +147,48 @@ func (a *Agent) reload() {
 	}
 }
 
+func (a *Agent) setupPolicyManager(limiter *policy.Limiter) error {
```
gulducat, line 150: by my eye, this is a cut-paste from above, with no changes, is that right?

sdk/policy.go
```diff
@@ -84,11 +87,15 @@ type ScalingPolicy struct {
 // Validate applies validation rules that are independent of policy source.
 func (p *ScalingPolicy) Validate() error {
 	if p == nil {
-		return nil
+		return errors.New("empty policy")
```
tgross: Just double-checking, but this looks like a behavior change? The relationship between `sdk.ScalingPolicy` and [`scaling.policy`](https://developer.hashicorp.com/nomad/docs/job-specification/scaling#policy) block has been a little fuzzy for me... but the default is for it to be nil, isn't it?

policy/check.go
```diff
@@ -51,10 +51,10 @@ func NewCheckRunner(config *CheckRunnerConfig, c *sdk.ScalingPolicyCheck) *check
 	}
 }
 
-// GetNewCountFromStrategy begins the execution of the checks and returns
+// getNewCountFromStrategy begins the execution of the checks and returns
```
gulducat, line 54: hope no one's using this or `QueryMetrics` -- they really shouldn't be, since `checkRunner` is a private type, but `NewCheckRunner` is public, so someone _could_ get their hands on a `checkRunner` and call these public methods...

maybe worth noting in a changelog, unless I'm missing some nuance about how this could never have been called by anyone.

policy/manager.go
```diff
@@ -304,21 +314,13 @@ func (m *Manager) stopHandlers() {
 	for source, handlers := range m.handlers {
 		for id := range handlers {
 			m.stopHandler(source, id)
+			delete(handlers, id)
```
tgross, line 317: This looks like a behavior change as well?

policy/manager.go
```diff
@@ -332,6 +334,30 @@ func (m *Manager) stopHandler(source SourceName, id PolicyID) {
...
+	defer m.handlersLock.Unlock()
+
+	for source, handlers := range m.handlers {
+		for id, tracker := range handlers {
+			if slices.Contains(typesToStop, tracker.policyType) {
+				m.stopHandler(source, id)
+				delete(handlers, id)
+			}
+		}
+	}
```
tgross, line 348: This doesn't include `delete(m.handlers, source)` call that `stopHandlers` does. Should we remove the source if `len(handlers) == 0` after stopping by type?

policy/check.go
```diff
@@ -176,3 +176,26 @@ func (ch *checkRunner) QueryMetrics(ctx context.Context) (sdk.TimestampedMetrics
 
 	return ms, nil
 }
+
+func (ch *checkRunner) Group() string {
+	return ch.check.Group
+}
+
+func (ch *checkRunner) RunCheckAndCapCount(ctx context.Context, currentCount int64) (sdk.ScalingAction, error) {
```
gulducat: why are these new ones public, if they're only called within this package?  are they called from somewhere else over in ent-land?

policy/handler.go
```diff
@@ -114,19 +125,21 @@ func NewPolicyHandler(config HandlerConfig) (*Handler, error) {
...
-		targetController: config.TargetController,
-		updatesCh:        config.UpdatesChan,
-		policy:           config.Policy,
-		errChn:           config.ErrChan,
-		limiter:          config.Limiter,
-		stateLock:        sync.RWMutex{},
-		state:            StateIdle,
+		pm:                  config.DependencyGetter,
+		historicalAPMGetter: config.HistoricalAPMGetter,
+		evaluateAfter:       config.EvaluateAfter,
```
gulducat: nit: can the field ordering be made consistent with the other structs? the other two have these ent-only things at the bottom.

might be nice to comment/doc them as ent-only in the `HandlerConfig`, too, since it's the public-est of these things.

policy/handler.go
```diff
@@ -114,19 +125,21 @@ func NewPolicyHandler(config HandlerConfig) (*Handler, error) {
...
+		errChn:              config.ErrChan,
+		limiter:             config.Limiter,
+		stateLock:           sync.RWMutex{},
+		state:               StateIdle,
 	}
 
 	err := h.loadCheckRunners()
 	if err != nil {
-		return nil, fmt.Errorf("failed to load check handlers: %w", err)
+		return nil, fmt.Errorf("unable to load the checks for the handler: %w", err)
```
gulducat: how important is this to change?  I always wonder what sorts of alerts folks might have in their log aggregators on certain key strings, and this is changed enough to affect monitoring systems.

policy/handler.go
```diff
@@ -161,30 +174,63 @@ func NewPolicyHandler(config HandlerConfig) (*Handler, error) {
 	return h, nil
 }
 
+// Convert the last event string.
+func checkForOutOfBandEvents(status *sdk.TargetStatus) (int64, error) {
```
gulducat: another func cut/pated verbatim from later in the same file, right?

policy/handler.go
```diff
@@ -411,34 +437,27 @@ func (h *Handler) calculateNewCount(ctx context.Context, currentCount int64) (sd
...
 	winner := pickWinnerActionFromGroups(checkGroups)
-	if winner.handler == nil || winner.action == nil || winner.action.Direction == sdk.ScaleDirectionNone {
+	if winner.handler == nil || winner.action == nil {
 		return sdk.ScalingAction{}, nil
 	}
 
-	h.log.Debug("check selected", "name", winner.handler.check.Name,
-		"direction", winner.action.Direction, "count", winner.action.Count)
+	h.log.Debug("check selected", "direction", winner.action.Direction,
+		"count", winner.action.Count)
```
gulducat, line 462: why drop `name` from this log?

agent/agent.go
Juanadelacuesta: yes it is, for a moment I thought I had to change it

policy/check.go
Juanadelacuesta: Maybe `NewCheckRunner` should be private too, but these are new types, so no legacy to worry about

policy/check.go
Juanadelacuesta: they are not, I was planning on making the Checkers into its own package, but decided not to in the end, lets make them private.

policy/manager.go
Juanadelacuesta: I realised it was missing, I will fix the comments

policy/handler.go
Juanadelacuesta: I dont remember, I think it was too long and made the logs too verbose

policy/manager.go
Juanadelacuesta: This function is meant to replace the one that would stop all  the vertical workers that where in charge of the recommendations in case of an invalid license, Im not usre removing the source is needed here.

sdk/policy.go
Juanadelacuesta: you are right, let's revert this one

PR #1161, opened 2025-09-08
feat(gce-mig): configurable `retry_attempts` and parse `cooldown_on_scale_up`

ashley-graham:
**TL;DR User Story**
As an autoscaler user with GCP infrastructure, I would like to have options for managing post-hoc event confirmation and cooldown policies so that I can: reduce errors when attempts are too quickly made to scale and confirm scaling during a longer-lasting (10 minute graceful shutdown) scale-in event.

**More Details**
In this PR, I am bringing the gce-mig target plugin into [feature parity with the aws-asg target plugin](https://github.com/hashicorp/nomad-autoscaler/pull/594/files), specifically on the point of a configurable `retry_attempts`. This is being proposed [due to recurring "retry limit reached" errors](https://discuss.hashicorp.com/t/configurable-retry-options-for-gce-mig-autoscaler-plugin/76445) encountered when a scale-in event takes longer than the hard-coded 2.5 minutes in the existing constants.

I am also updating the `cooldown` and `cooldown_on_scale_up` distinction [made available here](https://github.com/hashicorp/nomad-autoscaler/pull/1116). We wanted to use this feature to further limit the possibility of trying to accomplish another scaling event when a cluster is in transition/unstable from the prior scaling event. I noticed that when adding `cooldown_on_scale_up` to policy files for the MIGs, nothing was parsing the policy. And therefore, it was setting the `cooldown` value for everything.

If this is approved, [I'd be happy to open a PR to update the relevant documentation.](https://github.com/hashicorp/nomad/blob/main/website/content/tools/autoscaling/plugins/target/gce-mig.mdx)

**Steps to Test Changes End-To-End**

- [x] 1. Create a docker image of the build on my local fork of the repo, expecting the platform of our Nomad clients. (`docker build --platform linux/amd64 --target=dev -t simplifi/nomad-autoscaler-dev:ashley-test-4 .`)
- [x] 2. Then run docker push `simplifi/nomad-autoscaler-dev:ashley-test-4`
- [x] 3. Confirm [as expected here in docker hub](https://hub.docker.com/repository/docker/simplifi/nomad-autoscaler-dev/general)
- [x] 4. Change the autoscaler image in our autoscaler.nomad job variables
```
nomad_autoscaler_image = "simplifi/nomad-autoscaler-dev:ashley-test-4"
count = 3
evaluate_after = "5m"
log_level = "DEBUG"
```
- [x] 5. See if there are any breaking changes before further testing the specific retry_attempts functionality
- [x] 6. Update our Terraform module for Nomad client migs so that any scaled up client has a programmatically set value in our Consul KV migs policy files.
- [x] 7. Only point my sandbox dev cluster to using this changed Terraform module; create the cluster with`retry_attempts = "1"` in policy file so that it is sure to trigger errors. (Also ensure evaluation interval is set lower than the time it might take to scale in or out).
- [x] 8. Change the `autoscaler_min` and `autoscaler_max` on my cluster terragrunt.hcl to force scaling policy re-evaluation and scaling events.
- [x] 9. Repeat for all clusters across the development environment if unable to get any errors with sandbox cluster.
- [x] 10. Deploy autoscaler image in production as well to get a greater load test (in the event the plugin errors occur not just when in the middle of a deleting node event, but also when gce api is under load). In this initial test, I will _not_ change the retry attempts (will keep at default of 15).
- [x] 11. Upon getting first retry limit errors, update the retry_attempts for those clusters. 
- [x] Update underlying terraform module to use `cooldown="10min"` and `cooldown_on_scale_up="3min"` so that we can also further limit the possibility of trying to do a scaling event too soon when scaling in (but not have too long a cooldown after scaling up/out).
- [x] Confirm different `cooldown` and `cooldown_on_scale_up` used where applicable.

**Outcome (WIP)**
- In sandbox cluster and in all other clusters in development: we no longer received "reached retry limit" errors as we had (even in dev environment under lighter load, fewer scaling events) before deploying the changes in this PR.
- In sandbox cluster and in all other clusters in development: different values used for `cooldown` vs `cooldown_on_scale_up` where differences are stated.
[cooldown_distinctions_dev.csv](https://github.com/user-attachments/files/22749025/cooldown_distinctions_dev.csv)
- In production, backwards compatibility demonstration: my image change did not cause any errors, aside from the expected error of "reached retry limit" errors since I kept the default to 15. Sample logs are attached from the first few minutes of deployment. There were no errors since then other than "reached retry limit" as expected.
[test_scaling_in_1m_evaluation_interval_15_retry_atttempts_prd.csv](https://github.com/user-attachments/files/22388298/test_scaling_in_1m_evaluation_interval_15_retry_atttempts_prd.csv)
- In production, demonstration of impact when increasing retry_attempts to 60 in two regions only (given the max window for our cooldown is 10minutes after a scaling in event/graceful shutdown of MIG instances); and making distinction between `cooldown` and `cooldown_on_scale_up`: drastic reduction in errors! 🎉  (deployment happened on October 6 around 11:45). 
[datadog_prd_logs_2_regions_60_retry_attempts.csv](https://github.com/user-attachments/files/22749523/datadog_prd_logs_2_regions_60_retry_attempts.csv)
- In production, demonstrated impact for now parsing `cooldown_on_scale_up` from policy file to have different cooldowns for MIGs after scale-in vs scale-out (also called scale up). 
[datadog_prd_logs_different_and_same_cooldowns.csv](https://github.com/user-attachments/files/22750894/datadog_prd_logs_different_and_same_cooldowns.csv)


plugins/builtin/target/gce-mig/plugin/gce_test.go
```diff
@@ -68,3 +71,74 @@ func Test_gceNodeIDMap(t *testing.T) {
...
+					return false, 0, nil
+				}
+				return true, 10, nil
+			},
+		}
+
+		err := tp.ensureInstanceGroupIsStable(context.Background(), mockIG)
+
+		assert.NoError(t, err, "expected no error when MIG becomes stable")
+		assert.Equal(t, 3, attempts, "expected 3 attempts to become stable")
```
gulducat: I realize you're following convention in this file, and there's no way you could know this, but we've been slowly moving our various repos to use [github.com/shoenig/test](https://github.com/shoenig/test) rather than `testify`.

Fortunately these will be real easy to switch over!

```suggestion
		test.NoError(t, err, test.Sprint("expected no error when MIG becomes stable"))
		test.Eq(t, 3, attempts, test.Sprint("expected 3 attempts to become stable"))
```

plugins/builtin/target/gce-mig/plugin/plugin.go
```diff
@@ -21,11 +22,18 @@ const (
...
 	pluginName = "gce-mig"
 
+	// configKeys represents the known configuration parameters required at
+	// varying points throughout the plugins lifecycle.	
 	configKeyCredentials = "credentials"
 	configKeyProject     = "project"
 	configKeyRegion      = "region"
 	configKeyZone        = "zone"
 	configKeyMIGName     = "mig_name"
+	configKeyRetryAttempts  = "retry_attempts"
```
gulducat: We should probably have a CI check for this, but this bit made me realize that `go fmt` must not have been run.  It still compiles, but Go has its formatting opinions built into the toolchain (this avoids a whole class of "well _I_ like to format it _this way_" arguments, which of course engineers love to spend time arguing)

For example, these would be indented to line up like this:


```suggestion
	// configKeys represents the known configuration parameters required at
	// varying points throughout the plugins lifecycle.	
	configKeyCredentials    = "credentials"
	configKeyProject        = "project"
	configKeyRegion         = "region"
	configKeyZone           = "zone"
	configKeyMIGName        = "mig_name"
	configKeyRetryAttempts  = "retry_attempts"
```

But since it's built into the toolchain, you don't have to do this manually.  You can run this to auto-format the whole repo, which in this case should only catch your files:

```
go fmt ./...
```

Most IDE Go plugins do this for you, but for some it's a toggle to auto-fmt.

plugins/builtin/target/gce-mig/plugin/gce.go
```diff
@@ -19,7 +19,6 @@ import (
 
 const (
 	defaultRetryInterval = 10 * time.Second
-	defaultRetryLimit    = 15
```
gulducat, line 22: It would be nice for the retry-related values to remain defined next to each other, since together they determine the total duration.

plugins/builtin/target/gce-mig/plugin/plugin.go
```diff
@@ -56,18 +70,20 @@ type TargetPlugin struct {
...
 
 // SetConfig satisfies the SetConfig function on the base.Base interface.
 func (t *TargetPlugin) SetConfig(config map[string]string) error {
-
 	t.config = config
 
-	if err := t.setupGCEClients(config); err != nil {
-		return err
+	if err := t.setupGCEClientsFunc(config); err != nil {
+        return err
```
gulducat: I'm not sure whether a `TargetPlugin` that wasn't instantiated with `NewGCEMIGPlugin()` was semi-functional before this, but as you may have noticed when writing the test(?), this will cause a `nil` panic unless instantiated correctly.

We could also (or instead) set it here, if it's not already set, rather than only in the constructor.


```suggestion
	if t.setupGCEClientsFunc == nil {
		t.setupGCEClientsFunc = t.setupGCEClients
	}
	if err := t.setupGCEClientsFunc(config); err != nil {
        return err
```

This way it can't be `nil` (at least not here) and defaults to the live-code behavior.

In general, when possible, it's nice to be able to pick up a struct without running its constructor and not have it bite you when you try to use it 😋

plugins/builtin/target/gce-mig/plugin/plugin.go
```diff
@@ -79,6 +95,12 @@ func (t *TargetPlugin) SetConfig(config map[string]string) error {
 	t.clusterUtils = clusterUtils
 	t.clusterUtils.ClusterNodeIDLookupFunc = gceNodeIDMap
 
+	retryLimit, err := strconv.Atoi(getConfigValue(config, configKeyRetryAttempts, configValueRetryAttemptsDefault))
+	if err != nil {
+		return err
```
gulducat: Does this bare error look sensible to a user if they mistakenly provide not-a-number?  Could they easily find their mistake and correct it?

plugins/builtin/target/gce-mig/plugin/gce.go
ashley-graham: First, love your Github name. Second, thanks for engaging with a Golang novice in a kind way. I had moved this out of the gce.go because I saw a [similar approach on the aws-asg PR](https://github.com/hashicorp/nomad-autoscaler/pull/594#discussion_r941380291). Would you like me to still keep this here?

gulducat: Ah, that tracks. Personally I would either move `defaultRetryInterval` over next to `configValueRetryAttemptsDefault`, and/or add a comment above `defaultRetryInterval` briefly explaining their paired behavior.  I just like to avoid needing to switch files to do some mental math to understand what the max default total duration might be.

To be clear though, I wouldn't block on this point. I'm just a little fussy about stuff like this.

gulducat: Also, I'm just pleased as punch that you appreciate my very-professional username :smile:

plugins/builtin/target/gce-mig/plugin/plugin.go
gulducat: I'll add that I wouldn't bend over backwards to _always_ make a bare struct behave itself; it's not a _rule_, just a guideline that I like to shoot for.

plugins/builtin/target/gce-mig/plugin/gce.go
ashley-graham: Deep Space Nine and cats for the win!

ashley-graham: Also, I like your approach to keep them together or add a comment. I nearly made `defaultRetryInterval` configurable, too. But I wasn't sure if people would want that given the `retryAttempts` alone would help them have a lever to reduce the ERROR logs. Thanks for helping to make my PR more reasonable to others!

plugins/builtin/target/gce-mig/plugin/plugin.go
ashley-graham: This is great feedback and helpful to me as I start my journey with Golang. Thank you!

plugins/builtin/target/gce-mig/plugin/gce_test.go
gulducat: Reopening this thread to point out that the rest of the `assert`s being added here should be changed to `test` too <3

plugins/builtin/target/gce-mig/plugin/plugin.go
ashley-graham: Great questions... 😅 I had borrowed it from the aws-asg PR when second-guessing what I had done in my first commit. I'll change this to give better error handling.

plugins/builtin/target/gce-mig/plugin/gce_test.go
ashley-graham: Roger that. Working through it now...

ashley-graham: I believe I've updated all tests now. I've run them locally to confirm.

plugins/builtin/target/gce-mig/plugin/gce_test.go
```diff
@@ -63,8 +66,85 @@ func Test_gceNodeIDMap(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			actualID, actualErr := gceNodeIDMap(tc.inputNode)
-			assert.Equal(t, tc.expectedOutputID, actualID, tc.name)
-			assert.Equal(t, tc.expectedOutputError, actualErr, tc.name)
+			test.Eq(t, tc.expectedOutputID, actualID, test.Sprint("IDs should be equal"))
+
+			if tc.expectedOutputError != nil {
+				test.Error(t, actualErr, test.Sprint("expected an error"))
+				test.Eq(t, tc.expectedOutputError.Error(), actualErr.Error(), test.Sprint("errors should be equal"))
```
gulducat: There is an [`EqError`](https://pkg.go.dev/github.com/shoenig/test#EqError) func that can simplify this a bit for ya.

```suggestion
				test.EqError(t, tc.expectedOutputError, actualErr.Error())
```

and a bunch of the `shoenig/test` methods have very sensible default failure outputs, so we often can get away without an extra custom message.

One thing I often like to do is cause the test to fail on purpose, look at its failure output, and ask myself, "if I hit this error unexpectedly, is it easy to figure out what I did wrong to break this test, or what I need to do to fix it?" and here "errors should be equal" is really too vague to be helpful.

plugins/builtin/target/gce-mig/plugin/gce_test.go
```diff
@@ -63,8 +66,85 @@ func Test_gceNodeIDMap(t *testing.T) {
...
+			name: "test-mig-failure",
+			statusFunc: func(ctx context.Context, s *compute.Service) (bool, int64, error) {
+				attempts++
+				return false, 0, nil
+			},
+		}
+
+		err := tp.ensureInstanceGroupIsStable(context.Background(), mockIG)
+
+		test.Error(t, err, test.Sprint("expected an error when MIG does not become stable"))
```
gulducat: I almost always prefer to check for a specific error, either with `test.ErrorIs` (if I can check for the specific named error var) or with `test.EqError` or `test.ErrorContains` for string matching, as above.

Here the extra Sprint message is informative for the context its in, so it's a welcome addition.

plugins/builtin/target/gce-mig/plugin/plugin_test.go
```diff
@@ -45,8 +47,47 @@ func TestTargetPlugin_calculateDirection(t *testing.T) {
...
+		test.Eq(t, customAttempts, mockPlugin.retryAttempts, test.Sprint("expected custom retry attempts"))
+	})
+
+	// Test case for an invalid retry attempts value (non-integer).
+	t.Run("invalid value returns an error", func(t *testing.T) {
+		invalidConfig := map[string]string{
+			configKeyRetryAttempts: "not-a-number",
+		}
+		err := mockPlugin.SetConfig(invalidConfig)
+		test.Error(t, err, test.Sprint("expected an error for invalid retry attempts"))
```
gulducat: ditto here it's nice to check for a specific error, beyond just that there is one.

plugins/builtin/target/gce-mig/plugin/retry_test.go
```diff
@@ -39,14 +39,14 @@ func Test_retry(t *testing.T) {
...
-			name:           "function never successful and reaches retry limit",
+			name:           "function never successful and reaches retry attempts limit",
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			actualOutput := retry(tc.inputContext, tc.inputInterval, tc.inputRetry, tc.inputFunc)
-			assert.Equal(t, tc.expectedOutput, actualOutput, tc.name)
+			test.Eq(t, tc.expectedOutput, actualOutput, test.Sprint(tc.name))
```
gulducat: I'm just being real nit-picky :P

```suggestion
			test.Eq(t, tc.expectedOutput, actualOutput)
```

plugins/builtin/target/gce-mig/plugin/plugin_test.go
```diff
@@ -45,8 +47,47 @@ func TestTargetPlugin_calculateDirection(t *testing.T) {
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			actualNum, actualString := tp.calculateDirection(tc.inputMigTarget, tc.inputStrategyDesired)
-			assert.Equal(t, tc.expectedOutputNum, actualNum, tc.name)
-			assert.Equal(t, tc.expectedOutputString, actualString, tc.name)
+			test.Eq(t, tc.expectedOutputNum, actualNum, test.Sprint(tc.name))
+			test.Eq(t, tc.expectedOutputString, actualString, test.Sprint(tc.name))
```
gulducat: these extra Sprints are redundant with the sub-test name (as specified up in `t.Run(tc.name, ...)`)


```suggestion
			test.Eq(t, tc.expectedOutputNum, actualNum)
			test.Eq(t, tc.expectedOutputString, actualString)
```

(this is what you get for generously converting other tests in the files you're touchin' :P)

plugins/builtin/target/gce-mig/plugin/retry_test.go
ashley-graham: I'm glad you are because I get to learn. Thank you!

plugins/builtin/target/gce-mig/plugin/gce_test.go
ashley-graham: Great point about the `ErrorContains`. In this case, I'm pretty familiar with the error message string since it is the one I'm trying to reduce with this PR. 😅  I'll adjust. Thank you!

plugins/builtin/target/gce-mig/plugin/gce_test.go
ashley-graham: Ooh! Much better with `test.EqError`! Thank you!"

gulducat:
Thanks bunches!  The code looks good to me, so let me know how the testing goes on your end (including if you have trouble getting it set up), and we can get this merged in :slightly_smiling_face:

gulducat:
Mornin' and thanks for your extra work on this!

I think I would be satisfied with maybe an easier case for you to reproduce without needing to induce extra load:

* Set the value _lower_ than the default, something that's easier to hit the retry limit
  * the time it took should be the expected (lower) amount
  * this confirms that the number is for sure being used
* Unset (back to default) or higher should be successful, as you intend!

If you wanted to add a log somewhere that says "retrying up to {X times} every {10 seconds}" when it starts, then that would be welcome, too. :)

plugins/builtin/target/gce-mig/plugin/retry.go
```diff
@@ -49,6 +51,8 @@ func retry(ctx context.Context, retryInterval time.Duration, retryAttempts int,
 			return nil
 		}
 
+		logger.Debug("retrying...", "retry_count", retryCount+1, "retry_attempts", retryAttempts, "error", err.Error())
```
ashley-graham: I can take out this debugging log if desired; but I wanted to share what I did to ensure that my logs were accurately counting up to the 30 retry attempts every 10 or so seconds. I think that many logs would annoy folk. 😅

ashley-graham: Before I turn off this logging, I wanted to provide the example in my sandbox MIG to show it working.
[datadog_dev_logs_scale_out_60_retry_attempts.csv](https://github.com/user-attachments/files/22728484/datadog_dev_logs_scale_out_60_retry_attempts.csv)

ashley-graham: I went ahead and removed here: https://github.com/hashicorp/nomad-autoscaler/pull/1161/commits/32b6def5c6ec86681350c477f019b7cb31537dd4

gulducat:
Heya @ashley-graham, I'm back from my week off! Please let me know if there's anything I can help out with, and/or when you're ready for a re-review! :slightly_smiling_face:

plugins/builtin/target/gce-mig/plugin/gce.go
```diff
@@ -131,7 +128,16 @@ func (t *TargetPlugin) scaleIn(ctx context.Context, group instanceGroup, num int
...
 
-func (t *TargetPlugin) ensureInstanceGroupIsStable(ctx context.Context, group instanceGroup) error {
+func (t *TargetPlugin) ensureInstanceGroupIsStable(ctx context.Context, group instanceGroup, config map[string]string) error {
+	// Check if policy overrides the plugin configuration for retry_attempts.
+	retryAttempts := t.retryAttempts
+	if str, ok := config[configKeyRetryAttempts]; ok {
+		attempts, err := strconv.Atoi(str)
+		if err != nil {
+			return fmt.Errorf("failed to parse %s value from policy: %w", configKeyRetryAttempts, err)
+		}
```
gulducat: I don't believe this error condition is covered by a test.

I also feel like this may be a little late in the call stack to be doing this validation, but I don't think it's necessarily worth a refactor.

if you _really want to_, then IMO what this func really wants to receive is the `retryAttempts` int, rather than receiving the entire `config` map and extracting only that one value from it. in general, I prefer to give a method only what it needs to do its job (unless it takes awkward contortions to accomplish that).

again, feel free not to change this! it works fine! but I would like to see it covered by a test, either way :)

policy/file/parse.go
```diff
@@ -49,6 +50,17 @@ func decodePolicyDoc(decodePolicy *sdk.FileDecodeScalingPolicy) error {
...
 
+	if decodePolicy.Doc.CooldownOnScaleUpHCL != "" {
+		d, err := time.ParseDuration(decodePolicy.Doc.CooldownOnScaleUpHCL)
+		if err != nil {
+			return err
+		}
+		decodePolicy.Doc.CooldownOnScaleUp = d
+		fmt.Printf("DEBUG: Parsed cooldown_on_scale_up from policy file: %s -> %v\n", decodePolicy.Doc.CooldownOnScaleUpHCL, d)
+	} else {
+		fmt.Printf("DEBUG: No cooldown_on_scale_up found in policy file\n")
```
gulducat: yeah, sadly I do think we should drop all these Prints, mainly because
* we have config options for logs
* so folks may have expectations for log level/format that integrates with their log aggregators
* and we can't easily pass our main logger in here to align with that

good to know they were useful in development, but let's delete em.

ashley-graham: 🫡 You got it.

Issue #1194, opened 2025-11-14
Autoscaler v0.4.8 submits scaling requests when from==to after min/max capping, causing deployment loops


edward-smith:
## Summary

The Nomad Autoscaler v0.4.8 submits scaling requests even when the target count equals the current count (from=X to=X) after min/max capping, triggering continuous deployment loops.

## Environment

- **Version**: v0.4.8
- **Environment**: Production (owf-dev)
- **Impact**: 100+ jobs experiencing deployment loops

## Root Cause

When the autoscaler calculates a target instance count below the configured minimum, it correctly caps the value to the minimum. However, it then submits a scaling request to Nomad even when the current count already matches the capped target value (from=2 to=2).

This causes Nomad to trigger a deployment even though no actual scaling is needed, creating an infinite loop:

1. Policy evaluates and calculates target (e.g., 1 instance based on low CPU)
2. Autoscaler caps to minimum (e.g., min=2)
3. Current count is already 2
4. **Autoscaler submits scaling request anyway** (from=2 to=2)
5. Nomad triggers deployment
6. Deployment completes
7. Next evaluation cycle begins → back to step 1

## Evidence from Logs

Hundreds of log entries showing:

```
from=2 to=2
reason="capped count from 1 to 2 to stay within limits"
reason_history:[scaling down because factor is 0.002038]
```

Followed by active deployment errors:

```
[ERROR] policy_manager.policy_handler: unable to scale target:
reason="failed to get scale target: skipping scaling group caps/web due to active deployment"
```

## Steps to Reproduce

1. Create a Nomad job with scaling stanza:
   ```hcl
   scaling {
     enabled = true
     min     = 2
     max     = 10
     
     policy {
       cooldown = "1m"
       evaluation_interval = "1m"
       
       check "avg_cpu" {
         source = "nomad-apm"
         query  = "avg_cpu"
         
         strategy "target-value" {
           target = 80
         }
       }
     }
   }
   ```

2. Run the job with very low CPU usage (< 1%)
3. Policy calculates target=1 (or 0) based on low CPU
4. Autoscaler caps to min=2
5. Current count is already 2
6. Observe autoscaler submits scaling request anyway
7. Observe continuous deployment loop in Nomad

## Expected Behavior

The autoscaler should skip scaling when the target count equals the current count, regardless of whether capping occurred:

```go
// After calculating target and capping to min/max
if target == currentCount {
    log.Info("skipping scaling, no action needed",
[...]

PR #1204, opened 2025-11-23
Fix scaling loops caused by change in `scalingNeeded` logic

waquidvp:
Fixes https://github.com/hashicorp/nomad-autoscaler/issues/1194.

Reverts the following commit https://github.com/hashicorp/nomad-autoscaler/commit/d2df3081464bcd5eee10d9991ca3a0cd456e2899 which is causing the behavior explained in the above issue.

Not 100% if this is the correct fix, especially seeing the comment in the function. It might fix the loop, but might brake something else that is relying on the current behavior.

### DIFF
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -334,10 +334,7 @@ func (h *Handler) Run(ctx context.Context) {
 }
 
 func scalingNeeded(a sdk.ScalingAction, countCount int64) bool {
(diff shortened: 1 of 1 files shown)


tgross:
I'm going to close this in lieu of https://github.com/hashicorp/nomad-autoscaler/pull/1221 (based on my notes in the issue) but I've retained your co-authorship in the PR.

PR #1220, opened 2025-12-15
remove module files for test plugins

tgross:
Having the test plugins as separate Go modules only increases toil for handling dependency updates without doing anything to improve reliability of our tests. Remove these files to have them use the root module.


gulducat (approved):
🔥

PR #1235, opened 2026-01-21
Add a log statement showing the count returned by each autoscaler check.

nathanmonteleone:
When troubleshooting more complex autoscaler policies, it can be useful to know what target count/reason each check is returning.  That information was not available in the log previously.  This changeset adds it as a Debug-level log.

no.


policy/check.go
```diff
@@ -78,6 +78,8 @@ func (ch *checkRunner) getNewCountFromStrategy(ctx context.Context, currentCount
 				return sdk.ScalingAction{}, err
 			}
 		}
+	} else {
+		ch.log.Debug("check returned count", "strategy", ch.check.Strategy.Name, "check", ch.check.Name, "count", a.Count, "reason", a.Reason)
```
tgross, line 134: Style nitpick: we don't need to guard this in the else because all code paths in the `if err != nil` return.

```suggestion
	}
	
	ch.log.Debug("check returned count", "strategy", ch.check.Strategy.Name, "check", ch.check.Name, "count", a.Count, "reason", a.Reason)
```

tgross (changes requested):
Thanks for the PR @nathanmonteleone! 

I know it's a bit of a pain but can you please sign the CLA? Because you git history is dirty with a merge from main you've got commits from dependabot in here and that might confuse the CLA bot. You might want to rebase to clean that up as well.

tgross:
@nathanmonteleone @jrpilat I'm not sure why there are two authors here, but I saw you pushed a new merge commit. You'll need to rebase and squash down to a single commit with a signed CLA.

jrpilat:
I'm sorry, this was created by mistake (I guess as part of some Github automation) while I was [merging](https://github.com/simplifi/nomad-autoscaler/pull/7) Hashicorp changes in the other direction for a fork.

Please excuse the annoyance and close.

PR #1240, opened 2026-02-05
admin: Update Copywrite version and ensure year set in config.

jrasell:
Jira: https://hashicorp.atlassian.net/browse/NMD-1236


tgross (approved):
LGTM. But looks like this is going to touch all the files? Is that correct?

PR #1248, opened 2026-03-02
feat: Add Telegraf/Influxdb (version-1.x) APM plugin for Telegraf metrics integration

hashi-divyansh:
**Description**

- Adds **Telegraf/Influx** version-1.x support as an APM plugin, allowing the autoscaler to
query metrics from Telegraf + InfluxDB deployments.
    - Supports InfluxDB 1.x HTTP/InfluxQL queries.
    - Config: address, database, optional auth.
    - Explicitly rejects v2/v3 with clear "not yet implemented" errors.
    
**Testing**
- Tested using [this](https://github.com/hashi-divyansh/Learning/tree/test-nomad-autoscaler-with-influxDB/nomad-autoscaler-setup) setup. 
- Ran InfluxDB version 1.8 in a Docker container (on a separate VM). Our autoscaler (newly built using this branch) fetched metrics from InfluxDB and scaled the application. Screenshot 2026-03-05 at 2 47 25 PM

- After sending hell lot of request using hey it scaled the webapp instance from 1 -> 10. 
Screenshot 2026-03-05 at 2 37 21 PM

[    Ticket](https://hashicorp.atlassian.net/browse/NMD-70?atlOrigin=eyJpIjoiMTIyMWFjODA3Y2ViNGQwNjk1OWUzYzM2Yjc1NzFjNmMiLCJwIjoiaiJ9)

### DIFF
--- a/plugins/strategy/client.go
+++ b/plugins/strategy/client.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package strategy
--- a/plugins/strategy/plugin.go
+++ b/plugins/strategy/plugin.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package strategy
--- a/plugins/strategy/server.go
+++ b/plugins/strategy/server.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package strategy
--- a/plugins/target/client.go
+++ b/plugins/target/client.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package target
--- a/plugins/target/plugin.go
+++ b/plugins/target/plugin.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package target
--- a/plugins/target/server.go
+++ b/plugins/target/server.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package target


.release/ci.hcl
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) HashiCorp, Inc.
+# Copyright IBM Corp. 2020, 2026
```
pkazmierczak, line 1: I'm ok with this but ideally copyright headers change would be its own PR? (also how in the world did this not get updated before already)

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -0,0 +1,338 @@
...
+	queryValues := queryURL.Query()
+	queryValues.Set("db", a.database())
+	queryValues.Set("q", q)
+	queryValues.Set("epoch", "s")
+
+	if user := strings.TrimSpace(a.config[configKeyUsername]); user != "" {
+		queryValues.Set("u", user)
+	}
+	if pass := strings.TrimSpace(a.config[configKeyPassword]); pass != "" {
+		queryValues.Set("p", pass)
```
pkazmierczak: hmm I wonder if we could auth against influxdb via token instead of passing a password as a string?

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -0,0 +1,338 @@
...
+	case 1:
+		return m[0], nil
+	default:
+		return nil, fmt.Errorf("query returned %d metric streams, only 1 is expected", len(m))
+	}
+}
+
+// QueryMultiple executes an InfluxQL query against the InfluxDB 1.x /query
+// endpoint and returns all result series as separate metric streams.
+func (a *APMPlugin) QueryMultiple(q string, r sdk.TimeRange) ([]sdk.TimestampedMetrics, error) {
```
pkazmierczak, line 182: I don't think we ever use the `sdk.TimeRange`?

pkazmierczak (commented):
Overall great, but I left some comments and questions.

plugins/builtin/apm/influxdb/plugin/plugin.go
hashi-divyansh: Yes, we're not using it. InfluxDB doesn't have time_start/time_end query params. User have to pass this range in the query ex: "SELECT mean(cpu) FROM metrics WHERE time >= now() - 10m", but for sure we have to document it. 

Where we maintain the documentation for the APM plugins integration?

.release/ci.hcl
hashi-divyansh: For me pipeline failed and asked me to update so, I did. 
The pipeline likely didn't trigger the checks for these files, maybe because nothing has changed since acquisition. There might be other files where this is missing as well. Raised a separate [PR](https://github.com/hashicorp/nomad-autoscaler/pull/1250).

plugins/builtin/apm/influxdb/plugin/plugin.go
pkazmierczak: > Where we maintain the documentation for the APM plugins integration?

I'm afraid that's the best we have: https://developer.hashicorp.com/nomad/tools/autoscaling/concepts/plugins/apm

.release/ci.hcl
pkazmierczak: it's just so weird that none of the bots caught this, no? anyways, thanks for making a separate PR, I approved 🙇‍♂️

PR #1251, opened 2026-03-10
feat: add multi-namespace support to nomad policy source

hashi-divyansh:
**Description**

The `nomad.namespace` config field and `-nomad-namespace` CLI flag now
accept multiple values, enabling the autoscaler to monitor scaling
policies across several Nomad namespaces simultaneously.

**Changes:**
- Change `config.Nomad.Namespace` from `string` to `[]string`
- Use `flaghelper.StringFlag` for repeated `-nomad-namespace` CLI flags
- Add `normalizeNamespaces()` to deduplicate, trim, and handle `*`
- In `MergeDefaultWithAgentConfig`: single namespace sets the client
  namespace directly; multiple namespaces set `"*"` so the Nomad API
  returns all policies, which are then filtered locally.
- Add `allowedNamespaces`/`multiNamespaceMode` fields to `Source` and
  filter policies by namespace in `MonitorIDs` when multiple namespaces
  are configured
- Add `SetNamespaces()` method on `Source` for SIGHUP reload path;
  call it before `ReloadSources()` in `agent.reload()`
- Update HCL fixture and all affected tests.

[Ticket](https://hashicorp.atlassian.net/browse/NMD-904?atlOrigin=eyJpIjoiNWU2ZmQ2MjI0MjlmNDhkYWE4N2MxYTc2NGI1NzNmODEiLCJwIjoiaiJ9) 
 [Issue](https://github.com/hashicorp/nomad-autoscaler/issues/65)
 
-------------------------------------------

**Testing**

This feature is tested using [this](https://github.com/hashi-divyansh/Learning/tree/test-NMD-904-add-multi-name-spacesupport-toautoscaler/nomad-autoscaler-setup) branch setup.

In the above setup I'm running web-app in different  namespaces.
Screenshot 2026-03-12 at 6 30 08 PM

Setup a LB for all web app so that if it scale out, load is distributed to the newly added instance. 
Screenshot 2026-03-12 at 6 56 19 PM

**Case-1:**  

nomad {
  address = "http://127.0.0.1:4646"
  namespace = ["ns1", "ns2"]
}
- When I send a load request (via the LB) to the web app running in ns1 and ns2, it scales out according to the defined rules. However, when I send the load request to the web app in the default namespace, it crashes and fails to scale out. This is the expected behavior, as the autoscaler must be explicitly configured to include the default namespace (e.g., namespace = ["ns1", "ns2", "default"]) to track jobs there.

**Case-2:** namespace = ["*"]
- Autoscaler is tracking jobs from all  namespaces (including default). 
 

### DIFF
--- a/agent/config/config_test.go
+++ b/agent/config/config_test.go
@@ -461,6 +461,118 @@ func TestConfig_Load(t *testing.T) {
 	assert.Equal(t, "/opt/nomad-autoscaler/plugins", cfg.PluginDir)
 }
 
+func TestConfig_Load_BackwardCompatibilityAndFormats(t *testing.T) {
+	t.Run("single file hcl scalar namespace", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.hcl")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("nomad { namespace = \"staging\" }")
+		must.NoError(t, err)
+
+		cfg, err := Load(fh.Name())
+		must.NoError(t, err)
+		must.NotNil(t, cfg.Nomad)
+		must.Eq(t, []string{"staging"}, cfg.Nomad.Namespaces)
+	})
+
+	t.Run("single file hcl list namespace", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.hcl")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("nomad { namespace = [\"staging\", \"prod\"] }")
+		must.NoError(t, err)
+
+		cfg, err := Load(fh.Name())
+		must.NoError(t, err)
+		must.NotNil(t, cfg.Nomad)
+		must.Eq(t, []string{"staging", "prod"}, cfg.Nomad.Namespaces)
+	})
+
+	t.Run("single file json scalar namespace", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.json")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("{\"nomad\":{\"namespace\":\"staging\"}}")
+		must.NoError(t, err)
+
+		cfg, err := Load(fh.Name())
+		must.NoError(t, err)
+		must.NotNil(t, cfg.Nomad)
+		must.Eq(t, []string{"staging"}, cfg.Nomad.Namespaces)
+	})
+
+	t.Run("single file json list namespace", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.json")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("{\"nomad\":{\"namespace\":[\"staging\",\"prod\"]}}")
+		must.NoError(t, err)
+
+		cfg, err := Load(fh.Name())
+		must.NoError(t, err)
+		must.NotNil(t, cfg.Nomad)
+		must.Eq(t, []string{"staging", "prod"}, cfg.Nomad.Namespaces)
+	})
+
+	t.Run("single file without nomad block", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.hcl")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("plugin_dir = \"/tmp/plugins\"")
+		must.NoError(t, err)
+
+		cfg, err := Load(fh.Name())
+		must.NoError(t, err)
+		must.Eq(t, "/tmp/plugins", cfg.PluginDir)
+	})
+
+	t.Run("single file unsupported extension rejected", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.txt")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("{\"nomad\":{\"namespace\":\"staging\"}}")
+		must.NoError(t, err)
+
+		_, err = Load(fh.Name())
+		must.Error(t, err)
+		if !strings.Contains(err.Error(), "Unsupported file format") {
+			t.Fatalf("expected unsupported format error, got: %v", err)
+		}
+	})
+
+	t.Run("single file duplicate nomad blocks rejected", func(t *testing.T) {
+		fh, err := os.CreateTemp("", "nomad-autoscaler*.hcl")
+		must.NoError(t, err)
+		defer os.Remove(fh.Name())
+
+		_, err = fh.WriteString("nomad { namespace = \"staging\" }\nnomad { namespace = \"prod\" }")
+		must.NoError(t, err)
+
+		_, err = Load(fh.Name())
+		must.Error(t, err)
+	})
+
+	t.Run("directory load json scalar namespace", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


agent/config/config.go
```diff
@@ -131,8 +132,14 @@ type Nomad struct {
...
-	// Namespace to use.
-	Namespace string `hcl:"namespace,optional"`
+	// Namespace is the list of Nomad namespaces to monitor for scaling
+	// policies. Accepts exact namespace names or "*" to monitor all
+	// namespaces. When a single namespace is provided the autoscaler
+	// restricts all Nomad API queries to that namespace. When multiple
+	// namespaces (or "*") are provided, policies are listed across all
+	// namespaces and filtered locally. Defaults to the namespace inferred
+	// from the Nomad client environment (typically "default").
+	Namespace []string `hcl:"namespace,optional"`
```
tgross: This should now be `Namespaces` (plural), which would in turn trickle down thru the rest of the code and make this a lot more understandable reviewing -- right now you can't tell at a glance whether we're passing around a string vs a slice of strings.

More critically: how does this impact backwards compatibility with existing configurations?

agent/config/config.go
```diff
@@ -689,6 +697,30 @@ func (n *Nomad) merge(b *Nomad) *Nomad {
...
+// normalizeNamespaces deduplicates and canonicalises a namespace slice.
+// If "*" appears alongside other entries the result is collapsed to ["*"],
+// empty/whitespace-only entries are dropped, and ordering is preserved
+// (de-duplicated in-place).
+func normalizeNamespaces(ns []string) []string {
+	if len(ns) == 0 {
+		return ns
+	}
+	seen := make(map[string]bool, len(ns))
+	result := ns[:0]
```
tgross: Be clear, not clever:

```suggestion
	result := make([]string)
```

policy/nomad/source.go
```diff
@@ -190,6 +227,17 @@ func (s *Source) MonitorIDs(ctx context.Context, req policy.MonitorIDsReq) {
 			return !p.Enabled
 		})
 
+		// If multi-namespace mode is active, filter to only the allowed namespaces.
+		s.namespaceMu.RLock()
+		multiMode := s.multiNamespaceMode
+		allowed := s.allowedNamespaces
+		s.namespaceMu.RUnlock()
+		if multiMode {
```
tgross: If `multiMode` is always set by there being more than one namespace in the list, why not check that directly rather than having an extra field?

Also, instead of modifying the policy slice and then doing the contains check (adding an iteration over the policies and some allocation), could we push this check down into the `ContainsFunc`?

command/agent.go
```diff
@@ -118,6 +118,8 @@ Nomad Options:
 
   -nomad-namespace=<namespace>
     The target namespace for queries and actions bound to a namespace.
+    May be specified multiple times to monitor policies across several
+    namespaces. Use '*' to monitor all namespaces.
```
tgross, line 122: We should probably document that this filtering is happening client-side. Clusters with many namespaces are going to have all policies across all namespaces queried, right?

policy/nomad/source_test.go
```diff
@@ -505,11 +542,13 @@ func TestMonitoringIDs(t *testing.T) {
 			}
 
 			testSource := Source{
-				log:               hclog.NewNullLogger(),
+				log:                hclog.NewNullLogger(),
```
tgross: Could be github misaligning the text, but ooks like a go fmt issue here

agent/config/config.go
hashi-divyansh: I rename the go struct to Namespaces but not the HCL, to support the backward compatibility.
After this implementation in HCL file we will support both syntax  
 1. namespace = "xyz"   // backward compatible 
 2. namespace = ["abc", "abc"]  
 
and internally we will maintain only []string (by parsing string -> []string)

command/agent.go
hashi-divyansh: Yes, with one clarification:

- Single namespace configured: filtering is effectively server-side via Nomad namespace scoping.
- Namespace not provided: Nomad uses default namespace resolution (default/env/client config), so this is still single-namespace behavior.
- Multiple explicit namespaces: we use wildcard namespace for the list call and then filter client-side.

I couldn’t find a Nomad API path that filters scaling policies by an explicit list of namespaces in one request. Do we have any?

agent/config/config.go
```diff
@@ -1065,7 +1068,16 @@ func policySourceConfigSetMerge(first, second []*PolicySource) []*PolicySource {
 }
 
 func parseFile(file string, cfg *Agent) error {
-	if err := hclsimple.DecodeFile(file, nil, cfg); err != nil {
+	src, err := os.ReadFile(file)
+	if err != nil {
+		return err
+	}
+
+	if strings.EqualFold(filepath.Ext(file), ".hcl") {
```
tgross: This implies that there's a code path where we're not using HCL files for configuration. Is that true? If so, how are we handling the backwards compatibility aspect for those files?  Either way, do we require the `.hcl` extension?

command/agent.go
tgross: > I couldn’t find a Nomad API path that filters scaling policies by an explicit list of namespaces in one request. Do we have any?

No. A lot of list query endpoints support [filter](https://developer.hashicorp.com/nomad/api-docs#filtering) expressions, but I just checked the code and the List Scaling Policies API isn't one of them. That'd be nice to fix in Nomad core, but we'd still need to do the client-side filtering because we can't guarantee we'd be making the queries against a new enough version of the API.

go.mod
```diff
@@ -118,7 +118,7 @@ require (
 	github.com/shopspring/decimal v1.3.1 // indirect
 	github.com/spf13/cast v1.5.0 // indirect
 	github.com/tv42/httpunix v0.0.0-20150427012821-b75d8614f926 // indirect
-	github.com/zclconf/go-cty v1.16.3 // indirect
+	github.com/zclconf/go-cty v1.16.3
```
tgross: Run `go mod tidy` to move this up to the correct block.

agent/config/config_test.go
```diff
@@ -386,6 +386,38 @@ func TestAgent_parseFile(t *testing.T) {
 	}
 	assert.Nil(t, parseFile(fh.Name(), cfg))
 	assert.Equal(t, "/opt/nomad-autoscaler/plugins", cfg.PluginDir)
+
+	// Reset the test file.
+	if err := fh.Truncate(0); err != nil {
+		t.Fatalf("err: %s", err)
+	}
```
tgross: Style nitpick: we can use `shoenig/test/must` or `testify/require` for these assertions. (On Nomad core we're gradually migrating away from `testify` to `shoenig/test`)

agent/config/config.go
hashi-divyansh: To be frank, I didn't think about other file types, so my current implementation only supports backward compatibility for .hcl files. However, I can extend this to support other types as well (e.g., .json).

I didn't see any test cases (inside config_test.go) that test JSON support for config loading.

However, in the [loadDir()](https://github.com/hashicorp/nomad-autoscaler/pull/1251/changes#diff-1308de62937b63bdfe5da80b0bf5216c4916d12ad7a40aee4843649c1ca9c02fL1209) method, we're filtering for .hcl and .json files only, which means users can provide config in JSON format as well. I'll add backward compatibility for JSON as well.

@tgross If we're only expecting config file to be .hcl or .json then inside Load() we should validate and check before passing load file to parseFile(). In case of directory we're checking and validating but in single file case this check is missing.

tgross: Yeah you'll find that pretty much anywhere we accept HCL we also accept JSON because the HCL parser can handle that. We'll want to account for backwards compatibility there as well.

> @tgross If we're only expecting config file to be .hcl or .json then inside Load() we should validate and check before passing load file to parseFile(). In case of directory we're checking and validating but in single file case this check is missing.

Sensible.

Issue #1209, opened 2025-12-05
Share Prometheus query results across checks that have the same query string and query window.


sluebbert:
Based on the output of trace logs, it looks like the auto scaler is sending an API call to Prometheus twice in quick succession for the following two checks:

```hcl
check "avg_cpu_down" {
    source = "prometheus"
    query = "avg(nomad_alloc_cpu_usage{job=\"myservice\"}) * 100"
    query_window = "1m"
    group = "avg_cpu"

    strategy "threshold" {
        upper_bound = 30
        delta = -1
    }
}

check "avg_cpu_up" {
    source = "prometheus"
    query = "avg(nomad_alloc_cpu_usage{job=\"myservice\"}) * 100"
    query_window = "1m"
    group = "avg_cpu"

    strategy "threshold" {
        lower_bound = 80
        delta = 1
    }
}
```

When the query string and query window are identical, we could reduce the burden on the system by making a single API call and sharing the results for both checks.

The query strings above are quite simple, but the impact grows as:
- The complexity of the query increases.
- We duplicate the checks above even more to cover more threshold ranges.
- We scale to 100s of jobs.

PR #1252, opened 2026-03-13
policy: reuse identical APM query results within one evalutaion

hashi-divyansh:
**Description**
- This change reduces duplicate APM queries during policy evaluation by reusing results for checks that share the same query inputs.
- The fix introduces a short-lived in-memory cache scoped to a single policy evaluation.
   -  It remains safe in HA mode, where only the leader instance evaluates policies.
   - It avoids stale data reuse across evaluation intervals.
- How it works? 
   - A cache is created at the start of calculateNewCount.
   - Each check lookup is keyed by: source, query, query_window, query_window_offset.
   - When a check queries APM: If the key is not present, the query runs normally and the result is stored. If the key is already present, the cached result is reused.

**Note:** The cache exists only for that evaluation pass and is discarded when the evaluation completes.

[Ticket](https://hashicorp.atlassian.net/browse/NMD-1095?atlOrigin=eyJpIjoiNGVmZGI2ZDUwNDlmNDYyMDljODRiNjkzODYyZTdjZWUiLCJwIjoiaiJ9)
[Issue](https://github.com/hashicorp/nomad-autoscaler/issues/1209)

### DIFF
--- a/policy/check.go
+++ b/policy/check.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package policy
--- a/policy/check_test.go
+++ b/policy/check_test.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package policy
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package policy
--- a/policy/handler_oss.go
+++ b/policy/handler_oss.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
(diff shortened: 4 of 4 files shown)


policy/check.go
```diff
@@ -40,6 +40,36 @@ type checkRunner struct {
...
+}
+
+type queryMetricsCacheEntry struct {
+	metrics sdk.TimestampedMetrics
+	err     error
+}
+
+type queryMetricsCache struct {
+	entries map[queryMetricsCacheKey]queryMetricsCacheEntry
+}
```
pkazmierczak, line 57: I think in principle this is fine, but... using `queryMetricsCacheKey` struct as a hashmap key is not optimal. For one, it's not fast, because on every lookup Go will have to compare all fields of the struct. Second, it _may_ be the case that for logically same keys the hash function output will be different, because it's hashing a struct. I'm actually not sure about that one, Go may be clever enough to know structs are the same, but the point stands—structs make poor hashmap keys. 

So why not make `queryMetricsCacheKey` a string? When you construct the key in `ch.queryMetrics`, you could just concatenate source, query etc., into a string and then use that? Or am I missing something?

hashi-divyansh: **Comparable structs** are valid and commonly used Go map keys. If two queryMetricsCacheKey values have the same field values, Go treats them as the same key, so there is no semantic risk there. The only limitation is that all struct fields must be comparable types (for example, slices and maps are not comparable). In our case all struct fields are comparable.

I agree a string key is possible, but I prefer the struct key here because it is type-safe and collision-safe by default. With a string key, we must define and maintain a canonical encoding (including separator/escaping) to avoid accidental collisions, and we typically pay per-lookup key construction cost.

PR #1256, opened 2026-03-16
Add support for instant queries for Prometheus (APM plugin)

hashi-divyansh:
**Description**

This PR adds instant-query support through a sentinel query window value.

- Use **query_window = "instant"** to evaluate a check at a single timestamp
- query_window_offset continues to control how far back from now that timestamp is
- Duration-based query_window behavior is unchanged
- Threshold strategy in instant mode requires **within_bounds_trigger = 1**

**Motivation**
Prometheus checks can return many datapoints for range queries, while some scaling decisions require a single point-in-time value. This change provides an explicit and consistent way to request instant evaluation.

**Behavior**
- Instant mode:
   - query_window = "instant"
   - query_window_offset is applied
    - Prometheus uses instant query execution
- Range mode:
   - query_window is a duration (for example 5m)
   - Prometheus uses range query execution

**Validation**
- query_window now accepts either:
   - a valid duration string, or the literal "instant"
- For threshold strategy with query_window = "instant", within_bounds_trigger must be set to 1

**Example:**     
 ```
check "cpu_5m_single_value" {
     source = "prometheus"
     query = "avg_over_time(my_cpu_metric[5m])"
     query_window = "instant"
     query_window_offset = "0s"   // optional field

     strategy "target-value" {
          target = "80"
     }
} 
```
[Ticket](https://hashicorp.atlassian.net/browse/NMD-1232?atlOrigin=eyJpIjoiZGRjMjZiZGRhMWY0NGNhZWE1MzdiMzlhMTk2OTgwMTUiLCJwIjoiaiJ9)
[Issue](https://github.com/hashicorp/nomad-autoscaler/issues/1210)

------------------------------------------------------------------

**Testing**
- Here is the exact validation error from the autoscaler logs when within_bounds_trigger was set to 3. The autoscaler will not accept a threshold strategy policy with query_instant=true unless within_bounds_trigger is set to 1.
```
Mar 20 16:00:01 server-vm-0 nomad-autoscaler[7492]: 2026-03-20T16:00:01.529+0530 [ERROR] policy_manager: encountered an error getting the latest version for policy: policyID=ce878c51-dd65-2fdd-4366-351b79ce6721
Mar 20 16:00:01 server-vm-0 nomad-autoscaler[7492]: error=
Mar 20 16:00:01 server-vm-0 nomad-autoscaler[7492]: | 2 errors occurred:
Mar 20 16:00:01 server-vm-0 nomad-autoscaler[7492]: | * policy validation failed
Mar 20 16:00:01 server-vm-0 nomad-autoscaler[7492]: | * scaling.policy.check[avg_cpu_up].strategy[threshold].within_bounds_trigger must be set to 1 when query_instant is true
```

### DIFF
--- a/policy/nomad/validate.go
+++ b/policy/nomad/validate.go
@@ -221,12 +221,6 @@ func validateCheck(c map[string]interface{}, path string, label string) error {
 		}
 	}
 
-	if removedQueryInstant, ok := c[removedKeyQueryInstant]; ok {
-		result = multierror.Append(result,
-			fmt.Errorf("%s.%s is not supported; use %s.%s = %q", path, removedKeyQueryInstant, path, keyQueryWindow, "instant"))
-		_ = removedQueryInstant
-	}
-
 	// Some strategy plugins do not require an APM
 	var strategyValidator validatorWithLabelFunc
 	if !queryOk && !sourceOk {
--- a/policy/nomad/source.go
+++ b/policy/nomad/source.go
@@ -24,20 +24,19 @@ import (
 // Keys represent the scaling policy document keys and help translate
 // the opaque object into a usable autoscaling policy.
 const (
-	keySource              = "source"
-	keyQuery               = "query"
-	keyQueryWindow         = "query_window"
-	keyQueryWindowOffset   = "query_window_offset"
-	keyEvaluationInterval  = "evaluation_interval"
-	keyOnCheckError        = "on_check_error"
-	keyOnError             = "on_error"
-	keyTarget              = "target"
-	keyChecks              = "check"
-	keyGroup               = "group"
-	keyStrategy            = "strategy"
-	keyCooldown            = "cooldown"
-	keyCooldownOnScaleUp   = "cooldown_on_scale_up"
-	removedKeyQueryInstant = "query_instant"
+	keySource             = "source"
+	keyQuery              = "query"
+	keyQueryWindow        = "query_window"
+	keyQueryWindowOffset  = "query_window_offset"
+	keyEvaluationInterval = "evaluation_interval"
+	keyOnCheckError       = "on_check_error"
+	keyOnError            = "on_error"
+	keyTarget             = "target"
+	keyChecks             = "check"
+	keyGroup              = "group"
+	keyStrategy           = "strategy"
+	keyCooldown           = "cooldown"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: sdk/policy.go [hunk 1]
```diff
@@ -258,6 +265,7 @@ type FileDecodePolicyCheckDoc struct {
 	QueryWindowHCL       string `hcl:"query_window,optional"`
 	QueryWindowOffset    time.Duration
 	QueryWindowOffsetHCL string                 `hcl:"query_window_offset,optional"`
+	QueryInstant         bool                   `hcl:"query_instant,optional"`
```
tgross: I wonder if this should really be a separate config value rather than being a sentinel value for `query_window` (ex. `query_window = "instant"`)? If I understand the plugin changes correctly, the instant query will respect the window offset (as the To value), so we'd be telling users they want to use `query_window_offset` but not `query_window`, which feels like it could be confusing.

If we used a sentinel value, we could have a configuration like this:

That is, you could set:
```hcl
query_window = "instant"
query_window_offset = "10m"
```

and get:

```go
&FileDecodePolicyCheckDoc{
	QueryWindow: 0,
	QueryWindowOffset: time.Minute * 10,
	QueryInstant: true,
}
```

or set:

```hcl
query_window = "5m"
query_window_offset = "10m"
```

and get:

```go
&FileDecodePolicyCheckDoc{
	QueryWindow: time.Minute * 5,
	QueryWindowOffset: time.Minute * 10,
	QueryInstant: false,
}
```

policy/nomad/validate.go
```diff
@@ -238,9 +248,40 @@ func validateCheck(c map[string]interface{}, path string, label string) error {
...
+	if err := validateInstantThresholdTrigger(c, path); err != nil {
+		result = multierror.Append(result, err)
+	}
+
 	return result.ErrorOrNil()
 }
 
+func validateInstantThresholdTrigger(c map[string]interface{}, path string) error {
+	queryInstant, ok := c[keyQueryInstant].(bool)
+	if !ok || !queryInstant {
```
tgross: If this isn't a bool, should it return an error rather than nil?

hashi-divyansh: This function is running after validateCheck() already validates query_window type/format via validateDurationOrInstant()

tgross (approved):
LGTM

PR #1270, opened 2026-04-08
Fix winner selection, checks outside schedule window should not participate in the winner selection process

hashi-divyansh:
**Description:**

This PR updates check participation handling so checks that should be skipped (outside schedule window or canceled evaluation) are represented with explicit sentinel errors and excluded from winner selection.

- Updated check runner behavior 
    - Added sentinel error:
       - `errCheckOutsideSchedule`
- `runCheckAndCapCount` now returns  sentinels for non-participation cases.
    - Treat sentinel errors as skip signals (do not include those checks in winner selection).
- Refactor: Centralise the schedule validation logic in sdk pkg.

[Ticket](https://hashicorp.atlassian.net/browse/NMD-1376?atlOrigin=eyJpIjoiZTY1OWJmZDU5MmM3NGE0YWFmNGJlMTg3N2FkZWM5ZDIiLCJwIjoiaiJ9)
--------------------------------
**Testing**

Example job file which can be used for testing.
```
job "scheduled-job" {
  datacenters = ["dc1"]
  type        = "service"

  group "scheduled-job" {
    # Start from zero and let autoscaler drive count changes.
    count = 0

    network {
      port "http" {
        to = 5678
      }
    }

    task "app" {
      driver = "docker"

      config {
        image = "hashicorp/http-echo:0.2.3"
        args  = ["-text=scheduled-job"]
        ports = ["http"]
      }

      resources {
        cpu    = 100
        memory = 64
      }
    }

    scaling {
      enabled = true
      min     = 0
      max     = 10

      policy {
        evaluation_interval = "15s"
        cooldown            = "30s"

        # scale case
        check "scheduled-job-scale-to-two" {
          source = "prometheus"
          query  = "vector(1)"

          schedule {
            start    = "30 10 * * *"
            duration = "1m"
          }

          strategy "fixed-value" {
            value = 2
          }
        }

       # descale case
        check "scheduled-job-descale-to-one" {
          source = "prometheus"
          query  = "vector(1)"

          schedule {
            start    = "40 10 * * *"
            duration = "1m"
          }

          strategy "fixed-value" {
            value = 1
          }
        }

        # again scale
        check "scheduled-job-scale-three" {
          source = "prometheus"
          query  = "vector(1)"

          schedule {
            start    = "36 10 * * *"
            duration = "1m"
          }

          strategy "fixed-value" {
            value = 3
          }
        }

        # Always (except schedule window)
        check "scheduled-job-always-active" {
          source = "prometheus"
          query  = "vector(1)"

          strategy "fixed-value" {
            value = 0
          }
        }
        target "nomad-target" {
          job   = "scheduled-job"
          group = "scheduled-job"
        }
      }
    }
  }
}
```


policy/check.go
```diff
@@ -253,30 +253,40 @@ func (ch *checkRunner) group() string {
...
 	}
 
 	metrics, err := ch.queryMetrics(ctx, cache)
 	if err != nil {
-		return sdk.ScalingAction{}, fmt.Errorf("failed to query source: %v", err)
+		return sdk.ScalingAction{}, false, fmt.Errorf("failed to query source: %v", err)
+	}
+	if ctx.Err() != nil {
+		return sdk.ScalingAction{}, false, nil
 	}
```
hashi-divyansh: We should check the context as well; since queryMetrics() involves an external API call, the context might be canceled and in that case we should not allow it to participate in the winner selection.

policy/check.go
```diff
@@ -253,30 +253,42 @@ func (ch *checkRunner) group() string {
...
+// runCheckAndCapCount evaluates the check and caps the returned action.
+//
+// participating reports whether the check took part in winner selection for
+// this evaluation cycle. It is false when the check is skipped or canceled,
+// such as when it falls outside its schedule window.
+func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (action sdk.ScalingAction, participating bool, err error) {
 	ch.log.Debug("received policy check for evaluation")
 
 	if err := ch.ensureCompiledSchedule(); err != nil {
-		return sdk.ScalingAction{}, fmt.Errorf("failed to compile check schedule: %w", err)
```
hashi-divyansh, line 260: Earlier, this was returning a 'no action' state, which was interpreted as a formal decision. If other checks returned 'descale,' the logic was choosing the 'no action' over the 'descale.' This was the issue: since this was outside the schedule window, it should not have participated in the decision-making process at all.

policy/check.go
```diff
@@ -253,30 +253,42 @@ func (ch *checkRunner) group() string {
...
 	if ch.schedule != nil && !ch.schedule.activeAt(nowFunc()) {
 		ch.log.Debug("skipping check, outside schedule window")
-		return sdk.ScalingAction{}, nil
+		return sdk.ScalingAction{}, false, nil
 	}
 
 	metrics, err := ch.queryMetrics(ctx, cache)
 	if err != nil {
-		return sdk.ScalingAction{}, fmt.Errorf("failed to query source: %v", err)
+		return sdk.ScalingAction{}, false, fmt.Errorf("failed to query source: %v", err)
```
jrasell: Where possible lets use the `%w` formatting to wrap the error and preserve this for callers.
```suggestion
		return sdk.ScalingAction{}, false, fmt.Errorf("failed to query source: %w", err)
```

policy/check.go
```diff
@@ -253,30 +253,42 @@ func (ch *checkRunner) group() string {
...
+		ch.log.Debug("skipping check participation, evaluation canceled after querying metrics", "error", ctx.Err())
+		return sdk.ScalingAction{}, false, nil
 	}
 
-	action, err := ch.getNewCountFromStrategy(ctx, currentCount, metrics)
+	action, err = ch.getNewCountFromStrategy(ctx, currentCount, metrics)
 	if err != nil {
-		return sdk.ScalingAction{}, fmt.Errorf("failed get count from metrics: %v", err)
-
+		return sdk.ScalingAction{}, false, fmt.Errorf("failed get count from metrics: %v", err)
```
jrasell: Where possible lets use the `%w` formatting to wrap the error and preserve this for callers.

```suggestion
		return sdk.ScalingAction{}, false, fmt.Errorf("failed get count from metrics: %w", err)
```

policy/handler_test.go
```diff
@@ -298,6 +315,44 @@ func Test_pickWinnerActionFromGroups(t *testing.T) {
 	}
 }
 
+func TestHandler_calculateNewCount_SkipsNonParticipatingChecks(t *testing.T) {
```
jrasell: I think it would be nice to expand the tests, so we confirm that if not checks are participating, we get a no action result.

policy/handler.go
```diff
@@ -61,7 +61,7 @@ type limiter interface {
 }
 
 type checker interface {
-	runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (sdk.ScalingAction, error)
+	runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (sdk.ScalingAction, bool, error)
```
jrasell: The boolean response is ambiguous to people reading the interface definition. I think it's worth adding a comment here to explain the output values and what they mean.

jrasell (commented):
Looking great @hashi-divyansh; I've added some minor inline comments that'll be good to look at before merging.

policy/handler_test.go
hashi-divyansh: Added a new test `TestHandler_calculateNewCount_NoParticipatingChecks_ReturnsNoAction` in [commit ](https://github.com/hashicorp/nomad-autoscaler/pull/1270/changes/e805db117480500ba0f31ee87ce61badd8ad006e) which will test the  case when no checks participated.

policy/check.go
```diff
@@ -253,30 +253,42 @@ func (ch *checkRunner) group() string {
 	return ch.check.Group
 }
 
-func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (sdk.ScalingAction, error) {
+// runCheckAndCapCount evaluates the check and caps the returned action.
+//
+// participating reports whether the check took part in winner selection for
+// this evaluation cycle. It is false when the check is skipped or canceled,
+// such as when it falls outside its schedule window.
+func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (action sdk.ScalingAction, participating bool, err error) {
```
mismithhisler: So the only time this actually returns a populated scaling action is when "true" is returned? Could we use something like the scale direction of None to indicate we are not doing anything here? This feels like it's bleeding scheduling logic into the main control loop.

hashi-divyansh: No. it won't solve the problem. Returning `Direction: None ` means we have evaluated and  suggesting no action. However, a policy can have multiple checks; in that case it will be missleading. 

ex:  One check suggests **descale** and another check (outside its schedule window) returns scale direction None (or NoAction) then a/c to current winner selection logic it will choose NoAction. 

Please let me know, if you think I didn't get your point 🙂.

mismithhisler: I think that is even more reason to look at where we are evaluating schedules. Right now, an empty policy basically means `Direction: None`.  In this code, when we evaluate a schedule and it is outside of our current window, we produce `Direction: None`. Because of that, we have to add another return value to this function, which is effectively bleeding schedule logic into the main loop that evaluates policies.

What if, instead of producing `Direction: None` when we evaluate a schedule, we produce no policy at all?

hashi-divyansh: > What if, instead of producing Direction: None when we evaluate a schedule, we produce no policy at all?

ok. so you mean change this method's return type to a pointer type and return nil when it is outside the schedule window right?

mismithhisler: I looked at this a bit more and have mixed thoughts here. It's perfectly valid to return a bool to indicate, `There was no error, but I've returned an empty value that is not valid`. You'll usually return this when "looking" for something. In which case, I think calling this `ok` is more idiomatic, whereas `participating` is a bit confusing.

Returning a nil pointer is another option, although you risk that someone is going to call this function in the future and do a nil check, and get a panic.

There's also examples in golang code like the sql package where you can return a specific error and check for that. Like [here](https://pkg.go.dev/errors#example-Is). Here you might return `ErrNotInSchedule = errors.New("check is not within defined schedule")`

And finally, you could have a method on the checkRunner that validates whether you should even call `run`, which might make writing tests all that much easier.
```
if !ch.isWithinSchedule { // or ch.shouldRun if there's ever other logic to add here
  continue
}
action, err := ch.run()
```

hashi-divyansh: I explored your suggestions

1. **Changing method return type to `(s*dk.ScalingAction, error)`**
The panic risk is not the main problem; because it is a private method and used at only one place, we can handle that. Bigger issue is the method `calculateNewCount()` which calls this `runCheckAndCapCount()` it's return type is `(sdk.ScalingAction, error)` so even if we return nil from `runCheckAndCapCount()`  there  we have to return an empty value again.  I think it will take us to the direction where we will endup refactoring too much code. 

2.  **Custom  error like `ErrNotInSchedule`**
Outside schedule is not an error condition. Treating it as one forces normal control flow through the error path, which makes logs, tests, and caller behavior noisier and less honest. It also does not cover the other non-participation case ex: `context canceled`.

mismithhisler: > The panic risk is not the main problem; because it is a private method and used at only one place, we can handle that

Please keep in mind, you're the only one calling this method _right now_. A community member or other team member might make a contribution to this and assume "well I checked the error, so the object probably isn't nil". So just document this in the function docs.

> Outside schedule is not an error condition

We're bikeshedding this a bit. You've run a check runner at a time when it shouldn't actually be run. Is that an error? Maybe, maybe not. It's really up to you. In golang, if you open a file that doesn't exist, you get an error, not a nil error and nil file handle.  So my point is, I think it could go either way.

hashi-divyansh: > You've run a check runner at a time when it shouldn't actually be run.

We don't have any other option; we're not maintaining any state. We have to run it, to decide whether we're in/out of the schedule window.

I changes the implementation a/c to your 2nd recommendation (custom error) and document about the sentinal error return.

policy/handler.go
```diff
@@ -488,8 +491,15 @@ func (h *Handler) calculateNewCount(ctx context.Context, currentCount int64) (sd
 	for _, ch := range h.checkRunners {
 		action, err := ch.runCheckAndCapCount(ctx, currentCount, queryCache)
 		if err != nil {
-			return sdk.ScalingAction{}, fmt.Errorf("failed get count from metrics: %v", err)
-
+			if errors.Is(err, errCheckOutsideSchedule) {
+				h.log.Debug("skipping check, outside schedule window")
+				continue
+			}
+			if errors.Is(err, errCheckEvaluationCanceled) {
```
mismithhisler: Instead of having a special error for the context being cancelled, could we just check it here?

sdk/schedule.go
```diff
@@ -51,6 +51,11 @@ func validateSchedule(s *ScalingPolicySchedule, path string) error {
 	return nil
 }
 
+// validateSchedule validates a policy/check schedule definition.
+func validateSchedule(s *ScalingPolicySchedule, path string) error {
```
mismithhisler: This is a super shallow function. By that I mean, it performs no logic and is simply a passthrough. There are use cases for these, but I'm not sure what this is providing. I think I'd rather just see a single `ValidateSchedule(s *ScalingPolicySchedule)` function.

Another thing to point out here is that we are complicating this function a good bit by adding a path parameter that is really only used for error formatting, _and_ is already known by the caller! So really if the caller gets an error, it can do the error formatting itself with the known path.

hashi-divyansh: I removed the shallow wrapper, but I’d keep `path` on the `ValidateScalingPolicySchedule(`). The function is doing field-level validation, so keeping contextual error formatting inside it avoids duplicating wrapping logic at every call site and keeps error messages consistent across callers.

I added a comment doc to `ValidateScalingPolicySchedule()` about `path `parameter.

mismithhisler: Adding contextual information to an error is the exact reason error wrapping exists in Golang.

I guess if you had some vendetta against error wrapping that would be one thing, but you are already doing it for the rest of [compileSchedule](https://github.com/hashicorp/nomad-autoscaler/blob/b-NMD-1376-checks-outside-the-schedule-window-should-not-participate-in-the-leader-selection/policy/schedule.go#L26), but just decided for cron validation to now do error formatting within this function.

I guess this leads to me another question though, which is why are we validating in the compile? Shouldn't the policy source have already validated the policy?

hashi-divyansh: Policy-source validation is the primary gate, yes. 

- Source validation: user-facing/config-time feedback.
- Compile validation: runtime safety and predictable behavior for all callers.

policy/check.go
```diff
@@ -275,14 +282,19 @@ func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int
 		var err error
 		metrics, err = ch.queryMetrics(ctx, cache)
 		if err != nil {
-			return sdk.ScalingAction{}, fmt.Errorf("failed to query source: %v", err)
+			return sdk.ScalingAction{}, fmt.Errorf("failed to query source: %w", err)
 		}
 	}
+	if ctx.Err() != nil {
```
mismithhisler: Do you think there is a better way to do this? (hint maybe look where we are checking the context chan)

policy/handler.go
```diff
@@ -488,8 +490,15 @@ func (h *Handler) calculateNewCount(ctx context.Context, currentCount int64) (sd
 	for _, ch := range h.checkRunners {
 		action, err := ch.runCheckAndCapCount(ctx, currentCount, queryCache)
 		if err != nil {
-			return sdk.ScalingAction{}, fmt.Errorf("failed get count from metrics: %v", err)
-
+			if errors.Is(err, errCheckOutsideSchedule) {
+				h.log.Debug("skipping check, outside schedule window")
+				continue
+			}
+			if errors.Is(err, context.Canceled) {
```
mismithhisler: Do we want to be checking cancelled, or deadline exceeded?

hashi-divyansh: I removed specific context error handling from this method. With the exception of `errCheckOutsideSchedule`, all errors are now returned to the caller for handling.

Previously, context errors triggered a debug log and allowed the next `checkRunner` to proceed. Now, an error will fail the entire check for the current evaluation; however, it will be retried during the next evaluation cycle.

I think this is better than the previous approach. Please let me know what you think.

policy/check.go
hashi-divyansh: I removed context error case handling from this method. 

`queryMetrics()` and `runStrategy()` will return the context error and `runCheckAndCapCount()` will return that error to the caller method. 

More details on this [comment](https://github.com/hashicorp/nomad-autoscaler/pull/1270#discussion_r3152699620)

policy/schedule.go
```diff
@@ -29,8 +28,11 @@ func compileSchedule(s *sdk.ScalingPolicySchedule) (*compiledSchedule, error) {
 		return nil, nil
 	}
 
-	if err := validateScheduleDefinition(s); err != nil {
-		return nil, err
+	// Validate at runtime compile boundary as defense-in-depth.
+	// Source validation is the primary gate, but compileSchedule may be called
+	// from non-standard paths and should never compile invalid schedules.
```
mismithhisler: If the core logic in the autoscaler has become so difficult that we need to add defense-in-depth checks because we don't really know if we've validated a policy before compiling it, then something is very very wrong.

If we want to make a comment here that we're going to follow up and fix this, I can accept that.  But I know from experience that stuff like this piles up and so I can't approve permanent "defensive" logic that's really because of  🤷.. 😄

mismithhisler: I'd also argue that this is not `defense-in-depth`, it is actually more like`shotgun parsing`.

https://langsec.org/papers/langsec-cwes-secdev2016.pdf

policy/handler.go
```diff
@@ -62,9 +62,6 @@ type limiter interface {
 }
 
 type checker interface {
-	// It returns errCheckOutsideSchedule when the check should be skipped
-	// from winner selection for this cycle. Any other error, including
-	// context errors, aborts the evaluation.
 	runCheckAndCapCount(ctx context.Context, currentCount int64, cache *queryMetricsCache) (sdk.ScalingAction, error)
```
hashi-divyansh, line 65: It was duplicated comment; [`runCheckAndCapCount()`](https://github.com/hashicorp/nomad-autoscaler/blob/cfcdf0879b45cb97e5643c645225c7a54895ac00/policy/check.go#L272) already cover.

mismithhisler (approved):
Looks great! Thanks for the good work on this!

PR #1271, opened 2026-04-08
keep scaling when nodes are draining

### LINKED ISSUE #672: keep scaling when nodes are draining

ritesh-harihar:
**Description**
This PR is a follow-up to #811.

#811 delivered the primary behavior change for this ticket: enabling the Nomad APM path to continue evaluations when nodes are draining (via the drain-ignore option).
This PR completes the ticket by hardening that behavior for production safety and improving validation coverage.

What this PR adds on top of #811:

1) Thread-safety hardening for Nomad APM config access during runtime reloads (prevents concurrent map access/race risk).
2) Defensive nil checks in node and allocation resource accounting paths (prevents panic on incomplete Nomad API payloads).
3) Tests

Why this split matters:

#811 = main functional resolution for drain-ignore behavior.
This PR = safety, resilience, and test-depth improvements required to confidently ship that behavior.

[Testing Guide](https://docs.google.com/document/d/1o3MW7QTTHNEbgJxS5TxdTGU5wGl91YTP9JKLsIc5fxo/edit?usp=sharing)

[JIRA Issue](https://hashicorp.atlassian.net/browse/NMD-571)

**Testing:**

Screenshot 2026-04-08 at 3 13 14 PM

*node_filter_ignore_drain: "false":  Getting draining error logs

Screenshot 2026-04-07 at 9 46 23 PM

*node_filter_ignore_drain: "true":  No such logs

Screenshot 2026-04-07 at 10 01 39 PM

Screenshot 2026-04-07 at 10 02 27 PM


plugins/builtin/apm/nomad/plugin/plugin.go
```diff
@@ -37,6 +38,8 @@ var (
 )
 
 type APMPlugin struct {
+	mu     sync.RWMutex
```
mismithhisler, line 41: Why do we need this mutex?

plugins/builtin/apm/nomad/plugin/plugin.go
```diff
@@ -48,18 +51,43 @@ func NewNomadPlugin(log hclog.Logger) apm.APM {
...
+}
+
+func copyConfigMap(config map[string]string) map[string]string {
+	if config == nil {
+		return nil
+	}
+
+	configCopy := make(map[string]string, len(config))
+	for k, v := range config {
+		configCopy[k] = v
```
mismithhisler, line 85: We should use the more modern, `maps.Copy` here.

CHANGELOG.md
```diff
@@ -9,7 +9,11 @@ IMPROVEMENTS:
 * agent: The `nomad.namespace` config field and `-nomad-namespace` CLI flag now accept multiple values, allowing the autoscaler to monitor scaling policies across several Nomad namespaces. Use `*` to monitor all namespaces. When a single namespace is provided the existing behaviour is preserved. [[GH-1251](https://github.com/hashicorp/nomad-autoscaler/pull/1251)]
 * build: Updated the Alpine container image used to `alpine:3.23` [[GH-1272](https://github.com/hashicorp/nomad-autoscaler/pull/1272)]
 * policy: Reuse identical APM query results within a single policy evaluation to avoid duplicate source requests for checks that share the same source, query, query window, and query window offset. [[GH-1252](https://github.com/hashicorp/nomad-autoscaler/pull/1252)] 
+<<<<<<< HEAD
```
mismithhisler, line 12: 😄

plugins/builtin/apm/nomad/plugin/plugin.go
ritesh-harihar: What i think we need a mutex here is because APMPlugin state is mutable and accessed concurrently:

1) SetConfig(...) updates both a.config (which is a map) and a.client.
2) Query paths (getPoolResources → getClientAndConfigSnapshot) read that same state while evaluations are running.
3) Without synchronization, config reload + active evaluation can race as Go maps are not safe for concurrent read/write, a.client and a.config could be observed from different “versions” mid-update.

This mutex is enabling atomic/coherent updates of {client, config} in SetConfig (write lock), consistent reads for each evaluation cycle (read lock), no concurrent map access panic.

mismithhisler: Aha, basically "it's because of reload". I think what I'm wondering is, should reload logic be happening in two separate goroutines, which exposes us to these shared memory issues in the first place?

ritesh-harihar: So in a perfectly serialized design, reload and evaluation wouldn’t overlap. But today they can overlap in practice (reload updates plugin state while workers may still be evaluating), so plugins have to treat SetConfig and query paths as concurrent. That’s why this lock exists: to prevents concurrent map read/write and ensures client + config are observed as a consistent snapshot. For now, keeping plugin-level synchronization is the safest and least disruptive approach is what i can see.

mismithhisler: I understand that's the current state. If you understand the tradeoffs, I'll approve it.

I'm not against using a mutex (which should have already existed because of the client pointer being updated), but in this case, you have to go searching for why it's needed, which I personally think is not good design. I also understand you didn't write this code in the first place, but what I'm really interested in seeing with PR's is not just feature completion, but incremental improvement on design, readability, and maintainability of the autoscaler.

ritesh-harihar: Understood, I agree that this is a local guard rather than the final architecture. We can implement a targeted fix now to address the user-facing correctness issue, and follow up with an architectural cleanup as a separate effort.

I’ll open an issue/tech-debt item to evaluate a cleaner model (e.g., serialized reload/evaluation) and document a clear plugin concurrency contract.

I’ll also make sure future PRs focus more on  improving design, readability, and overall maintainability of the autoscaler.

mismithhisler (approved):

PR #1282, opened 2026-04-22
fix: honor requested num in IdentifyScaleInNodes

### LINKED ISSUE #572: IdentifyScaleInNodes scales all nodes down
I'm using the [nomad-hcloud-autoscaler plugin](https://github.com/AndrewChubatiuk/nomad-hcloud-autoscaler/) by @AndrewChubatiuk (thank you for publishing it!). I noticed that it tries to drain & destroy all nodes on scale in.

I think this is caused by a regression in the autoscaler API:
The plugin [calls `IdentifyScaleInNodes(config, int(count))`](https://github.com/AndrewChubatiuk/nomad-hcloud-autoscaler/blob/c51851a1936cbb53e2e2d7cc48250aa4b92d6aaf/plugin/hcloud.go#L185). I looked at the [current method definition](https://github.com/hashicorp/nomad-autoscaler/blob/10c271807c4a4da1a0a3eacf2e34440f34c4177c/sdk/helper/scaleutils/cluster.go#L176) and in the latest release and the second argument called `num` is unused. Previously, it would be read by a loop that [selects `num` nodes](https://github.com/hashicorp/nomad-autoscaler/blob/702fdb51c800c4dced345bc5928bfa661debe040/sdk/helper/scaleutils/cluster.go#L147) to be destroyed from the list of live nodes, now it returns the whole array of live nodes instead.

I'm reporting this here because the method [RunPreScaleInTasks](https://github.com/hashicorp/nomad-autoscaler/blob/10c271807c4a4da1a0a3eacf2e34440f34c4177c/sdk/helper/scaleutils/cluster.go#L62) is annotated with a comment that says `COMPAT(0.4)`, I assume that you're trying to be backwards-compatible and that this was not an intentional change.

ritesh-harihar:
**Description**
 scale-in identification could return more nodes than requested, causing plugins to potentially drain/destroy too many nodes.

Problem
IdentifyScaleInNodes(cfg, num) was not enforcing the requested num consistently, which broke expected compatibility behavior for external plugins relying on this API.

Changes
- Ensure scale-in identification respects requested node count.
- Keep remote-check pre-scale-in behavior intact (identify eligible nodes, filter by remote IDs, then select).
- Add/extend regression coverage for:
      - honoring requested num
      - remote-check behavior
      - rejecting non-positive num inputs

[JIRA ISSUE](https://hashicorp.atlassian.net/browse/NMD-1357)

Testing:

<details>
- validated via live Nomad test-setup flow + focused test suite.

 - **Previous flow:**

Screenshot 2026-04-22 at 12 05 07 PM

-

Screenshot 2026-04-22 at 12 05 23 PM

-  **Current flow:**

Screenshot 2026-04-22 at 10 56 01 AM

-

Screenshot 2026-04-22 at 10 56 24 AM

-

</details>

### DIFF
--- a/sdk/helper/scaleutils/cluster.go
+++ b/sdk/helper/scaleutils/cluster.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package scaleutils
--- a/sdk/helper/scaleutils/cluster.go
+++ b/sdk/helper/scaleutils/cluster.go
@@ -27,6 +27,8 @@ type nodeDrainer interface {
 	MonitorDrain(ctx context.Context, nodeID string, index uint64, ignoreSys bool) <-chan *api.MonitorMessage
 }
 
+var errInvalidScaleInNum = errors.New("number of nodes requested for removal must be greater than zero")
+
 // ClusterScaleUtils provides common functionality when performing horizontal
 // cluster scaling evaluations and actions.
 type ClusterScaleUtils struct {
@@ -117,7 +119,7 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
 	}
 
 	if num < 1 {
-		return nil, fmt.Errorf("number of nodes requested for removal must be greater than zero")
+		return nil, errInvalidScaleInNum
 	}
 
 	// Find nodes in Nomad that match the node filtering criteria.
@@ -199,7 +201,7 @@ func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int)
 	}
 
 	if num < 1 {
-		return nil, fmt.Errorf("number of nodes requested for removal must be greater than zero")
+		return nil, errInvalidScaleInNum
 	}
 
 	// If the caller has requested more nodes than we have available once
--- a/sdk/helper/scaleutils/cluster.go
+++ b/sdk/helper/scaleutils/cluster.go
@@ -167,6 +167,7 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
 	if num > len(filteredNodes) {
 		c.log.Warn("can only identify portion of requested nodes for removal",
 			"requested", num, "available", len(filteredNodes))
+		num = len(filteredNodes)
 	}
 
 	// Select which nodes to drain and terminate based on the policy's
@@ -195,13 +196,13 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
 
 func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int) ([]*api.NodeListStub, error) {
 
+	if num < 1 {
+		return nil, errInvalidScaleInNum
+	}
+
 	filteredNodes, err := c.identifyEligibleScaleInNodes(cfg)
 	if err != nil {
 		return nil, err
-	}
-
-	if num < 1 {
-		return nil, errInvalidScaleInNum
 	}
 
 	// If the caller has requested more nodes than we have available once
--- a/sdk/helper/scaleutils/cluster.go
+++ b/sdk/helper/scaleutils/cluster.go
@@ -116,8 +116,12 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
 		return nil, errors.New("required ClusterNodeIDLookupFunc not set")
 	}
 
+	if num < 1 {
+		return nil, fmt.Errorf("number of nodes requested for removal must be greater than zero")
+	}
+
 	// Find nodes in Nomad that match the node filtering criteria.
-	nodes, err := c.IdentifyScaleInNodes(cfg, num)
+	nodes, err := c.identifyEligibleScaleInNodes(cfg)
 	if err != nil {
 		return nil, err
 	}
@@ -189,6 +193,34 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
 
 func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int) ([]*api.NodeListStub, error) {
 
+	filteredNodes, err := c.identifyEligibleScaleInNodes(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	if num < 1 {
+		return nil, fmt.Errorf("number of nodes requested for removal must be greater than zero")
+	}
+
+	// If the caller has requested more nodes than we have available once
+	// filtered, adjust the value. This shouldn't cause the whole scaling
+	// action to fail, but we should warn.
+	if num > len(filteredNodes) {
+		c.log.Warn("can only identify portion of requested nodes for removal",
+			"requested", num, "available", len(filteredNodes))
+		num = len(filteredNodes)
+	}
+
+	selectedNodes, err := c.SelectScaleInNodes(filteredNodes, cfg, num)
+	if err != nil {
+		return nil, err
+	}
+
+	return selectedNodes, nil
+}
+
+func (c *ClusterScaleUtils) identifyEligibleScaleInNodes(cfg map[string]string) ([]*api.NodeListStub, error) {
+
 	poolID, err := nodepool.NewClusterNodePoolIdentifier(cfg)
 	if err != nil {
 		return nil, err
@@ -234,14 +266,6 @@ func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int)
 	// Ensure we have not filtered out all the available nodes.
 	if len(filteredNodes) == 0 {
 		return nil, fmt.Errorf("no nodes unfiltered for %s with value %s", poolID.Key(), poolID.Value())
-	}
-
-	// If the caller has requested more nodes than we have available once
-	// filtered, adjust the value. This shouldn't cause the whole scaling
-	// action to fail, but we should warn.
-	if num > len(filteredNodes) {
-		c.log.Warn("can only identify portion of requested nodes for removal",
-			"requested", num, "available", len(filteredNodes))
 	}
 
 	return filteredNodes, nil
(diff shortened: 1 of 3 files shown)


sdk/helper/scaleutils/cluster.go
```diff
@@ -189,6 +195,34 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
...
 func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int) ([]*api.NodeListStub, error) {
 
+	filteredNodes, err := c.identifyEligibleScaleInNodes(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	if num < 1 {
+		return nil, errInvalidScaleInNum
+	}
```
gulducat: nit: this could go first, before `idenfityEligible...(cfg)` since it's a simple static comparison.

sdk/helper/scaleutils/cluster.go
```diff
@@ -189,6 +195,34 @@ func (c *ClusterScaleUtils) RunPreScaleInTasksWithRemoteCheck(ctx context.Contex
...
+
+	// If the caller has requested more nodes than we have available once
+	// filtered, adjust the value. This shouldn't cause the whole scaling
+	// action to fail, but we should warn.
+	if num > len(filteredNodes) {
+		c.log.Warn("can only identify portion of requested nodes for removal",
+			"requested", num, "available", len(filteredNodes))
+		num = len(filteredNodes)
```
gulducat, line 214: To confirm, `num = len(filteredNodes)` here is the main bugfix, yes?

Is this not also a problem up in `RunPreScaleInTasksWithRemoteCheck`? On line 168, it warns on this same condition, but then doesn't clamp num to the number of filteredNodes, and passes the possibly-too-big `num` into `c.SelectScaleInNodes()`.

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"encoding/json"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/nomad-autoscaler/sdk"
+	"github.com/hashicorp/nomad/api"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
gulducat: We tend to prefer `github.com/shoenig/test` rather than `testify` in Nomad and related repos.

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,204 @@
...
+			expectedResourceCount: 0,
+			expectedError:         "number of nodes requested for removal must be greater than zero",
+			name:                  "rejects non positive num",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			cu := newTestClusterScaleUtilsWithRemoteCheck(t)
+			ids, err := cu.RunPreScaleInTasksWithRemoteCheck(context.Background(), testScaleInCfg(), tc.inputRemoteIDs, tc.inputNum)
```
gulducat: nit:
```suggestion
			ids, err := cu.RunPreScaleInTasksWithRemoteCheck(t.Context(), testScaleInCfg(), tc.inputRemoteIDs, tc.inputNum)
```

ditto elsewhere, basically anywhere I see `context.Background()` in a test file.

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,204 @@
...
+	nodes := []*api.NodeListStub{
+		{
+			ID:                    "node-a",
+			NodeClass:             "pool-a",
+			Status:                api.NodeStatusReady,
+			SchedulingEligibility: api.NodeSchedulingEligible,
+		},
+		{
+			ID:                    "node-b",
+			NodeClass:             "pool-a",
```
gulducat, line 167: If I'm understanding the test setup correctly, these two nodes being in the same NodeClass makes it impossible to filter them separately, so we can't exercise the bugfix.

gulducat (changes requested):
The fix in general looks pretty good, but I think the same fix might be needed in `RunPreScaleInTasksWithRemoteCheck`, and I have concerns with the unit tests.

For the PR description, I almost always prefer text over screenshots, so I can ctrl+f or copy-paste.  I love that you have a test script in addition to the unit tests, but I can't actually see what it's doing. Including that in a [`<details>`](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/organizing-information-with-collapsed-sections) section is really nice to keep the description tidy while also including scripts like that.

sdk/helper/scaleutils/cluster.go
ritesh-harihar: No, num = len(filteredNodes) is just a safety clamp for the edge case where the caller requests more nodes than are available after filtering. The actual bugfix is that IdentifyScaleInNodes now calls SelectScaleInNodes (previously it returned every filtered node, completely ignoring num).

Added num = len(filteredNodes) in **RunPreScaleInTasksWithRemoteCheck.**

sdk/helper/scaleutils/cluster_scalein_test.go
ritesh-harihar: Both nodes being in the same NodeClass is needed because we need more nodes to survive the pool filter than num requests. With 2 nodes in "pool-a" and num=1:

- Without the fix: returns 2 nodes (ignores num)
- With the fix: SelectScaleInNodes limits to 1 — correct

If they were in different classes, only 1 node would pass the filter, so num=1 would trivially succeed with or without the fix.

sdk/helper/scaleutils/cluster.go
gulducat: Ah, I see now. Thanks for the correction!

I often like to leave a github PR comment on the line(s) that constitute the fix. In this case `SelectScaleInNodes` wasn't mentioned by name in the PR description, nor in the screenshots from what I can tell (again, cmd+f would be helpful if it were text), so I had room to confuse myself with this clamp. Github comments on your own PR can be really nice to save the reviewer some cognitive load.

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,203 @@
...
+			expectedNodeCount:   1,
+			expectedFirstNodeID: "node-a",
+			expectedError:       "",
+			name:                "honors requested num",
+		},
+		{
+			inputNum:            0,
+			expectedNodeCount:   0,
+			expectedFirstNodeID: "",
+			expectedError:       "number of nodes requested for removal must be greater than zero",
```
gulducat: nit: if we reuse the same `error` instead of a string, then we can be sure it's identical (not really a concern here, since the strings are already identical, despite what the func name "ErrorContains" implies) with `must.ErrorIs()`, which checks by identity rather than value.

It would also make this test show up when searching for references to the `errInvalidScaleInNum` in an IDE. I just personally find it easier to debug issues when I can cmd+click around an IDE, or otherwise chase references through the language server, rather than only `grep`.


```suggestion
			expectedError:       errInvalidScaleInNum,
```

sdk/helper/scaleutils/cluster.go
ritesh-harihar: Noted.

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+			if tc.expectedError != "" {
+				must.Error(t, err)
+				must.Nil(t, nodes)
+				must.ErrorContains(t, err, tc.expectedError)
+				return
+			}
+
+			must.NoError(t, err)
+			must.Eq(t, tc.expectedNodeCount, len(nodes))
```
gulducat: nit: some of these assertion methods are really vague, like if I induce this one to fail by setting expected=2 in the test case,

```
--- FAIL: TestClusterScaleUtils_IdentifyScaleInNodes (0.00s)
    --- FAIL: TestClusterScaleUtils_IdentifyScaleInNodes/honors_requested_num (0.00s)
        cluster_scalein_test.go:56:
            cluster_scalein_test.go:56: expected equality via cmp.Equal function
            ↪ Assertion | differential ↷
              int(
            - 	2,
            + 	1,
              )
```

Arguably the test name elucidates this mysterious integer, but some extra color is often nice for future people debugging. A different test method can help:

```suggestion
			must.Len(t, tc.expectedNodeCount, nodes)
```

highlights the type:

> cluster_scalein_test.go:56: expected slice to be different length
> ↪ len(slice): 1, expected: 2

And/or an extra annotation can make it really clear:

```suggestion
			must.Len(t, tc.expectedNodeCount, nodes, must.Sprint("unexpected node count"))
```

produces:

> cluster_scalein_test.go:56: expected slice to be different length
> ↪ len(slice): 1, expected: 2
> ↪ PostScript | annotation ↷
>    unexpected node count

sdk/helper/scaleutils/cluster_scalein_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+
+		switch r.URL.Path {
+		case "/v1/nodes":
+			must.NoError(t, json.NewEncoder(w).Encode(nodes))
+		case "/v1/node/node-a", "/v1/node/node-b":
+			n, ok := nodeInfo[r.URL.Path[len("/v1/node/"):]]
+			must.True(t, ok)
```
gulducat: Bare `must.True` (and False) can be especially annoying, IMO. Example failure:

> cluster_scalein_test.go:187: expected condition to be true; is false

You could write `StatusNotFound` like you do in the `default` condition to be more like a normal HTTP API, but this would help too:

```suggestion
			nID := r.URL.Path[len("/v1/node/"):]
			n, ok := nodeInfo[nID]
			must.True(t, ok, must.Sprintf("unknown node: %q", nID))
```

so for a bad node, we'd get:

>  cluster_scalein_test.go:187: expected condition to be true; is false
>  ↪ PostScript | annotation ↷
> unknown node: "node-x"

gulducat (dismissed):
LGTM! All I have left are my usual nit-picks about test failure legibility that can make a future debugger's day a little easier.

sdk/helper/scaleutils/cluster_scalein_test.go
gulducat: Nice, thanks for the explanation.

Some code comments to explain why the data has the shape it does might have helped me out a bit here.

gulducat (approved):
🚀

PR #1291, opened 2026-05-07
Fix: added support for datacenter and combined query

### LINKED ISSUE #599: nomad-apm: add support to group nodes by datacenter
When expanding a nomad-apm short form query, the nomad-autoscaler will [check if the target has `node_class` or `datacenter`](https://github.com/hashicorp/nomad-autoscaler/blob/main/policy/policy.go#L117-L120) set to expand the query. When expanding the query though, `node_class` is assumed and therefore makes the use of `datacenter` with the nomad-apm plugin impossible. The way in which the [node pool query](https://github.com/hashicorp/nomad-autoscaler/blob/main/plugins/builtin/apm/nomad/plugin/node.go#L169-L171) is generated will need to be changed to utilise [this helper function](https://github.com/hashicorp/nomad-autoscaler/blob/562c3192c10c8361d6dd6d5db152feaad8e25f07/sdk/helper/scaleutils/nodepool/nodepool.go#L30). These changes should be enough to enable support.

We should also consider supporting the combined node pooling functionality, however, this will require additional changes to the query expansion than described above. Whoever implements the initial datacenter support should make this choice, and raise follow up issues as needed.

ritesh-harihar:
**_Problem_**

[JIRA ISSUE](https://hashicorp.atlassian.net/browse/NMD-1394)

When writing a cluster scaling policy, operators could only filter nodes by node_class:

```
target "aws-asg" {
  node_class = "hashistack"    
  node_drain_deadline = "5m"
}
```

There was no way to filter by datacenter or node_pool. Setting either was silently ignored — the autoscaler would include all nodes regardless, producing incorrect resource totals and wrong scaling decisions.

**_Solution_**

Operators can now use datacenter and node_pool in the target block, individually or combined:

# Filter by datacenter
```
target "aws-asg" {
  datacenter = "us-east-1"
  node_drain_deadline = "5m"
}
```

# Filter by node_pool
```
target "aws-asg" {
  node_pool = "gpu"
  node_drain_deadline = "5m"
}
```
# Filter by both (AND logic — only nodes matching ALL keys)
```
target "aws-asg" {
  node_class = "hashistack"
  datacenter = "us-east-1"
  node_drain_deadline = "5m"
}
```

# Before (only node_class worked)

Policy target config | Internal query generated | Result
-- | -- | --
node_class = "hashistack" | node_percentage-allocated_memory/hashistack/class |  Correct — filters by class
datacenter = "dc1" | Ignored — no query expansion, falls through to unfiltered |  Returns ALL nodes
node_class = "hashistack" + datacenter = "dc1" | node_percentage-allocated_memory/hashistack/class |  Only filters by class, ignores datacenter
node_pool = "gpu" | Ignored | Returns ALL nodes

# After (all keys supported)

Policy target config | Internal query generated | Result
-- | -- | --
node_class = "hashistack" | node_percentage-allocated_memory/node_class=hashistack | Same behavior
datacenter = "dc1" | node_percentage-allocated_memory/datacenter=dc1 | Now filters by datacenter
node_class = "hashistack" + datacenter = "dc1" | node_percentage-allocated_memory/node_class=hashistack+datacenter=dc1 | AND logic — only nodes matching both
node_pool = "gpu" | node_percentage-allocated_memory/node_pool=gpu | Now filters by node_pool
node_class = "hashistack" + datacenter = "dc1" + node_pool = "gpu" | node_percentage-allocated_memory/datacenter=dc1+node_class=hashistack+node_pool=gpu | AND — all three must match

# Old long-form queries still work

Old format (still accepted) | Normalized to
-- | --
node_percentage-allocated_memory/hashistack/class | node_percentage-allocated_memory/node_class=hashistack
node_percentage-allocated_cpu/class=hashistack | node_percentage-allocated_cpu/node_class=hashistack

The key change: old format used /value/key, new format uses /key=value (with + for combining multiple keys).

**_Data Flow Summary:_**

```
HCL target config
  → IsNodePoolTarget()
  → NewClusterNodePoolIdentifier()
  → EncodeCombinedQueryIdentifiers()
  → CanonicalizeAPMQuery() builds "node_<op>_<metric>/key=val+key=val"
  → normalizeNodePoolQuery() handles legacy formats
  → APM plugin: DecodeCombinedQueryIdentifiers() → getPoolResources() filters nodes
```

**_Testing:_**

<details>
   1) client1 ->  cpu_total_compute = 2000, memory_total_mb   = 2048 
   2) client2 ->  cpu_total_compute = 4000, memory_total_mb   = 4096
   3)  client3 ->   cpu_total_compute = 8000, memory_total_mb   = 8192

- 

Screenshot 2026-05-07 at 12 46 58 PM
- 

Screenshot 2026-05-11 at 2 36 38 PM

- 

Screenshot 2026-05-11 at 2 37 21 PM

- 

Screenshot 2026-05-07 at 12 49 23 PM

</details>


policy/policy.go
```diff
@@ -167,12 +168,27 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+		if combined, ok := poolID.(nodepool.CombinedPoolIdentifier); ok {
+			encoded := nodepool.EncodeCombinedQueryIdentifiers(combined.Identifiers())
+			c.Query = fmt.Sprintf("%s_%s/%s",
+				nomadAPM.QueryTypeNode, c.Query, encoded)
+			return
+		}
+
+		// Single identifier: node_<op>_<metric>/<value>/<key>
+		c.Query = fmt.Sprintf("%s_%s/%s/%s",
+			nomadAPM.QueryTypeNode, c.Query, poolID.Value(), poolID.Key())
```
hashi-divyansh: `poolID.Key()` will return `node_class`  and for backward compatibility we will keep `class `as well right? 

I'd suggest using **only** `class` as the key to simplify the implementation. I noticed the code is handling both keys in several places, which seems unnecessary if we can stick to one.

policy/policy.go
```diff
@@ -167,12 +168,27 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+		}
+
+		// Check if this is a combined identifier (multiple pool keys set).
+		// Use the new combined format: node_<op>_<metric>/key1=val1+key2=val2
+		if combined, ok := poolID.(nodepool.CombinedPoolIdentifier); ok {
+			encoded := nodepool.EncodeCombinedQueryIdentifiers(combined.Identifiers())
+			c.Query = fmt.Sprintf("%s_%s/%s",
+				nomadAPM.QueryTypeNode, c.Query, encoded)
+			return
+		}
```
hashi-divyansh: We can adopt the new combined query format internally and drop the old one.

Note: The parser should still accept the old query format for backward compatibility, then transform it into the new format for internal use.

policy/policy.go
ritesh-harihar: Keeping the node_class as the canonical key in the combined format because it matches the target config key users write, aligns with Nomad's API and CLI naming, and is consistent with the other keys (datacenter, node_pool). Since the combined format is new with no existing usage, there's no legacy reason to use the shorter name.

 We accept class in the parser purely for backward compatibility with old hand-written queries like /hashistack/class

CHANGELOG.md
```diff
@@ -10,12 +10,12 @@ IMPROVEMENTS:
 * build: Updated protobuf generation tooling to use `protoc-gen-go` and `protoc-gen-go-grpc`. The generated plugin gRPC interfaces are unchanged and wire-compatible; existing external plugins do not require any changes. [[GH-1289](https://github.com/hashicorp/nomad-autoscaler/pull/1289)]
 * build: Updated the Alpine container image used to `alpine:3.23` [[GH-1272](https://github.com/hashicorp/nomad-autoscaler/pull/1272)]
 * policy: Reuse identical APM query results within a single policy evaluation to avoid duplicate source requests for checks that share the same source, query, query window, and query window offset. [[GH-1252](https://github.com/hashicorp/nomad-autoscaler/pull/1252)]
+* plugin/apm/nomad: Add support for grouping nodes by `datacenter`, `node_pool`, or a combination of multiple pool identifiers in node pool APM queries. Previously only `node_class` was supported. A new combined query format (`key1=val1+key2=val2`) enables filtering by multiple attributes simultaneously. [[GH-1291](https://github.com/hashicorp/nomad-autoscaler/pull/1291)]
```
hashi-divyansh: Information about our internal new query format is not required and also please update the **official doc** [PR ](https://github.com/hashicorp/web-unified-docs/pull/2272) about the newly added feature.

```suggestion
* plugin/apm/nomad: Add support for grouping nodes by `datacenter`, `node_pool`, or a combination of multiple pool identifiers in node pool APM queries. Previously only `node_class` was supported. [[GH-1291](https://github.com/hashicorp/nomad-autoscaler/pull/1291)]
```

policy/policy.go
hashi-divyansh: This [commit](https://github.com/hashicorp/nomad-autoscaler/pull/1291/changes/e1ea945ed77e16be45cb1cb36dad00db4c8a25cf) does not handling the point I mentioned. In the previous comment, I mentioned that parser should accept the old query and unfortunately  in this PR there is a method name `parseNodePoolQuery()`  😀 so I think you misunderstood that I'm saying `parseNodePoolQuery()` should accept the old query format. 

I’ll rephrase my point in more detail below.

hashi-divyansh: `policy.go/CanonicalizeAPMQuery()` this method is the entry point for the query, We can extend `CanonicalizeAPMQuery()` to handle 3 cases instead of just short queries:

```
Short query       → expand to new format  (existing logic)
Old long format   → transform to new format  (new logic)
New long format   → leave as-is  (already correct)
```

Then `parseNodePoolQuery()` only ever sees the new format and can be simplified, Currently the case checker logic is complex 🤨.

| Feature / Component | Before | After |
| :--- | :--- | :--- |
| `CanonicalizeAPMQuery` | short queries only | **short + old-format** long queries |
| `parseNodePoolQuery` | complex switch, 2 formats | **single path**, new format only |
| `isValidPoolKey` | needed in parser | **moved** to policy layer, **deleted** from plugin |
| Ambiguity guard | needed | **gone** |

This also **properly centralizes** all format normalization in one place — which is semantically where it belongs.

suggestion: Please add these logic in a separate new method, for better readability/maintainability.

policy/policy.go
hashi-divyansh: I like your approach of using node_class internally and dropping the class key. We don't need class key in parseNodePoolQuery(); I think we can handle this inside CanonicalizeAPMQuery() or the new method you will add. 🙂 More about the CanonicalizeAPMQuery() approach in my other comment.

hashi-divyansh (changes requested):
I added my comment regarding approach; code/test-case review I'll do it later 😊.

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -70,3 +84,61 @@ func (c *combinedClusterPoolIdentifier) Value() string {
...
+			return nil, fmt.Errorf("failed to decode value in pair %q: %v", pair, err)
+		}
+
+		switch kv[0] {
+		case "node_class", "class":
+			ids = append(ids, NewNodeClassPoolIdentifier(value))
+		case "datacenter":
+			ids = append(ids, NewNodeDatacenterPoolIdentifier(value))
+		case "node_pool":
+			ids = append(ids, NewNodePoolClusterPoolIdentifier(value))
```
gulducat: Any particular reason not to use the sdk constants here? Risk of circular dependencies?


```suggestion
		case sdk.TargetConfigKeyClass, "class":
			ids = append(ids, NewNodeClassPoolIdentifier(value))
		case sdk.TargetConfigKeyDatacenter:
			ids = append(ids, NewNodeDatacenterPoolIdentifier(value))
		case sdk.TargetConfigKeyNodePool:
			ids = append(ids, NewNodePoolClusterPoolIdentifier(value))
```

policy/policy.go
```diff
@@ -167,12 +176,74 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+		return q
+	}
+
+	// Map the old key to the canonical key name.
+	key := normalizePoolKey(parts[2])
+	if key == "" {
+		// Unrecognized key — return as-is; the parser will error.
+		return q
+	}
+
```
hashi-divyansh, line 234: **The current silent return is deceptive** — the function contract implies "I normalized your query", but in the unrecognized key case it silently hands back a query that **will 100% fail** in `parseNodePoolQuery`. You're just deferring the error with worse context.

> Fail fast, fail clearly

```
func (pr *Processor) normalizeNodePoolQuery(q string) (string, error) {
    ...
    key := normalizePoolKey(parts[2])
    if key == "" {
        return "", fmt.Errorf("unrecognized pool identifier key %q in query %q, "+
            "allowed values are: node_class, datacenter, node_pool", parts[2], q)
    }

    normalized := fmt.Sprintf("%s/%s=%s", parts[0], key, url.QueryEscape(parts[1]))
    pr.logger.Debug("normalized legacy node pool query",
        "original", q, "normalized", normalized)
    return normalized, nil
}
```

And the call site in `CanonicalizeAPMQuery`:

```
if !isShortQuery(c.Query) {
    normalized, err := pr.normalizeNodePoolQuery(c.Query)
    if err != nil {
        pr.logger.Warn("skipping query normalization", "error", err)
        return
    }
    c.Query = normalized
    return
}
```

hashi-divyansh: The key question is: **should `CanonicalizeAPMQuery` propagate the error or absorb it?**

Currently `CanonicalizeAPMQuery` has no return value — it silently skips on problems, So absorbing with a `Warn` log fits the existing pattern.

But honestly, that existing pattern is also questionable — a misconfigured query should probably surface early at canonicalization time, not bubble up as a confusing APM parse error at evaluation time. If you're open to changing the signature of `CanonicalizeAPMQuery` to return an error, that would be the most correct design end-to-end. But I'll not force you to do it since it is an architectural refactoring you can do it in follow up PR.

policy/policy.go
```diff
@@ -167,12 +176,74 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+		c.Query = fmt.Sprintf("%s_%s/%s",
+			nomadAPM.QueryTypeNode, c.Query, encoded)
+	}
+}
+
+// normalizeNodePoolQuery converts old-format node pool queries
+// (node_<op>_<metric>/<value>/<key>) into the new combined format
+// (node_<op>_<metric>/key=value). Queries already in the new format
+// or non-node-pool queries are returned unchanged.
+func normalizeNodePoolQuery(q string) string {
```
hashi-divyansh: I think it will be worth adding test-method for `normalizeNodePoolQuery()`.

policy/check.go
```diff
@@ -286,7 +286,7 @@ func (ch *checkRunner) runCheckAndCapCount(ctx context.Context, currentCount int
 		var err error
 		metrics, err = ch.queryMetrics(ctx, cache)
 		if err != nil {
-			return sdk.ScalingAction{}, fmt.Errorf("failed to query source: %w", err)
+			return sdk.ScalingAction{}, err
```
hashi-divyansh, line 289: 🤨 ?

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -70,3 +84,61 @@ func (c *combinedClusterPoolIdentifier) Value() string {
...
+		}
+
+		value, err := url.QueryUnescape(kv[1])
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode value in pair %q: %v", pair, err)
+		}
+
+		switch kv[0] {
+		case "node_class", "class":
+			ids = append(ids, NewNodeClassPoolIdentifier(value))
```
hashi-divyansh, line 61: I think `class` is not required anymore, we already replaces class key -> node_class  in the `normalizeNodePoolQuery` so w will never encounter class key here. Am I right?

suggestion: instead of using string literal ( ex: datacenter, node_lass, node_pool) all over the places, check  if there is any common constant variable which you can use.  I think we have these const var in some pkg, probably you have to export them.

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -23,6 +24,19 @@ const (
...
+// exposes the sub-identifiers and mode for serialization purposes.
+type CombinedPoolIdentifier interface {
+	ClusterNodePoolIdentifier
+
+	// Identifiers returns the list of sub-identifiers that compose this
+	// combined identifier.
+	Identifiers() []ClusterNodePoolIdentifier
+
+	// Mode returns the combination mode (and/or).
+	Mode() CombinedClusterPoolIdentifierMode
```
hashi-divyansh: `Mode() ` is not used anywhere, if not required then remove it.

hashi-divyansh (commented):
Thanks for adopting the new approach! It's looking a lot better 🙂. I've pointed out a few remaining areas where we can refine this further.

policy/check.go
ritesh-harihar: [ERROR] policy_manager: encountered an error monitoring policy IDs: error="handler: unable to execute policy ID: ed38f46d-24b2-ff42-0b71-5e9b3ee4549a, failed get count from metrics: failed to query source: failed to query source: no nodes identified within pool"

The double "failed to query source" is a wrapping issue — two layers are adding the same prefix, thats why removed the wrapping.

sdk/helper/scaleutils/nodepool/combined.go
ritesh-harihar: Removed.

sdk/helper/scaleutils/nodepool/combined.go
ritesh-harihar: 1)  Partially, the internal flow is safe because CanonicalizeAPMQuery → normalizeNodePoolQuery always converts class to node_class before decoding.

But DecodeCombinedQueryIdentifiers is public and can be called directly, bypassing normalization. A manually constructed long-form query like node_percentage-allocated_cpu/class=compute skips normalizeNodePoolQuery entirely because queries with = in the second segment are returned unchanged (if strings.Contains(parts[1], "=") { return q }).

So an edge case when user writes node_percentage-allocated_cpu/class=compute as the query. The normalizer sees class=compute already contains = and returns it unchanged. The decoder then receives class=compute. Without the alias, this would error with "unknown pool key".

For this added a comment why "class" has to be there.

2)  Added constants instead of string literals.

policy/policy.go
ritesh-harihar: Added.

policy/policy.go
ritesh-harihar: Added hclog.Logger to Processor,  normalizeNodePoolQuery is now a method and logs on successful normalization and [WARN] when it encounters an unrecognized key (before the parser error at eval time).

All 3 callers(agent.go, testing.go, source_test.go) already had a logger available; tests use hclog.NewNullLogger(). 

No signature change to CanonicalizeAPMQuery.

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -70,3 +82,62 @@ func (c *combinedClusterPoolIdentifier) Value() string {
...
+		}
+
+		switch kv[0] {
+		// "class" is accepted as a legacy alias for sdk.TargetConfigKeyClass
+		// ("node_class"). While normalizeNodePoolQuery rewrites old 3-part
+		// queries, a user can still write "class=value" directly in a
+		// long-form query (e.g. node_percentage-allocated_cpu/class=compute),
+		// which bypasses normalization because it already contains "=".
+		case sdk.TargetConfigKeyClass, "class":
+			ids = append(ids, NewNodeClassPoolIdentifier(value))
```
hashi-divyansh, line 61: But that's a **design hole in `normalizeNodePoolQuery` itself**, not a reason to put legacy logic in the decoder.
**=** symbol introduced in the new query format the old one was **/**.  so you are considering the case where user will use old key **class** with the new separator symbol **=** which is quite unlikely but I like that you are thinking about this case 👍.  

**The fix:**`normalizeNodePoolQuery` should handle this case too — scan key names in the combined format and normalize `class` → `node_class`

Example you can do something like this in `normalizeNodePoolQuer`: 
```
// If the second part contains "=", it might be new format but with
// legacy "class" key. Normalize any "class" key to "node_class".
if strings.Contains(parts[1], "=") {
    normalized := normalizeClassAlias(parts[1])
    if normalized != parts[1] {
        result := fmt.Sprintf("%s/%s", parts[0], normalized)
        pr.logger.Debug("normalized legacy class key in combined query",
            "original", q, "normalized", result)
        return result, nil
    }
    return q, nil
}

// normalizeClassAlias replaces "class=value" with "node_class=value"
// in a combined format string (key1=val1+key2=val2).
func normalizeClassAlias(s string) string {
    return strings.ReplaceAll(s, "class=", sdk.TargetConfigKeyClass+"=")
}
```

Then `DecodeCombinedQueryIdentifiers` drops the `"class"` case entirely — it only ever sees `node_class` post-normalization. The decoder stays **clean and strict**, and **all** legacy handling is in one place: `normalizeNodePoolQuery`.

policy/policy.go
```diff
@@ -6,28 +6,33 @@ package policy
...
+	logger    hclog.Logger
 	defaults  *ConfigDefaults
 	nomadAPMs []string
 }
 
 // NewProcessor returns a pointer to a new Processor for use.
-func NewProcessor(defaults *ConfigDefaults, apms []string) *Processor {
+func NewProcessor(log hclog.Logger, defaults *ConfigDefaults, apms []string) *Processor {
 	return &Processor{
+		logger:    log,
```
hashi-divyansh:
```suggestion
		log:       log.ResetNamed("policy_processor"),
```

policy/policy.go
```diff
@@ -6,28 +6,33 @@ package policy
...
 	"github.com/hashicorp/nomad-autoscaler/sdk"
+	"github.com/hashicorp/nomad-autoscaler/sdk/helper/scaleutils/nodepool"
 )
 
 const thresholdWithinBoundsTriggerConfigKey = "within_bounds_trigger"
 
 // Processor helps process policies and perform common actions on them when
 // they are discovered from their source.
 type Processor struct {
+	logger    hclog.Logger
```
hashi-divyansh: nit pick: Every other struct in the project uses log

```suggestion
	log   hclog.Logger
```

policy/policy.go
```diff
@@ -145,10 +150,23 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
-	// internal model.
+	// If the query is already in long form, normalize any old-format
+	// node pool queries to the new combined format and return.
+	// Operators can write long queries directly if they know the
+	// autoscaler internal model.
 	if !isShortQuery(c.Query) {
+		normalized, err := pr.normalizeNodePoolQuery(c.Query)
+		if err != nil {
+			pr.logger.Warn("failed to normalize node pool query, will fail at evaluation",
+				"query", c.Query, "error", err)
```
hashi-divyansh: I think Error level log make sense here, what do you think?

```suggestion
			pr.log.Error("failed to normalize node pool query, evaluation will fail",
                    "query", c.Query, "error", err)

```

ritesh-harihar: I think Warn is correct. The normalization failure doesn't crash anything,  the query passes through and the parser gives a proper error at eval time. The actual error is already logged as [ERROR] by the policy manager when evaluation fails. Warn gives early visibility without that duplication.

sdk/helper/scaleutils/nodepool/combined.go
ritesh-harihar: Moved all legacy class → node_class handling into normalizeNodePoolQuery.  Now, normalizeNodePoolQuery scans combined-format queries too — it splits on +, checks each key individually, and rewrites bare "class" to "node_class" (using pair-level splitting instead of strings.ReplaceAll to avoid corrupting node_class into node_node_class). 

All legacy rewriting are in one place,  decoder is cleaned up to only accept canonical keys (node_class, datacenter, node_pool), and tests were updated accordingly.

policy/policy.go
```diff
@@ -135,7 +140,7 @@ func (pr *Processor) CanonicalizeCheck(c *sdk.ScalingPolicyCheck, t *sdk.Scaling
 func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.ScalingPolicyTarget) {
```
ritesh-harihar, line 140: CanonicalizeAPMQuery() expands short queries into combined format using target config.

policy/policy.go
```diff
@@ -167,12 +185,108 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+			nomadAPM.QueryTypeNode, c.Query, encoded)
+	}
+}
+
+// normalizeNodePoolQuery converts old-format node pool queries
+// (node_<op>_<metric>/<value>/<key>) into the new combined format
+// (node_<op>_<metric>/key=value). Queries already in the new format
+// or non-node-pool queries are returned unchanged.
+// Returns an error if the old-format key is not recognized.
+func (pr *Processor) normalizeNodePoolQuery(q string) (string, error) {
```
ritesh-harihar, line 206: normalizeNodePoolQuery() Converts old 3-part format → new key=value format; normalizes class alias

policy/policy.go
gulducat: To elaborate on the signature question:

`Canoncalize` methods in Nomad projects do not return errors. I don't love patterns encouraged by Canonicalize methods (they often end up doing things that a `NewX()` constructor should do, and can paper over an otherwise-unsafe struct design), but the general idea behind them is to set defaults or move values around within a complex type. They may evolve over time in order to support backwards-compatibility, but they almost always just set values.

In the canoncalize world, `Validate` is a totally distinct step, and that's usually how you'd surface a configuration error (as soon as possible in the call stack).

To get a sense of its usage in Nomad, which inspired its presence in the Autoscaler, cmd+f `) Canonicalize` in nomad's excessively long [`nomad/structs/structs.go` file](https://github.com/hashicorp/nomad/blob/main/nomad/structs/structs.go).

hashi-divyansh (dismissed):

plugins/builtin/apm/nomad/plugin/node.go
```diff
@@ -180,13 +181,23 @@ func (a *APMPlugin) getNodeAllocatedResources(client *api.Client, nodeID string,
...
+
+	// Parse the combined format: key1=val1[+key2=val2...]
+	ids, err := nodepool.DecodeCombinedQueryIdentifiers(mainParts[1])
+	if err != nil {
+		return nil, err
+	}
+	if len(ids) == 1 {
+		query.poolIdentifier = ids[0]
+	} else {
+		query.poolIdentifier = nodepool.NewCombinedClusterPoolIdentifier(ids, nodepool.CombinedClusterPoolIdentifierAnd)
```
gulducat: This holds on to some assumptions that we don't have to retain.

---

First, `query.poolIdentifier` is a `nodepool.ClusterNodePoolIdentifier`, but the only method that gets called is `query.poolIdentifier.IsPoolMember` (via `a.getPoolResources()`), so really all we need is `IsPoolMember()`, either by passing in a func directly, or it could be a little interface.

It's nice to define as narrow an interface as you actually need, and near where it's used, so we could add to this package (or `nodepool` sdk, if we want to, but then we're beholden to it)

```go
type nodePoolFilter interface {
	IsPoolMember(*api.NodeListStub) bool
}
```

and change `poolIdentifier nodepool.ClusterNodePoolIdentifier` to: `poolFilter nodePoolFilter` and `query.poolIdentifier` => `query.poolFilter`, etc etc.

_If_ we cared about `Key()` or `Value()` from the `ClusterNodePoolIdentifier` interface, I'd have a different opinion, but I don't see them being used.

---

This `if/else` also distinguishes between 1 vs many underlying identifiers, which we don't really need to care about here. We just want to filter the nodes.

This part would be really easy to handle here, it could be a blanket `query.poolIdentifier = nodepool.NewCombinedClusterPoolIdentifier(ids, nodepool.CombinedClusterPoolIdentifierAnd)`

But I have more to say on this subject, in another comment.

policy/policy.go
```diff
@@ -167,12 +185,108 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
+			return
+		}
+
+		// Extract identifiers: combined has multiple, single has one.
+		var ids []nodepool.ClusterNodePoolIdentifier
+		if combined, ok := poolID.(nodepool.CombinedPoolIdentifier); ok {
+			ids = combined.Identifiers()
+		} else {
+			ids = []nodepool.ClusterNodePoolIdentifier{poolID}
+		}
```
gulducat: This is another spot where we're distinguishing between 1 and many when we don't really need to, and we added a new interface to support it.

We don't have to use `nodepool.NewClusterNodePoolIdentifier()`, and I'm not sure why _it_ distinguishes between 1 and many itself... If we use it, we have to deal with its return type, but we don't have to use it.

What if we instead always use `[]nodepool.ClusterNodePoolIdentifier`? We could make _it_ adhere to `ClusterNodePoolIdentifier` interface (or not!), rather than making a new one that embeds the other one to surface the slice/array.

e.g. in the `nodepool` package, replace the new `CombinedPoolIdentifier interface` with a new concrete type:

```go
type ClusterNodePoolIdentifierList []ClusterNodePoolIdentifier

func NewClusterNodePoolIdentifierList(cfg map[string]string) (ClusterNodePoolIdentifierList, error) {
	ids := parseNodePoolIdentifiers(cfg) // split out from NewClusterNodePoolIdentifier() to reuse
	if len(ids) == 0 {
		return nil, fmt.Errorf("node pool identification method required") // this should be an error type...
	}
	return ids, nil
}

// IsPoolMember returns true if all identifiers are members.
func (c ClusterNodePoolIdentifierList) IsPoolMember(n *api.NodeListStub) bool {
	for _, id := range c {
		if !id.IsPoolMember(n) { // produces "AND" behavior
			return false
		}
	}
	return true
}

// ... and Key() and Value() *if* we care to, which we don't necessarily need to.
// It's not clear to me what they're actually needed for; it looks to me like they're
// only used in errors/logs.
```

So then that thing just already _is_ what we need. We could also put Encode right on it.

```go
func (c ClusterNodePoolIdentifierList) Encode() string {
	// NOTE: EncodeCombinedQueryIdentifiers would not need to be exported, or even exist.
	// this Encode() method could replace it unless we have a compelling reason to keep it.
	return encodeCombinedQueryIdentifiers(c)
}
```

Your `DecodeCombinedQueryIdentifiers` method already returns `[]ClusterNodePoolIdentifier` -- it could instead return this new `ClusterNodePoolIdentifierList` (which remember actually is the same, just with extra convenience methods).

```go
func DecodeCombinedQueryIdentifiers(encoded string) (ClusterNodePoolIdentifierList, error) {
```

Sorry if this commentary is a bit scattered, but all together, imagine this stretch of code in `if t.IsNodePoolTarget()` could be something like


```go
ids, err := nodepool.NewClusterNodePoolIdentifierList(t.Config)
if err != nil {
	return // an error condition feels really out of place for a Canonicalize() method, but I digress.
}
encoded := ids.Encode()
c.Query = // ...
```

And up in `nomad/plugin/node.go`'s `parseNodePoolQuery` method, it would be almost identical:

```go
	ids, err := nodepool.DecodeCombinedQueryIdentifiers(mainParts[1])
	if err != nil {
		return nil, err
	}
	query.poolFilter = ids // this field rename intersects with my other comment
```

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -23,6 +25,16 @@ const (
 // ClusterNodePoolIdentifiers are combined.
 type CombinedClusterPoolIdentifierMode string
 
+// CombinedPoolIdentifier is an extension of ClusterNodePoolIdentifier that
+// exposes the sub-identifiers and mode for serialization purposes.
+type CombinedPoolIdentifier interface {
+	ClusterNodePoolIdentifier
```
gulducat: This smells funny to me. Embedding interfaces is a language feature, and sometimes it's appropriate, and it does work! But this feels tacked-on where instead we can treat `[]ClusterNodePoolIdentifier` as a first-class type, per my other comment, and avoid this new public interface definition.

policy/policy.go
```diff
@@ -145,10 +150,23 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
...
-	// internal model.
+	// If the query is already in long form, normalize any old-format
+	// node pool queries to the new combined format and return.
+	// Operators can write long queries directly if they know the
+	// autoscaler internal model.
 	if !isShortQuery(c.Query) {
+		normalized, err := pr.normalizeNodePoolQuery(c.Query)
+		if err != nil {
+			pr.log.Warn("failed to normalize node pool query, will fail at evaluation",
+				"query", c.Query, "error", err)
```
gulducat, line 161: This is a code smell:

> will fail at evaluation

Like we know it will be bad later, but we're helpless to prevent it? What does it look like when it inevitably fails later? What other options might we have to bubble this error up to an actual error, aside from changing `Canonicalize`'s signature?

gulducat (commented):
Thanks a bunch for updating the PR description!

And I can see that you guys have made great progress to isolate and sharpen the functionality.

I haven't looked at the tests yet, nor all of the details of the normalization, but I have a structural critique for you to consider. Let me know what you think, and I'll get to the string manipulation and tests when I can later.

policy/policy.go
hashi-divyansh: Option 1:  Remove `normalizeNodePoolQuery` from `Canonicalize `method and move it inside `ValidatePolicy` . Validate policy have return error signature.

ritesh-harihar: Option 2:  New `NormalizeChecks(policy) error `on Processor

Move `normalizeNodePoolQuery` out of `Canonicalize `into a separate method that returns error. Both file source and nomad source already have error paths where we can call it post-canonicalization.

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -1,72 +1,76 @@
...
+		}
+	}
+	return true
 }
 
-// NewCombinedClusterPoolIdentifier returns a new combinedClusterPoolIdentifier.
-func NewCombinedClusterPoolIdentifier(poolIdentifiers []ClusterNodePoolIdentifier, mode CombinedClusterPoolIdentifierMode) ClusterNodePoolIdentifier {
-	return &combinedClusterPoolIdentifier{
-		poolIdentifiers: poolIdentifiers,
-		mode:            mode,
```
gulducat, line 38: Although I prefer the `ClusterNodePoolIdentifierList` approach, we shouldn't delete this func outright, because it's an exported method in an SDK library. There may be people in the world who use it (and it's basically impossible to prove no one is).

`combinedClusterPoolIdentifier` is at least a private struct, so we don't have to keep it if we don't want to, but sadly we are beholden to the signature of exported funcs like this, including `NewClusterNodePoolIdentifier` below (and any others). If we want to ditch `combinedClusterNodePoolIdentifier` and fully replace it with `ClusterNodePoolIdentifierList`, then we could make `..List` satisfy  `ClusterNodePoolIdentifier` interface (just a couple small methods), but we'd still be left without an `CombinedClusterPoolIdentifierOr` option (which, sheesh, I really wonder if anyone actually uses, because why would they? We certainly don't...)

As unfortunate as it can be to have two very-similar solutions to the same problem, the simplest thing might be to just hold on to this one for any library user that *might* be using it.

It's probably worth deciding on some deprecation strategy, so we can warn people a couple releases in advance of deleting them, because it's not a lot of fun to maintain probably-dead code indefinitely.

OR it is always an option to be more aggressive and just leave a breaking-change changelog entry saying that we've deleted it, and hope library users notice if it's relevant to them... 🤷

plugins/builtin/apm/nomad/plugin/node.go
```diff
@@ -180,14 +187,20 @@ func (a *APMPlugin) getNodeAllocatedResources(client *api.Client, nodeID string,
 
 func parseNodePoolQuery(q string) (*nodePoolQuery, error) {
 
-	mainParts := strings.SplitN(q, "/", 3)
-	if len(mainParts) != 3 {
-		return nil, fmt.Errorf("expected <query>/<pool_identifier_value>/<pool_identifier_key>, received %s", q)
+	// Split into query prefix and pool identifier part.
+	mainParts := strings.SplitN(q, "/", 2)
+	if len(mainParts) != 2 || mainParts[1] == "" {
+		return nil, fmt.Errorf("expected <query>/<key>=<value>[+<key>=<value>...], received %s", q)
```
gulducat: I like to use `%q` for user-provided input, to set the value apart from the error a bit. It also makes empty strings `""` clearer.

```suggestion
		return nil, fmt.Errorf("expected <query>/<key>=<value>[+<key>=<value>...], received %q", q)
```

sdk/helper/scaleutils/nodepool/combined.go
```diff
@@ -1,72 +1,76 @@
...
+// Encode serializes the identifier list into the combined query format:
+// key1=value1+key2=value2. Values are URL-encoded to handle special
+// characters (+, =, /).
+func (l ClusterNodePoolIdentifierList) Encode() string {
+	parts := make([]string, 0, len(l))
+	for _, id := range l {
+		parts = append(parts, fmt.Sprintf("%s=%s", id.Key(), url.QueryEscape(id.Value())))
```
gulducat, line 36: There's a hitch here: `url.QueryEscape(" ") == "+"` so any values with spaces will look like a new key/val pair.

We could error (as early as we can - policy validation?) if a value has a space, but Nomad allows spaces in some (all?) of these.

plugins/builtin/apm/nomad/plugin/node_test.go
```diff
@@ -424,7 +484,7 @@ func TestAPMPlugin_getPoolResources(t *testing.T) {
 			if tc.expectError {
 				assert.Error(t, err)
 				if tc.expectedErr != "" {
-					assert.ErrorContains(t, err, tc.expectedErr)
```
gulducat, line 427: This is a weird little change. Did this line of code change more than once over the course of this PR?

gulducat (commented):
This still isn't quite a complete review, but here are some more comments for ya.

policy/policy.go
gulducat: `Processor.ValidatePolicy` seems right to me to catch an error, but I don't usually expect a `Validate` method to also secretly mutate the thing being validated. It should just tell you if stuff is ok.

There are various bits of strangeness in this repo, but one of them is the intermingling of "sdk" with application internals, but short of refactoring an awful lot of stuff, maybe we lean into that. `normalizeNodePoolQuery` only refers to `sdk.TargetConfigKey**` values (the `(pr *Processor)` receiver is only used to log, which also smells...), so it kinda seems like this could live in `sdk`, too.

`sdk/ScalingPolicy` has a `Validate()` method that iterates over `ScalingPolicy.Checks`. We could add the static validation there, namely the one single error condition:

```
	key := normalizePoolKey(parts[2])
	if key == "" { ... error ... }
```

`sdk/ScalingPolicy.Validate()` gets called by the `policy/Manager` in its policy monitoring loop, so if it fails there, I don't think it ever makes it to this Processor.

The author of this method also left us a doc comment to consider. 

> // ... The function can be called without any validation on the check.

Which seems to _want_ to be able to leave the `c.Query` "fully hydrated" without requiring any pre-validation. The func in its original state: https://github.com/hashicorp/nomad-autoscaler/commit/8103537718e443e9438fa84192650f426b5dd59e#diff-82bdb5b5421b9da021e9c0af06d41fa22ff7ac433bb0c7cbef393c60be371c36R58-R98

That could mean that the query it produces is bad and not good and will fail, but it is "hydrated"?

Sorry I don't have a good conclusion here. I have to run away to a short vacation! I'll gladly take a look when I get back next week, but also feel free to ask other folks to review in the meantime.

plugins/builtin/apm/nomad/plugin/node_test.go
ritesh-harihar: Yes, it changed during an intermediate iteration. I have reverted it. Sorry for the noise.

PR #1296, opened 2026-05-13
feat(apm/influxdb): JWT auth fixes, env var credential fallback, and plugin hardening

hashi-divyansh:
This PR improves the InfluxDB APM plugin in three areas: fixing JWT-based authentication, adding environment variable credential fallback, and a broad set of code quality improvements.

**Changes:**

**JWT Token Authentication Fix**
- This PR improves the InfluxDB APM plugin in three areas: fixing JWT-based authentication, adding environment variable credential fallback, and a broad set of code quality improvements.
- Corrected the auth header format and token transmission.

**Environment Variable Credential Fallback**
- Credentials can now be supplied via environment variables, eliminating the need to hardcode them in HCL config files. The config map always takes priority; env vars are only consulted when a key is absent or empty.

[PR which introduces this new APM plugin](https://github.com/hashicorp/nomad-autoscaler/pull/1248)

[InfluxDB Documentation](https://docs.influxdata.com/influxdb/v1/administration/authentication_and_authorization/)
[
Autoscaler APM plugin doc for Influxdb](https://github.com/hashicorp/web-unified-docs/blob/update-nomad-autoscaler-docs-release-v0.5.0/content/nomad/v2.0.x/content/tools/autoscaling/plugins/apm/influxdb.mdx)

### DIFF
--- a/plugins/builtin/apm/influxdb/plugin/plugin.go
+++ b/plugins/builtin/apm/influxdb/plugin/plugin.go
@@ -10,10 +10,13 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
+	jwtlib "github.com/golang-jwt/jwt/v5"
 	hclog "github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/nomad-autoscaler/plugins"
 	"github.com/hashicorp/nomad-autoscaler/plugins/apm"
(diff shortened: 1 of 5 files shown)


plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -10,10 +10,13 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
+	jwtlib "github.com/golang-jwt/jwt/v5"
```
tgross, line 19: Why are we aliasing this?

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -249,36 +311,78 @@ func (a *APMPlugin) QueryMultiple(q string, r sdk.TimeRange) ([]sdk.TimestampedM
...
+	return token, nil
+}
+
+// generateJWT creates a new HS256-signed JWT for InfluxDB 1.x Bearer auth.
+// The JWT payload contains the configured username and an expiry of now+tokenTTL,
+// matching the format expected by InfluxDB's shared-secret JWT authentication.
+func (a *APMPlugin) generateJWT() (string, time.Time, error) {
+	expiry := time.Now().Add(a.tokenTTL)
+	claims := jwtlib.MapClaims{
+		"username": strings.TrimSpace(a.config[configKeyUsername]),
```
tgross, line 372: Why are we trimming configuration way down here in the token generation, rather than when the configuration is being parsed?

tgross (changes requested):
@hashi-divyansh you're missing a changelog, and a changelog entry would illustrate to you quickly that this PR should be 3 different PRs. You've described this as a bug, but we're _newly introducing_ the JWT auth code here, which doesn't look like a bug to me. You're also adding this env var credential fallback feature, and 100s of lines of style changes. You _might_ be able to get away with that if the PR was very small, but this is a chonky one, _especially_ given we're writing brand new auth code.

You'll need to break this up before it can be reviewed.

PR #1297, opened 2026-05-14
feat(apm/influxdb): add JWT Bearer authentication via shared secret for influxdb 1.x

hashi-divyansh:
Adds HS256 JWT Bearer authentication to the InfluxDB APM plugin, supporting InfluxDB 1.x shared-secret auth (`INFLUXDB_HTTP_SHARED_SECRET`).

[InfluxDB Auth Documentation](https://docs.influxdata.com/influxdb/v1/administration/authentication_and_authorization/)

**Configuration Changes**
- shared_secret`  HS256 signing secret (requires `username`) 
- shared_secret` and `password` are mutually exclusive.

How

- All config is parsed into a typed `pluginConfig` struct in `SetConfig` — values are trimmed once at the boundary, not scattered across downstream methods.
- JWTs are stateless(not cached), new token will be created every time with expiry duration = 2m. 

### DIFF
--- a/plugins/builtin/apm/influxdb/plugin/auth.go
+++ b/plugins/builtin/apm/influxdb/plugin/auth.go
@@ -30,6 +30,12 @@ const (
 	maxTokenTTL = 24 * time.Hour
 )
 
+// influxClaims are the JWT claims expected by InfluxDB 1.x shared-secret auth.
+type influxClaims struct {
+	Username string `json:"username"`
+	jwt.RegisteredClaims
+}
+
 // setAuthHeader applies the appropriate authentication method to the HTTP
 // request based on the plugin configuration:
 //   - shared_secret + username → auto-generated HS256 JWT Bearer token
@@ -85,9 +91,11 @@ func (a *APMPlugin) getOrRefreshJWT() (string, error) {
 // matching the format expected by InfluxDB's shared-secret JWT authentication.
 func (a *APMPlugin) generateJWT() (string, time.Time, error) {
 	expiry := time.Now().Add(a.cfg.TokenTTL)
-	claims := jwt.MapClaims{
-		"username": a.cfg.Username,
-		"exp":      expiry.Unix(),
+	claims := influxClaims{
+		Username: a.cfg.Username,
+		RegisteredClaims: jwt.RegisteredClaims{
+			ExpiresAt: jwt.NewNumericDate(expiry),
+		},
 	}
 	tok := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
 	signed, err := tok.SignedString([]byte(a.cfg.SharedSecret))
--- a/plugins/builtin/apm/influxdb/plugin/auth.go
+++ b/plugins/builtin/apm/influxdb/plugin/auth.go
@@ -35,6 +35,8 @@ const (
 )
 
 // influxClaims are the JWT claims expected by InfluxDB 1.x shared-secret auth.
+// InfluxDB only validates exp and username; the remaining fields from
+// jwt.RegisteredClaims are omitted from the token payload via omitempty.
 type influxClaims struct {
 	Username string `json:"username"`
 	jwt.RegisteredClaims
--- a/plugins/builtin/apm/influxdb/plugin/auth.go
+++ b/plugins/builtin/apm/influxdb/plugin/auth.go
@@ -26,8 +26,12 @@ const (
 	defaultTokenTTL = time.Hour
 
 	// minTokenTTL / maxTokenTTL are the allowed bounds for token_ttl.
-	minTokenTTL = time.Minute
+	minTokenTTL = 10 * time.Minute
 	maxTokenTTL = 24 * time.Hour
+
+	// jwtRefreshBuffer is how long before expiry a new JWT is generated.
+	// Must be less than minTokenTTL to ensure the token is always cached.
+	jwtRefreshBuffer = 2 * time.Minute
 )
 
 // influxClaims are the JWT claims expected by InfluxDB 1.x shared-secret auth.
@@ -58,20 +62,13 @@ func (a *APMPlugin) setAuthHeader(req *http.Request) error {
 }
 
 // getOrRefreshJWT returns the cached JWT if it is still fresh, or generates
-// and caches a new one. The refresh window is max(30s, 10% of token_ttl),
-// so a token is regenerated slightly before it would expire.
+// and caches a new one. The token is refreshed jwtRefreshBuffer before expiry,
+// ensuring it remains valid for at least one full evaluation cycle.
 func (a *APMPlugin) getOrRefreshJWT() (string, error) {
-	var refreshWindow time.Duration
-	if w := a.cfg.TokenTTL / 10; w > 30*time.Second {
-		refreshWindow = w
-	} else {
-		refreshWindow = 30 * time.Second
-	}
-
 	a.jwtMu.Lock()
 	defer a.jwtMu.Unlock()
 
-	if a.cachedToken != "" && time.Until(a.tokenExpiry) > refreshWindow {
+	if a.cachedToken != "" && time.Until(a.tokenExpiry) > jwtRefreshBuffer {
 		return a.cachedToken, nil
 	}
 
--- a/plugins/builtin/apm/influxdb/plugin/auth.go
+++ b/plugins/builtin/apm/influxdb/plugin/auth.go
@@ -1,0 +1,98 @@
+// Copyright IBM Corp. 2020, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package plugin
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+
+	"github.com/golang-jwt/jwt/v5"
+)
+
+const (
+	// configKeySharedSecret is the InfluxDB shared secret used to sign HS256 JWTs.
+	// Requires: username. Conflicts with: password.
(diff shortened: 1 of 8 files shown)


plugins/builtin/apm/influxdb/plugin/auth.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// minTokenTTL / maxTokenTTL are the allowed bounds for token_ttl.
+	minTokenTTL = time.Minute
+	maxTokenTTL = 24 * time.Hour
+)
+
+// influxClaims are the JWT claims expected by InfluxDB 1.x shared-secret auth.
+type influxClaims struct {
+	Username string `json:"username"`
+	jwt.RegisteredClaims
```
tgross, line 31: Noting here for any other reviewers because I had to look it up myself: for some reason the Influx docs show that they only care about `exp` and not any of the other fields we're embedding here. But `jwt.RegisteredClaims` has `omitempty` on all those fields, so 🤷 seems ok.

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +112,90 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
 }
 
 // SetConfig parses and validates the plugin configuration. All required fields
-// are checked before any client is stored.
+// are checked before any state is mutated, so a failed re-configuration leaves
+// the plugin in its previous working state.
```
tgross: > re-configuration

This implies that we can call `SetConfig` again on a running plugin? If so, we'll need to put a RW mutex around the writes to and reads from `a.cfg`, `a.baseURL`, and `a.client` so that we're not mutating `APMPlugin` unsafely.

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +112,90 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
...
+	a.tokenExpiry = time.Time{}
+	a.jwtMu.Unlock()
+
+	return nil
+}
+
+// parseConfig validates the raw config map and returns the normalised
+// pluginConfig and its parsed base URL. Returns an error if any required
+// field is missing, a field value is invalid, or auth options conflict.
+func parseConfig(config map[string]string) (pluginConfig, *url.URL, error) {
```
tgross: Good call on returning a struct here instead of the map 👍 

But why do we return the base URL here instead of passing it back in the configuration? It seems like we're carrying around the address in the config but I don't know if it ever gets consumed outside of this function?

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +112,90 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
...
+	var cfg pluginConfig
+
+	cfg.Address = strings.TrimSpace(config[configKeyAddress])
+	if cfg.Address == "" {
+		return pluginConfig{}, nil, fmt.Errorf("%q config value cannot be empty", configKeyAddress)
```
tgross: Nitpick, but we've already allocated this struct:

```suggestion
		return cfg, nil, fmt.Errorf("%q config value cannot be empty", configKeyAddress)
```

(Typical of a few of these)

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +112,90 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
...
-	if token != "" && (username != "" || password != "") {
-		return fmt.Errorf("conflicting auth configuration: %q cannot be used together with %q or %q", configKeyToken, configKeyUsername, configKeyPassword)
+	cfg.Username = strings.TrimSpace(config[configKeyUsername])
+	cfg.Password = strings.TrimSpace(config[configKeyPassword])
+	cfg.SharedSecret = strings.TrimSpace(config[configKeySharedSecret])
+
+	if cfg.SharedSecret != "" {
+		if cfg.Username == "" {
+			return pluginConfig{}, nil, fmt.Errorf("auth configuration error: %q requires %q (used as the JWT username claim)", configKeySharedSecret, configKeyUsername)
+		}
```
tgross, line 148: I think we're missing validation for when we have no shared secret and need to check that we have both username and password?

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +112,90 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
...
+			return pluginConfig{}, nil, fmt.Errorf("conflicting auth configuration: %q cannot be used together with %q", configKeySharedSecret, configKeyPassword)
+		}
 	}
 
-	switch version := strings.TrimSpace(a.config[configKeyVersion]); version {
+	cfg.TokenTTL = defaultTokenTTL
+	if raw := strings.TrimSpace(config[configKeyTokenTTL]); raw != "" {
+		parsed, err := time.ParseDuration(raw)
+		if err != nil {
+			return pluginConfig{}, nil, fmt.Errorf("invalid %q value %q: %w", configKeyTokenTTL, raw, err)
```
tgross: Style nitpick: I'd drop the quotes around the config key because this is the key and not the value.

plugins/builtin/apm/influxdb/plugin/auth.go
```diff
@@ -0,0 +1,105 @@
...
+// getOrRefreshJWT returns the cached JWT if it is still fresh, or generates
+// and caches a new one. The refresh window is max(30s, 10% of token_ttl),
+// so a token is regenerated slightly before it would expire.
+func (a *APMPlugin) getOrRefreshJWT() (string, error) {
+	var refreshWindow time.Duration
+	if w := a.cfg.TokenTTL / 10; w > 30*time.Second {
+		refreshWindow = w
+	} else {
+		refreshWindow = 30 * time.Second
+	}
```
tgross: Design question: why such a low % of expiry (10%) before refreshing the token, and if that's important, why the 30s floor if we let the operator configure a 1min expiry? Are the queries we make expected to take a very long time where this large window is necessary?

plugins/builtin/apm/influxdb/plugin/auth_test.go
```diff
@@ -0,0 +1,144 @@
...
+	must.NotEq(t, "", tok1)
+
+	// Second call within TTL — same token returned (cached).
+	tok2, err := p.getOrRefreshJWT()
+	must.NoError(t, err)
+	must.Eq(t, tok1, tok2, must.Sprint("expected cached token to be reused"))
+
+	// Sleep 1s so the next token gets a different exp. JWT timestamps are
+	// whole seconds, so same-second regeneration produces an identical string.
+	time.Sleep(time.Second)
```
tgross: Something that can make tests nicer is to have the `getOrRefreshJWT()` function take a timestamp. Then you can test determinstically. But short of that, you can probably use https://pkg.go.dev/testing/synctest to accelerate time here and speed up your sleep.

plugins/builtin/apm/influxdb/plugin/plugin_test.go
tgross, line 156: Combine these branches?

plugins/builtin/apm/influxdb/plugin/plugin.go
hashi-divyansh: No, it will only be called at startup. I'll update the comment.

plugins/builtin/apm/influxdb/plugin/auth.go
hashi-divyansh: I was thinking about the case, What if evaluation window is large. I think it will be difficult to choose perfect value here because plugin is not aware of autoscaler evaluation interval. 

I drop cached based approach and opt for stateless JWT in this [commit ](https://github.com/hashicorp/nomad-autoscaler/pull/1297/changes/afaecbaef0d37ff7a196047c4c048aa00fd37957). I believe this is the best approach (for this specific type of JWT creation).  For more details please read comment on `generateJWT()
`

plugins/builtin/apm/influxdb/plugin/plugin.go
```diff
@@ -104,49 +106,75 @@ func NewInfluxDBPlugin(log hclog.Logger) apm.APM {
...
+		if cfg.Username == "" {
+			return cfg, fmt.Errorf("auth configuration error: %s requires %s (used as the JWT username claim)", configKeySharedSecret, configKeyUsername)
+		}
+		if cfg.Password != "" {
+			return cfg, fmt.Errorf("conflicting auth configuration: %s cannot be used together with %s", configKeySharedSecret, configKeyPassword)
+		}
+	} else if cfg.Username != "" && cfg.Password == "" {
+		return cfg, fmt.Errorf("auth configuration error: %s requires %s for Basic authentication", configKeyUsername, configKeyPassword)
+	} else if cfg.Password != "" && cfg.Username == "" {
+		return cfg, fmt.Errorf("auth configuration error: %s requires %s for Basic authentication", configKeyPassword, configKeyUsername)
```
tgross, line 155: You've got duplicate cases here

plugins/builtin/apm/influxdb/plugin/auth.go
tgross: Ok, probably fine. I imagine there's a bit of extra garbage to clean up as a result, but we're not making 1000s of concurrent queries or anything like that

tgross (approved):
LGTM!

PR #1300, opened 2026-05-18
Fix: added support for datacenter, combined query

### LINKED ISSUE #599: nomad-apm: add support to group nodes by datacenter

ritesh-harihar:
[JIRA ISSUE](https://hashicorp.atlassian.net/browse/NMD-1394)

**_Previous PR:_** https://github.com/hashicorp/nomad-autoscaler/pull/1291

**_Problem_**

When writing a cluster scaling policy, operators could only filter nodes by node_class:

```
target "aws-asg" {
  node_class = "hashistack"    
  node_drain_deadline = "5m"
}
```

There was no way to filter by datacenter or node_pool. Setting either was silently ignored — the autoscaler would include all nodes regardless, producing incorrect resource totals and wrong scaling decisions.

**_Solution_**

Operators can now use datacenter and node_pool in the target block, individually or combined:

# Filter by datacenter
```
target "aws-asg" {
  datacenter = "us-east-1"
  node_drain_deadline = "5m"
}
```

# Filter by node_pool
```
target "aws-asg" {
  node_pool = "gpu"
  node_drain_deadline = "5m"
}
```

# Filter by both (AND logic — only nodes matching ALL keys)
```
target "aws-asg" {
  node_class = "hashistack"
  datacenter = "us-east-1"
  node_drain_deadline = "5m"
}
```

# Before (only node_class worked)
Policy target config | Internal query generated | Result
-- | -- | --
node_class = "hashistack" | node_percentage-allocated_memory/hashistack/class | Correct — filters by class
datacenter = "dc1" | Ignored — no query expansion, falls through to unfiltered | Returns ALL nodes
node_class = "hashistack" + datacenter = "dc1" | node_percentage-allocated_memory/hashistack/class | Only filters by class, ignores datacenter
node_pool = "gpu" | Ignored | Returns ALL nodes

# After (all keys supported)
Policy target config | Internal query generated | Result
-- | -- | --
node_class = "hashistack" | node_percentage-allocated_memory/node_class=hashistack | Same behavior
datacenter = "dc1" | node_percentage-allocated_memory/datacenter=dc1 | Now filters by datacenter
node_class = "hashistack" + datacenter = "dc1" | node_percentage-allocated_memory/node_class=hashistack,datacenter=dc1 | AND logic — only nodes matching both
node_pool = "gpu" | node_percentage-allocated_memory/node_pool=gpu | Now filters by node_pool
node_class = "hashistack" + datacenter = "dc1" + node_pool = "gpu" | node_percentage-allocated_memory/node_class=hashistack,datacenter=dc1,node_pool=gpu | AND — all three must match

# Old long-form queries are not supported now

Old format | Output
-- | --
node_percentage-allocated_memory/hashistack/class  | ERROR: invalid pool identifier "hashistack/class": expected key=value

The key change: old format used /value/key, new format uses /key=value (with , for combining multiple keys).

**_Data Flow Summary:_**
```
HCL target config
  → IsNodePoolTarget()          [sdk/policy.go — checks node_class || datacenter || node_pool]
  → CanonicalizeAPMQuery()      [policy/policy.go — builds "node_<op>_<metric>/key=url.QueryEscape(val),key=val"]
  → APM plugin: parseNodePoolQuery()  [node.go — splits on "," then "=", url.QueryUnescape, creates identifiers]
  → getPoolResources()          [node.go — filters nodes via ClusterNodePoolIdentifierList.IsPoolMember (AND logic)]
```

**_Testing:_**
<details>

1) client1 -> cpu_total_compute = 2000, memory_total_mb = 2048 2) client2 -> cpu_total_compute = 4000, memory_total_mb = 4096 3) client3 -> cpu_total_compute = 8000, memory_total_mb = 8192

Screenshot 2026-05-18 at 10 49 35 AM

```
> 2026-05-21T18:50:33.019+0530 [DEBUG] internal_plugin.nomad-apm: collected node pool resource data: query=node_percentage-allocated_cpu/datacenter=dc2 allocated_cpu=0 allocated_memory=0 allocatable_cpu=8000 allocatable_memory=8192
> 2026-05-21T18:50:33.019+0530 [DEBUG] internal_plugin.nomad-apm: collected node pool resource data: query="node_percentage-allocated_memory/node_class=hashistack,datacenter=dc1" allocated_cpu=0 allocated_memory=0 allocatable_cpu=2000 allocatable_memory=2048
> 2026-05-21T18:50:33.020+0530 [DEBUG] internal_plugin.nomad-apm: collected node pool resource data: query=node_percentage-allocated_memory/datacenter=dc1 allocated_cpu=10 allocated_memory=12 allocatable_cpu=6000 allocatable_memory=6144
> 2026-05-21T18:50:33.020+0530 [DEBUG] internal_plugin.nomad-apm: collected node pool resource data: query=node_percentage-allocated_cpu/node_class=hashistack allocated_cpu=0 allocated_memory=0 allocatable_cpu=10000 allocatable_memory=10240
```

Screenshot 2026-05-21 at 7 15 40 PM

-

</details>

### DIFF
--- a/plugins/builtin/apm/nomad/plugin/node.go
+++ b/plugins/builtin/apm/nomad/plugin/node.go
@@ -180,6 +180,10 @@ func (a *APMPlugin) getNodeAllocatedResources(client *api.Client, nodeID string,
 	return nil
 }
 
+// parseNodePoolQuery decodes a node pool query in the internal wire format:
+// node_<op>_<metric>/key1=val1[,key2=val2...]
+//
+// The inverse operation is formatNodePoolQuery in policy/policy.go.
 func parseNodePoolQuery(q string) (*nodePoolQuery, error) {
 
 	// Split into query prefix and pool identifier part.
--- a/plugins/builtin/apm/nomad/plugin/node.go
+++ b/plugins/builtin/apm/nomad/plugin/node.go
@@ -6,6 +6,7 @@ package plugin
 import (
 	"errors"
 	"fmt"
+	"net/url"
 	"strings"
 	"time"
 
@@ -18,9 +19,9 @@ import (
 // nodePoolQuery is the plugins internal representation of a query and contains
 // all the information needed to perform a Nomad APM query for a node pool.
 type nodePoolQuery struct {
-	metric         string
-	poolIdentifier nodepool.ClusterNodePoolIdentifier
-	operation      string
+	metric    string
+	poolIDs   nodepool.ClusterNodePoolIdentifierList
+	operation string
 }
 
 type nodePoolResources struct {
@@ -46,11 +47,12 @@ func (a *APMPlugin) queryNodePool(q string) (sdk.TimestampedMetrics, error) {
 	a.logger.Debug("performing node pool APM query", "query", q)
 
 	// Identify the resource available and consumed within the target pool.
-	resources, err := a.getPoolResources(query.poolIdentifier)
+	resources, err := a.getPoolResources(query.poolIDs)
 	if err != nil {
 		return nil, err
 	}
 	a.logger.Debug("collected node pool resource data",
+		"query", q,
 		"allocated_cpu", resources.allocated.cpu, "allocated_memory", resources.allocated.mem,
 		"allocatable_cpu", resources.allocatable.cpu, "allocatable_memory", resources.allocatable.mem)
 
@@ -82,7 +84,7 @@ func (a *APMPlugin) queryNodePool(q string) (sdk.TimestampedMetrics, error) {
 // specified node pool. Any error in calling the Nomad API for details will
 // result in an error. This is because with missing data, we cannot reliably
 // make calculations.
-func (a *APMPlugin) getPoolResources(id nodepool.ClusterNodePoolIdentifier) (*nodePoolResources, error) {
+func (a *APMPlugin) getPoolResources(ids nodepool.ClusterNodePoolIdentifierList) (*nodePoolResources, error) {
 
 	client, config := a.getClientAndConfigSnapshot()
 	if client == nil {
@@ -101,7 +103,7 @@ func (a *APMPlugin) getPoolResources(id nodepool.ClusterNodePoolIdentifier) (*no
 
 	// Perform our node filtering so we are left with a list of nodes that form
 	// our pool and that are in the correct state.
-	nodePoolList, err := scaleutils.FilterNodesWithOptions(nodes, id.IsPoolMember, filterOpts)
+	nodePoolList, err := scaleutils.FilterNodesWithOptions(nodes, ids.IsPoolMember, filterOpts)
 	if err != nil {
 		return nil, fmt.Errorf("failed to identify nodes within pool: %v", err)
 	}
@@ -180,13 +182,36 @@ func (a *APMPlugin) getNodeAllocatedResources(client *api.Client, nodeID string,
 
 func parseNodePoolQuery(q string) (*nodePoolQuery, error) {
 
-	mainParts := strings.SplitN(q, "/", 3)
-	if len(mainParts) != 3 {
-		return nil, fmt.Errorf("expected <query>/<pool_identifier_value>/<pool_identifier_key>, received %s", q)
+	// Split into query prefix and pool identifier part.
+	mainParts := strings.SplitN(q, "/", 2)
+	if len(mainParts) != 2 || mainParts[1] == "" {
+		return nil, fmt.Errorf("expected <query>/<key>=<value>[,<key>=<value>...], received %q", q)
 	}
 
-	query := nodePoolQuery{
-		poolIdentifier: nodepool.NewNodeClassPoolIdentifier(mainParts[1]),
+	var query nodePoolQuery
+
+	// Parse the combined format: key1=val1[,key2=val2...]
+	for _, pair := range strings.Split(mainParts[1], ",") {
+		kv := strings.SplitN(pair, "=", 2)
+		if len(kv) != 2 || kv[0] == "" || kv[1] == "" {
+			return nil, fmt.Errorf("invalid pool identifier %q: expected key=value", pair)
+		}
+
+		value, err := url.QueryUnescape(kv[1])
+		if err != nil {
+			return nil, fmt.Errorf("failed to decode value in %q: %v", pair, err)
+		}
+
+		switch kv[0] {
+		case sdk.TargetConfigKeyClass:
+			query.poolIDs = append(query.poolIDs, nodepool.NewNodeClassPoolIdentifier(value))
+		case sdk.TargetConfigKeyDatacenter:
+			query.poolIDs = append(query.poolIDs, nodepool.NewNodeDatacenterPoolIdentifier(value))
+		case sdk.TargetConfigKeyNodePool:
+			query.poolIDs = append(query.poolIDs, nodepool.NewNodePoolClusterPoolIdentifier(value))
+		default:
+			return nil, fmt.Errorf("unknown pool key %q: must be node_class, datacenter, or node_pool", kv[0])
+		}
 	}
 
 	opMetricParts := strings.SplitN(mainParts[0], "_", 3)
--- a/policy/policy.go
+++ b/policy/policy.go
@@ -85,11 +85,15 @@ func (pr *Processor) ValidatePolicy(p *sdk.ScalingPolicy) error {
 
 	// Validate that Nomad APM queries for node pool targets use the short format.
 	// Long-form queries are internal only and must not be written by operators.
-	if p.Target != nil && p.Target.IsNodePoolTarget() {
-		for _, c := range p.Checks {
-			if c == nil {
-				continue
-			}
+	isNodePool := p.Target != nil && p.Target.IsNodePoolTarget()
+
+	for _, c := range p.Checks {
+		if c == nil {
+			continue
+		}
+
+		// Node pool APM query format validation.
+		if isNodePool {
 			source := c.Source
 			if source == "" {
 				source = plugins.InternalAPMNomad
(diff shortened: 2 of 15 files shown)


plugins/builtin/apm/nomad/plugin/node.go
```diff
@@ -180,13 +182,36 @@ func (a *APMPlugin) getNodeAllocatedResources(client *api.Client, nodeID string,
 
 func parseNodePoolQuery(q string) (*nodePoolQuery, error) {
 
-	mainParts := strings.SplitN(q, "/", 3)
-	if len(mainParts) != 3 {
-		return nil, fmt.Errorf("expected <query>/<pool_identifier_value>/<pool_identifier_key>, received %s", q)
+	// Split into query prefix and pool identifier part.
+	mainParts := strings.SplitN(q, "/", 2)
```
ritesh-harihar, line 190: parseNodePoolQuery() splits on` ,`  then `=` ,  applies url.QueryUnescape() to values, builds ClusterNodePoolIdentifierList for multi-key AND filtering

policy/policy.go
```diff
@@ -82,6 +83,25 @@ func (pr *Processor) ValidatePolicy(p *sdk.ScalingPolicy) error {
 		mErr = multierror.Append(mErr, errors.New("policy Min must not be greater Max"))
 	}
 
+	// Validate that Nomad APM queries for node pool targets use the short format.
+	// Long-form queries are internal only and must not be written by operators.
+	if p.Target != nil && p.Target.IsNodePoolTarget() {
```
ritesh-harihar: This block rejects any longform query (containing `/ `) for Nomad APM checks on node pool targets — operators must only write short-form queries.

sdk/helper/scaleutils/nodepool/nodepool.go
```diff
@@ -46,12 +47,21 @@ func NewClusterNodePoolIdentifier(cfg map[string]string) (ClusterNodePoolIdentif
...
 	}
+	return ids, nil
+}
+
+// Deprecated: Use NewClusterNodePoolIdentifierList for multi-key AND filtering.
+// NewClusterNodePoolIdentifier returns a single ClusterNodePoolIdentifier from
+// the provided configuration. For backward compatibility, if multiple keys are
+// present only the first recognized identifier is returned.
+// TODO: Remove this function in next Release.
+func NewClusterNodePoolIdentifier(cfg map[string]string) (ClusterNodePoolIdentifier, error) {
```
ritesh-harihar, line 61: Config passed | ids[0] returned | Behavior
-- | -- | --
{"node_class": "hashistack"} | classIdentifier | Filters by class only  (same as old)
{"datacenter": "dc1"} | dcIdentifier | Filters by datacenter only 
{"node_pool": "prod"} | poolIdentifier | Filters by node_pool only 
{"node_class": "hashistack", "datacenter": "dc1"} | classIdentifier | Filters by class only (datacenter ignored — same as old)
{"node_class": "hashistack", "datacenter": "dc1", "node_pool": "prod"} | classIdentifier | Filters by class only (dc + pool ignored — same as old)
{"datacenter": "dc1", "node_pool": "prod"} | dcIdentifier | Filters by datacenter only (pool ignored)
{} (empty) | error | "node pool identification method required" ✓ (same as old)

sdk/helper/scaleutils/nodepool/combined_deprecated.go
```diff
@@ -0,0 +1,78 @@
+// Copyright IBM Corp. 2020, 2026
+// SPDX-License-Identifier: MPL-2.0
+// This file contains a deprecated implementation of a combined ClusterNodePoolIdentifier 
+// It is retained for backward compatibility with existing code that uses this implementation
+// TODO: Remove this file and all references in next Release.
```
ritesh-harihar, line 5: Re-added `NewCombinedClusterPoolIdentifier` and `NewClusterNodePoolIdentifier `as
deprecated wrappers in standalone files (combined_deprecated.go, combined_deprecated_test.go) so external plugin authors continue to compile without changes while they migrate to `ClusterNodePoolIdentifierList`.

Intentionally kept decoupled from the new `ClusterNodePoolIdentifierList` flow. I considered an alternative approach of delegating internally to ClusterNodePoolIdentifierList but the coupling would make future removal harder. As standalone files, removal in a later release is a clean with no effects on the rest of the codebase.

policy/policy.go
```diff
@@ -82,6 +83,25 @@ func (pr *Processor) ValidatePolicy(p *sdk.ScalingPolicy) error {
...
+		for _, c := range p.Checks {
+			if c == nil {
+				continue
```
tehut: I noticed that we're ranging over the same checks twice. I see how the really opinionated checks on lns 105 & 106 might have made you want to just slide your loop in above it but I think the right move would be to put your query validation in the loop first.

policy/policy.go
tehut: I know you didn't write this section but it would make the function easier to extend to rewrite this so that we execute the validation when QueryInstant is true, & when Strategy.Name == plugins.InternalStrategyThreshold.

That will make the code easier to reason about and prevent other people from feeling compelled to do separate iterations over this section of config.

sdk/helper/scaleutils/nodepool/nodepool_test.go
```diff
@@ -7,88 +7,140 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/shoenig/test/must"
 	"github.com/stretchr/testify/assert"
 )
 
-func Test_NewClusterNodePoolIdentifier(t *testing.T) {
```
tehut, line 13: If we're keeping this function for backwards compatibility in nodepool.go we should keep the test unchanged and make new tests for Test_NewClusterNodePoolIdentifierList.

sdk/helper/scaleutils/nodepool/nodepool.go
```diff
@@ -46,12 +47,21 @@ func NewClusterNodePoolIdentifier(cfg map[string]string) (ClusterNodePoolIdentif
...
 	}
 
-	switch len(ids) {
-	case 0:
+	if len(ids) == 0 {
 		return nil, fmt.Errorf("node pool identification method required")
-	case 1:
-		return ids[0], nil
-	default:
-		return NewCombinedClusterPoolIdentifier(ids, CombinedClusterPoolIdentifierAnd), nil
```
tehut, line 55: this case looks like it's been dropped

sdk/helper/scaleutils/nodepool/nodepool.go
```diff
@@ -46,12 +47,21 @@ func NewClusterNodePoolIdentifier(cfg map[string]string) (ClusterNodePoolIdentif
...
 	}
+	return ids, nil
+}
+
+// Deprecated: Use NewClusterNodePoolIdentifierList for multi-key AND filtering.
+// NewClusterNodePoolIdentifier returns a single ClusterNodePoolIdentifier from
+// the provided configuration. For backward compatibility, if multiple keys are
+// present only the first recognized identifier is returned.
```
tehut: I don't think this is doing the same thing as the original version of the function. If we're deprecating it, why change it?

policy/policy.go
```diff
@@ -167,12 +186,16 @@ func (pr *Processor) CanonicalizeAPMQuery(c *sdk.ScalingPolicyCheck, t *sdk.Scal
 	}
 
 	// If the target is a Nomad client node pool, format the query in the
-	// expected manner. Once the autoscaler supports more than just class
-	// identification of pools this func and logic will need to be updated. For
-	// now keep it simple.
+	// combined format: node_<op>_<metric>/key1=val1,key2=val2
```
tehut: It took me a bit to find where this encoding was happening. I think it would be easier to find this if it were split into a small private function that cannonicalize can call. 

That way you can reference the new function in a doccomment on parseNodePoolQuery and vice versa.

sdk/helper/scaleutils/cluster.go
```diff
@@ -224,11 +224,11 @@ func (c *ClusterScaleUtils) IdentifyScaleInNodes(cfg map[string]string, num int)
 
 func (c *ClusterScaleUtils) identifyEligibleScaleInNodes(cfg map[string]string) ([]*api.NodeListStub, error) {
 
-	poolID, err := nodepool.NewClusterNodePoolIdentifier(cfg)
+	ids, err := nodepool.NewClusterNodePoolIdentifierList(cfg)
 	if err != nil {
 		return nil, err
 	}
-	c.log.Debug("performing node pool filtering", poolID.Key(), poolID.Value())
+	c.log.Debug("performing node pool filtering", "identifiers", fmt.Sprintf("%v", ids))
```
tehut: What does the slice of structs type look like as a `"%v"`?  My instinct says it might get displayed as a pointer value since it's a slice.

Something @gulducat reminded me recently is that you can add a `String()` method to a struct you define and then the fmt package will use your method when called.

https://www.zetcode.com/golang/fmt-stringer-interface/

sdk/helper/scaleutils/cluster.go
```diff
@@ -275,7 +275,7 @@ func (c *ClusterScaleUtils) identifyEligibleScaleInNodes(cfg map[string]string)
 
 	// Ensure we have not filtered out all the available nodes.
 	if len(filteredNodes) == 0 {
-		return nil, fmt.Errorf("no nodes unfiltered for %s with value %s", poolID.Key(), poolID.Value())
+		return nil, fmt.Errorf("no nodes unfiltered for pool identifiers %v", ids)
```
tehut: same as above `"%v"` on a slice will likely drop information

tehut (changes requested):
Overall, this looks really good. I didn't get too deep into tests but have some questions about backwards compatibility and a bit of clean up in the validation and canonicalization functions

sdk/helper/scaleutils/cluster.go
ritesh-harihar: Added string() method.

sdk/helper/scaleutils/nodepool/nodepool.go
ritesh-harihar: yeah. we can keep the original version.

sdk/helper/scaleutils/nodepool/nodepool.go
ritesh-harihar: Got re-added in the original version `NewCombinedClusterPoolIdentifier()`

ritesh-harihar:
@tehut I have resolved the requested changes.

tehut (approved):
thanks!

hashi-divyansh:
I think these new additions require documentation updates. @ritesh-harihar ?

ritesh-harihar:
> I think these new additions require documentation updates. @ritesh-harihar ?

Yes. will update. Docs PR Link:  https://github.com/hashicorp/web-unified-docs/pull/2574

PR #1310, opened 2026-06-08
Add Instana APM plugin scaffold and registration

hashi-divyansh:
**Description:**
[Ticket](https://hashicorp.atlassian.net/browse/NMD-1576?atlOrigin=eyJpIjoiMDcyNTU5N2RjNDM5NDlhYjhkMWJkMDBkMmNiNGNhNTAiLCJwIjoiaiJ9)
Adds the scaffolding for a new built-in Instana APM plugin. This is the first of two PRs to deliver full Instana support in the autoscaler.

**What's in this PR**
Adds the skeleton for the Instana built-in APM plugin — the struct, config validation, and registration wiring — so the autoscaler recognises `driver = "instana"` in a policy config. The plugin validates the required `endpoint` URL and `api_token` (with `INSTANA_API_TOKEN` env fallback) at startup but does not make any HTTP calls yet. Query logic comes in next follow-up PR .

### DIFF
--- a/plugins/builtin/apm/instana/main.go
+++ b/plugins/builtin/apm/instana/main.go
@@ -1,0 +1,19 @@
+// Copyright IBM Corp. 2020, 2026
+// SPDX-License-Identifier: MPL-2.0
+
+package main
+
+import (
+	hclog "github.com/hashicorp/go-hclog"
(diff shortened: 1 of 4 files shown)


Juanadelacuesta (approved):
So far it looks good!

PR #1311, opened 2026-06-10
feat(apm/instana): implement Instana infrastructure metrics APM plugin

hashi-divyansh:
**Description:** [Ticket](https://hashicorp.atlassian.net/browse/NMD-1577?atlOrigin=eyJpIjoiMTM0OWFlYTk5M2FmNGUzNmE3NjViNjc2NmZkNTNhNWYiLCJwIjoiaiJ9)

What

Adds a built-in Instana APM plugin so scaling policies can use infrastructure metrics from Instana — things like CPU, memory, JVM heap, NGINX request rate, Kafka throughput, and more.

### Configuration

```hcl
apm "instana" {
  driver = "instana"
  config = {
    endpoint  = "https://<unit>.instana.io"
    api_token = "<token>"                    # or set INSTANA_API_TOKEN env var
  }
}
```

Works with both Instana SaaS and self-hosted.

### Usage in a scaling policy

The `query` is a JSON object sent directly to the Instana metrics API. The time window is filled in automatically from `query_window`.

```hcl
check "mem_usage" {
  source       = "instana"
  query = jsonencode({
    plugin      = "host"
    snapshotIds = ["id1", "id2"]
    metrics     = ["memory.used"]
   })
  query_window = "5m"

  strategy "target-value" {
    target = 0.7   # scale out when memory exceeds 70%
  }
}
```

Use `snapshotIds` to target a single host (required for `Query()`). Use the `query` filter for fleet-wide queries with `QueryMultiple()`.

### Metrics you can scale on

| Plugin              | Example metrics                                        | Unit          |
|---------------------|--------------------------------------------------------|---------------|
| `host`              | `cpu.used`, `memory.used`, `cpu.idle`, `cpu.sys`      | Fraction 0–1  |
| `docker`            | `cpu.total_usage`, `memory.used_percentage`            | Fraction / %  |
| `jvmRuntimePlatform`| `memory.used`, `threads.runnable`, `threads.blocked`  | Bytes / Count |
| `nginx`             | `connections.active`, `requests`                       | Count / Rate  |
| `kafka`             | `broker.messagesIn`, `broker.bytesIn`                  | Rate          |
| `redis`             | `connected_clients`, `hit_rate`                        | Count / %     |

> Host metrics like `memory.used` are fractions — `0.77` means 77%. Set `target` accordingly.

### Testing

**End-to-end** — ran against IBM Instana sandbox with a real 3-node Nomad cluster:

- Scale-out: sandbox host at 77% memory → autoscaler scaled webapp `from=1 to=2` (`factor=1.55`)
- Scale-in: sandbox host at ~3% CPU → autoscaler scaled webapp `from=5 to=1` (`factor=0.06`)

> The sandbox host metrics aren't tied to the webapp's (our nomad cluster application) actual load — It was some already running application on the sandbox.


plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -139,7 +150,128 @@ func (a *APMPlugin) Query(q string, r sdk.TimeRange) (sdk.TimestampedMetrics, er
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, metricsURL.String(), bytes.NewReader(body))
+	if err != nil {
+		return nil, fmt.Errorf("failed to build instana query request: %w", err)
+	}
+
+	req.Header.Set("Authorization", "apiToken "+a.cfg.APIToken)
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := a.client.Do(req)
```
gulducat: I'm a little surprised there's no go lib for Instana API. There is a "go-sensor" lib to send application traces and such to Instana, but that's another task entirely.

At any rate, hopefully our needs for API interaction are simple enough, but I'd like to isolate the API-calling logic in its own file (doesn't need to be a separate package, necessarily, but that might be nice), so this "QueryMultiple" method would build the request then defer to the API client, keeping HTTP activity distinct from the business logic we care about in autoscaler methods.

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -139,7 +150,128 @@ func (a *APMPlugin) Query(q string, r sdk.TimeRange) (sdk.TimestampedMetrics, er
...
+	if err != nil {
+		return nil, fmt.Errorf("error querying metrics from instana: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusTooManyRequests {
+		return nil, fmt.Errorf("metric queries are ratelimited by instana, resets at %s",
+			resp.Header.Get(rateLimitResetHdr))
```
gulducat: Maybe we could include retries from the get-go? This makes it pretty easy: https://github.com/hashicorp/go-retryablehttp

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -139,7 +150,128 @@ func (a *APMPlugin) Query(q string, r sdk.TimeRange) (sdk.TimestampedMetrics, er
...
+			}
+		}
+	}
+
+	return results
+}
+
+// instanaQueryRequest is the JSON body POSTed to the Instana infrastructure
+// metrics endpoint: POST /api/infrastructure-monitoring/metrics
+type instanaQueryRequest struct {
```
gulducat: This being unexported means we can easily change it later, but this name is too broad for its purpose. Call it what it is, IMO: `instanaMetricsRequest` or something like that.

I now see below the response is sensibly named `instanaMetricsResponse`, so yeah this should match that.

In general when making a Go API client, since we have to (or at least it's a very good idea to) make structs for each request and response per endpoint, the client will have a method associated with the endpoint, like

```go
type instanaAPI struct {
    http *http.Client
    token string
}
func (a *instanaAPI) getInfrastructureMetrics(req *instanaMetricsRequest)  (*instanaMetricsResponse, error){
    // do stuff that all endopints do (maybe a little helper or middleware helps with this)
    // do the stuff unique to this endpoint
} 
```

I think you get the idea. Basically I think we need a good API client design separate from the autoscaler stuff, which then should be pretty standard compared to other plugins, I think?

plugins/builtin/apm/instana/plugin/plugin_test.go
```diff
@@ -0,0 +1,449 @@
...
+			}
+		})
+	}
+}
+
+// TestAPMPlugin_QueryMultiple verifies that QueryMultiple sends the correct
+// HTTP request (method, path, auth header, content-type, injected timeFrame)
+// and correctly handles success responses, empty results, HTTP 429, and
+// generic HTTP errors.
+func TestAPMPlugin_QueryMultiple(t *testing.T) {
```
gulducat: This test will be very different if we extract the API client. We would instead test the distinct logic separately, and it'll be really obvious if we're not testing something important to the autoscaler (skimming this now, I'm not sure what in here even is autoscaler related vs API logic).

gulducat (commented):
I'd like to refactor this to separate these concerns:
- Autoscaler logic
- API calls

Let me know when you're ready for me to take another look!

plugins/builtin/apm/instana/plugin/plugin.go
hashi-divyansh: I think retries from the `get-go`  will not be helpful here, our **autoscaler's ticker** logic is itself a kind of retries. 

Even if retries return the response (after rate limit wait completed) that response is not useful for us. Autoscaler should always takes decision based on the new/latest **ticker** result.

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// getInfrastructureMetrics sends a POST request to the Instana infrastructure metrics endpoint.
+func (c *instanaClient) getInfrastructureMetrics(ctx context.Context, request instanaMetricsRequest) (*instanaMetricsResponse, error) {
+	body, err := json.Marshal(request)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal instana query request: %w", err)
+	}
+
+	metricsURL := *c.baseURL // take a copy
```
gulducat: This pointer dereference means that this field is a hard requirement (not-nil) or else the program will panic. If that's the case, then the constructor (`newInstanaClient`) really ought to take the value as an argument.

Also, why "take a copy"? We're not mutating the URL, right? So we can just continue using it as a pointer. I would say we don't even need a pointer, that we can pass this simple struct by value rather than reference, but that's what `url.Parse()` provides, so we may as well just keep it that way.

The `apiToken` being a plain string isn't a panic risk, but effectively it is required for anyone using the service, so it should be an argument too.

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,70 @@
...
+	metricsURL.Path = strings.TrimSuffix(metricsURL.Path, "/") + metricsPath
```
gulducat: There's a [`path`](https://pkg.go.dev/path) package that is great for stuff like this. It'll strip superfluous `/`es and such.

```go
	metricsURL.Path = path.Join(metricsURL.Path, metricsPath)
```

Aaaand I see now that you _are_ mutating the `&url.URL`, so that's why you wanted to take a copy.

Instead I think you should move the `url.Parse` into the constructor here, as it is an API client concern more than an Autoscaler thing. Then your constructor can return `error` if the provided `baseURL string` is invalid, and you can bank on that for the rest of the API code.

Then here, rather than mutate the `url.URL`, just use it to build the `metricsURL`. There's even a `url.URL` method just for this: [`JoinPath()`](https://pkg.go.dev/net/url#URL.JoinPath)

```go
	metricsURL := c.baseURL.JoinPath(metricsPath).String()
```

and plop that in `NewRequestWithContext()`

plugins/builtin/apm/instana/plugin/client_test.go
```diff
@@ -0,0 +1,92 @@
...
+				if tc.fixture != "" {
+					data, err := os.ReadFile(filepath.Join("test-fixtures", tc.fixture))
+					must.NoError(t, err)
+					_, _ = w.Write(data)
+				}
+			}))
+			defer srv.Close()
+
+			u, _ := url.Parse(srv.URL)
+			c := &instanaClient{baseURL: u, apiToken: "test-token", http: &http.Client{}}
```
gulducat: Needing to set these values on the struct, rather than use the constructor, reveals a flaw in the constructor.

Conversely, only using the constructor can be a blind spot, too, if you want to test that your "bare struct" can operate without panics. It might not *work*, but ideally a bare struct doesn't blow up the whole program.

I think I see why you did it this way, so I'll comment next on how to allow the constructor to accept the config values.

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -47,71 +55,48 @@ var (
...
-	client *http.Client
 	logger hclog.Logger
+	client *instanaClient
 }
 
-// NewInstanaPlugin returns a new Instana APM plugin instance.
 func NewInstanaPlugin(log hclog.Logger) apm.APM {
 	return &APMPlugin{
 		logger: log,
+		client: newInstanaClient(),
```
gulducat: This here establishes a broken default. Since the client depends on config, set it *in the config func*. I.e. `SetConfig()` should set `a.client, err = newInstanaClient(baseURL, token)`, rather than setting `a.client.baseURL` and `a.client.token` separately afterwards. Then methods on `APMPlugin` that need the client can just check whether they have one at all, and expect that if they do, then it is safe to use.

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -29,6 +31,12 @@ const (
 
 	// envKeyAPIToken is the environment variable fallback for the API token.
 	envKeyAPIToken = "INSTANA_API_TOKEN"
+
+	// metricsPath is the Instana REST API path for infrastructure metrics.
+	metricsPath = "/api/infrastructure-monitoring/metrics"
+
+	// rateLimitResetHdr is the response header Instana sets when rate-limiting.
+	rateLimitResetHdr = "X-RateLimit-Reset"
```
gulducat: These are specific to the API client, so I'd put em in `client.go`

gulducat (commented):
More comments about struct design, pointer safety, and separation of concerns before I dig into the actual autoscaler logic.

Quick note on the `query` format, if we're providing examples in documentation. I like the transparency of providing the query literally to the Instana API, but I personally really dislike all the `\"` escaping needed for plain JSON strings. It's a recipe for syntax errors, but HCL v2 has a [`jsonencode()`](https://developer.hashicorp.com/nomad/docs/reference/hcl2/functions/encoding/jsonencode) func that makes it way nicer to work with:

```hcl
query = jsonencode({
  plugin      = "host"
  snapshotIds = ["id1", "id2"]
  metrics     = ["memory.used"]
})
```

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,89 @@
...
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+const (
+	// metricsPath is the Instana REST API path for infrastructure metrics.
+	metricsPath = "/api/infrastructure-monitoring/metrics"
```
gulducat, line 18: Nice to link to API docs in comments when you can. I think this is the one?

https://developer.ibm.com/apis/catalog/instana--instana-rest-api/api/API--instana--instana-rest-api-documentation#getInfrastructureMetrics

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// rateLimitResetHdr is the response header Instana sets when rate-limiting.
+	rateLimitResetHdr = "X-RateLimit-Reset"
+)
+
+// instanaClient handles HTTP communication with the Instana REST API.
+type instanaClient struct {
+	metricsURL string
```
gulducat, line 26: This isn't exactly locking us in to a single API endpoint, but some refactoring would be required if we want to include, say, ["application metrics"](https://developer.ibm.com/apis/catalog/instana--instana-rest-api/api/API--instana--instana-rest-api-documentation#getApplicationMetrics) (whatever that means exactly, it just seems plausible that someone might ask for it later).

I'm not saying you should do it now, necessarily, because overgeneralizing is a common pitfall and one I'm quite vulnerable to myself. But I find it's a useful exercise to ask myself, "What about this implementation is specific to the problem right in front of me, and what aspects may extend to other instances or similar cases?"

Here, consider what it would look like to add one (or more) additional API endpoints.

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,89 @@
...
+	apiToken   string
+	http       *http.Client
+}
+
+func newInstanaClient(endpoint, apiToken string) (*instanaClient, error) {
+	parsedURL, err := url.Parse(endpoint)
```
gulducat, line 32: Excellent validating the specific values that this thing needs to succeed 👍

plugins/builtin/apm/instana/plugin/client.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, fmt.Errorf("instana query failed with status %d: %s",
+			resp.StatusCode, strings.TrimSpace(string(respBody)))
+	}
+
+	var metricsResp instanaMetricsResponse
+	if err := json.NewDecoder(resp.Body).Decode(&metricsResp); err != nil {
+		return nil, fmt.Errorf("failed to decode instana response: %w", err)
+	}
+
+	return &metricsResp, nil
```
gulducat, line 88: I think this is the only thing in this method that is specific to this one API endpoint, the rest could be abstracted into a `doRequest()` method, or similar, to be reused later.

Again, I'm not saying you must do it now! This little client is perfectly serviceable for our current needs and does its job of only doing its job and nothing else. 👍

plugins/builtin/apm/instana/plugin/client_test.go
```diff
@@ -0,0 +1,138 @@
...
+			statusCode:     http.StatusTooManyRequests,
+			rateLimitReset: "1717000000000",
+			expectErr:      "metric queries are rate limited by instana, resets at 1717000000000",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				// verify request shape
```
gulducat, line 105: I like the live http server, and I think you're testing the main paths we care about. Nice!

I am taking it on faith that the fixtures are representative of real API responses. 🙏

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -71,47 +63,24 @@ func NewInstanaPlugin(log hclog.Logger) apm.APM {
 // SetConfig parses and validates the plugin configuration. All required fields
 // are checked before any state is mutated.
 func (a *APMPlugin) SetConfig(config map[string]string) error {
-	cfg, err := parseConfig(config)
```
gulducat, line 74: This is very nice and tight now, great work.

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -139,7 +108,65 @@ func (a *APMPlugin) Query(q string, r sdk.TimeRange) (sdk.TimestampedMetrics, er
...
+		return nil, fmt.Errorf("failed to unmarshal instana query: %w", err)
+	}
+
+	queryReq.TimeFrame = instanaTimeFrame{
+		WindowSize: r.To.Sub(r.From).Milliseconds(),
+		To:         r.To.UnixMilli(),
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
```
gulducat, line 133: With a 10s timeout, it's not a huuuge deal, but is there a context other than `Background()` that we could use? I don't see one on `APMPlugin`, nor in the `NewInstanaPlugin` constructor... I forget if autoscaler has like a "shutdown context" associated with the overall agent.

plugins/builtin/apm/instana/plugin/plugin.go
```diff
@@ -139,7 +108,65 @@ func (a *APMPlugin) Query(q string, r sdk.TimeRange) (sdk.TimestampedMetrics, er
...
+
+	metricsResp, err := a.client.getInfrastructureMetrics(ctx, queryReq)
+	if err != nil {
```
gulducat, line 136: This QueryMultiple method feels so much cleaner with the API logic isolated in here, very nice.

plugins/builtin/apm/instana/plugin/plugin_test.go
```diff
@@ -0,0 +1,283 @@
...
+			expectErr: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.envToken != "" {
+				// t.Setenv sets the value and automatically restores the
+				// original state once the subtest finishes.
+				t.Setenv(envKeyAPIToken, tc.envToken)
```
gulducat, line 104: Love a good `t.Setenv`. You probably noticed, but it does prevent using `t.Parallel()`, so sometimes I'll split env var test(s) into separate cases so other ones can run in parallel. I don't think it matters much in this case, but when tests are slow it can add up.

gulducat (approved):
LGTM Nice work!

I have a few more comments on API client design, but no need to act on them now unless you want to.

I'm comfortable merging this, but I would like to better understand this part of your description:

> Use `snapshotIds` to target a single host (required for `Query()`). Use the `query` filter for fleet-wide queries with `QueryMultiple()`.

I see your implementation here is not out of the norm for APM plugins (having Query call QueryMultiple and expect exactly one response), but QueryMultiple [on the interface](https://github.com/hashicorp/nomad-autoscaler/blob/main/plugins/apm/apm.go#L28-L30) claims that it's only meant for DAS (actually, only *used by*), and I'm left a little confused about its intent.

hashi-divyansh:
@gulducat My view on few of your comments: 

>`Regarding Query() is calling QueryMultiple()`

The reason `Query()` calls `QueryMultiple()` is that the APM plugin API can return zero, one, or many metrics streams for a single request. In a few of the previous implementations, the API client layer was not separated. To avoid code duplication (building the API request, parsing, marshaling, unmarshaling, etc.), this pattern was followed (at least, I think so! 🙂).

However, we can separate the common logic into a helper method that both `Query()` and `QueryMultiple()` consume. That way, `Query()` doesn't have to call `QueryMultiple()` anymore.


> `Pasing some existing autoscaler context to the plugin instead of context.Background()`

We're implementing the APM plugin interface, whose methods don't accept any context as a method argument. Currently, I'm not able to think of how to do it, but I'll definitely think about it.


I'll raise a separate PR to accommodate your remaining comments.

gulducat:
> However, we can separate the common logic into a helper method that both Query() and QueryMultiple() consume.

I don't think this is necessary, but I think the guidance in the code (doc comments) could be made clearer for folks implementing the interface.

PR #1320, opened 2026-06-23
enable jsonencode in file policy decoding

hashi-divyansh:
**Description:**

Added `jsonencode(...)` support for **file-based autoscaler policies** so users can write structured HCL objects (for example in `check.query`) instead of manually escaping JSON strings.

**Example usage:**
-  The Instana APM plugin expects `check.query` as a JSON string.   This change improves file-policy authoring by allowing `jsonencode({...})`, so users can write readable HCL objects and let autoscaler produce valid JSON, instead of manually maintaining escaped JSON strings/heredocs.

**Note:**
- `jsonencode(...)` already works in Nomad job-sourced policy flows because Nomad evaluates job HCL before autoscaler reads the policy. This PR brings equivalent ergonomics to autoscaler **file-sourced** policies, where that evaluation previously did not happen.


tehut (approved):
looks good!

Issue #839, opened 2024-01-26
Autoscaler forever waiting for gce-mig "target is not ready" when cpu quota is exceeded


sam-sla:
Hello,
So this is an edge case we got stuck in very recently. During some increased testing we apparently reached our default vcpu quota in the gcp project, but because we still didn't have alerts configured it went unnoticed for a while, until we realized that the nomad autoscaler wasn't scaling down nodes from the cluster.

The scenario:

1. gce-mig is trying to scale up but vcpu quota in the project is reached, the instance group has errors of the type `QUOTA_EXCEEDED 	dev-nomad-client-zpgq 	europe-west1-b 	Creating 	Jan 25, 2024, 3:38:35 PM UTC+01:00 	Instance 'dev-nomad-client-zpgq' creation failed: Quota 'N2D_CPUS' exceeded. Limit: 500.0 in region europe-west1. 	
`
2. Nomad autoscaler is periodically checking the mig but doesn't go further because the mig is not ready (stuck trying to scale-up) `[TRACE] policy_manager.policy_handler: target is not ready: policy_id=c00c0934-a44e-c3eb-5200-e71e44255633`

In our case we requested a vcpu increase and that solved the problem, the mig finished the scaling up event that it had started many hours ago, then the nomad autoscaler saw the mig ready and started to function properly again (in our case to scale down as the load had decreased a lot)

I would like to know if this could have been handled better, maybe the autoscaler could check the mig errors, or force a scaling event (not sure if that's possible)

lgfa29:
Hi @sam-sla 👋 

Do you know what would happen if you tried to scale a MIG in that state? I guess maybe a scale down would work and probably desirable if possible 🤔 

But I'm also not sure how to detect this state. Currently we check this [`isStable`](https://cloud.google.com/compute/docs/reference/rest/v1/instanceGroupManagers/get#body.InstanceGroupManager.FIELDS.inlinedField_22) flag to determine if the target is ready for scaling:
https://github.com/hashicorp/nomad-autoscaler/blob/a4e70bcdba4995e4edc24628f6a9806bc33b4270/plugins/builtin/target/gce-mig/plugin/mig.go#L75

Do you know if there's a field we can check for an error like this?

sam-sla:
Hi @lgfa29,

[...]

PR #1335, opened 2026-07-22
feature: best-effort, non-retrying scale-out option for GCP MIGs

jrpilat:
When we request an increase in GCP MIG instance count, but not all of the requested capacity is available, the MIG will keep retrying until the requested capacity is available.

During this period, GCP marks the MIG as “unstable” pending completion of the scale-out, and the nomad-autoscaler, seeing this “unstable” status, does not attempt to re-scale the MIG, until the MIG is once again “stable”.

This bites us because, while waiting for this to happen, the demand for the extra compute may have been time-sensitive, and since abated. Or, we may have scaled to, say, 90% of the request, and gotten “stuck” there until we get 100%, even if the 90% capacity was able to deal with the workload (by running for longer).

nomad-autoscaler performs GCP MIG autoscaling by calling the [resize](https://googleapis.github.io/google-api-python-client/docs/dyn/compute_alpha.instanceGroupManagers.html#resize) function of the GCP API. This will, if capacity is lacking (which it often is), continue retrying the scale-out until 100% of the requested capacity is acquired. This is described in more detail in the parent ticket, https://simplifi.atlassian.net/browse/DSPP-1144 .

The Beta GCP API exposes an ability to disable this retry behavior: The [resizeAdvanced](https://googleapis.github.io/google-api-python-client/docs/dyn/compute_alpha.instanceGroupManagers.html#resizeAdvanced) function can take in it’s body parameter an option called noCreationRetries, described thusly:

```
"noCreationRetries": True or False, # If this flag is true, the managed instance group attempts to create all
      # instances initiated by this resize request only once. If there is an error
      # during creation, the managed instance group does not retry create this
      # instance, and we will decrease the targetSize of the request
      # instead. If the flag is false, the group attempts to recreate each instance
      # continuously until it succeeds.
```

If we can modify nomad-autoscaler to (optionally) use this functionality, then we can effectively offload the retry logic from GCP to Nomad, where smarter decisions can be made: The next normal cluster autoscaling evaluation by nomad-autoscaler will simply try again, with the latest metric numbers as inputs. Additionally, we can avoid going into cooldown when we received fewer than the requested number of resources, allowing for more frequent policy evaluation while we’re not getting enough resources, and thus allowing for more responsiveness to decreases in demand in cases of “we waited so long that we don’t need all this capacity anymore”.

It seems this would also resolve the issue described [here](https://github.com/hashicorp/nomad-autoscaler/issues/839)

This also includes an additional debug log entry from https://github.com/hashicorp/nomad-autoscaler/pull/1235


jrpilat:
Will rebase and squash down to a single commit here

PR #1336, opened 2026-08-12
fix(policy): recreate missing handlers for unchanged policies

### LINKED ISSUE #1334: Policy handlers are never re-created for unchanged policies after the manager recovers from an unrecoverable source error (all pre-existing policies silently stop being evaluated)
### Versions

- nomad-autoscaler v0.5.0 (official Docker image `hashicorp/nomad-autoscaler:0.5.0`)
- The affected code is unchanged on current `main`, so nightly is affected too.
- The mechanism was introduced by the policy handler refactor in #1102 (shipped in 0.4.7), so 0.4.7+ should all be affected.
- Nomad policy source, agent configured with `namespace = "*"` (the namespace config is incidental — the code path is the same for any namespace setting).

### Bug description

When the policy manager hits an "unrecoverable" source error (e.g. the blocking policy-list query fails with `EOF` / `connection refused` during a Nomad server restart or leader election), it stops **all** policy handlers and restarts its monitoring loop. However, the restarted loop reuses the same `Source` instance, whose internal modify-index state was never reset. As a result, every policy that existed before the error is reported as "not updated" and **never gets a new handler**. Those policies silently stop being evaluated until the agent process is restarted.

Only policies created or modified *after* the incident get handlers again, which makes the failure easy to miss: the agent keeps running, logs look healthy at INFO level, and some policies still scale.

### Observed behavior (production timeline)

1. Autoscaler running fine, evaluating 113 policies.
2. Nomad server restart / leader election occurs. All handlers stop:

   ```
   [ERROR] policy_manager.policy_handler: stopping policy handler due to context done: policy_id=...
   ```

   (one line per policy, for all 113)
3. From that point on, only policies created or modified afterwards produce:

   ```
   [DEBUG] policy_manager: creating new policy handler: policy_id=...
   ```

   The 93 pre-existing, unchanged policies are never re-monitored. No further errors are logged for them; the agent otherwise looks healthy (DEBUG logs, other evaluations proceeding).
4. Affected jobs stayed frozen at their last count for ~3 days. In our case a batch worker group stayed at count 1 while its queue grew past 1000 tasks.
5. Restarting the autoscaler task fully recovered: all policies were re-loaded and evaluated again.

### Root cause analysis (v0.5.0 source)

1. `policy/manager.go` — `monitorPolicies()` receives the source error on `policyIDsErrCh`; `isUnrecoverableError()` (line 407) matches `EOF` / `connection refused` and `monitorPolicies` returns the error (line 171-173).
2. Its `defer m.stopHandlers()` (line 158) cancels every handler context — producing the `stopping policy handler due to context done` error log from `policy/handler.go:293`.
3. `Manager.Run()` (line 116-154) then resets `m.handlers` to an empty map (line 149), sleeps 10s, and restarts `MonitorIDs` — **on the same long-lived `Source` object**.
4. `policy/nomad/source.go` — the `Source` struct's `monitoredPolicies map[policy.PolicyID]modifyIndex` and `latestIndex` fields are **never reset** across these restarts. On the next list result, every pre-existing policy hits line 257-258:

   ```go
   if oldPolicyModifyIndex, ok := s.monitoredPolicies[newPolicy.ID]; ok {
       policyUpdates[newPolicy.ID] = oldPolicyModifyIndex < newPolicy.ModifyIndex
   }
   ```

   which evaluates to `false` for any policy whose `ModifyIndex` didn't change.
5. `policy/manager.go` — `processMessageAndUpdateHandlers()` skips those policies at line 210-212:

   ```go
   if !updated {
       continue
   }
   ```

   **before** checking whether a handler actually exists for them. Since the handler map was wiped in step 3, no handler is ever created again for unchanged policies.

The manager's state (handlers map) and the source's state (`monitoredPolicies`) are out of sync after recovery: the source believes the policies are known and unchanged; the manager has no handlers for them.

Note: #1206 / #1207 (fixed in 0.4.9) addressed a sibling bug in the SIGHUP-reload path ("monitor doesn't start again after failing once"), but that fix does not cover this unrecoverable-error restart path.

### Reproduction

1. Run nomad-autoscaler 0.5.0 with the Nomad policy source against a cluster that has existing scaling policies.
2. Restart the Nomad server the autoscaler is connected to (or force a leader election) so the blocking list query fails with `EOF` or `connection refused`.
3. Observe `stopping policy handler due to context done` for every policy.
4. Wait for the manager to restart its loop (10s) and for the next policy-list change to be delivered. Unchanged policies never log `creating new policy handler` and are never evaluated again.
5. Modify one policy (bumping its `ModifyIndex`) — only that policy gets a new handler, confirming the `updated=false` skip.

### Suggested fixes

Either (or both) would resolve it:

- Reset the source's tracking state when `Manager.Run` re-enters after an unrecoverable error — e.g. extend `ReloadIDsMonitor()` or add a `Reset()` to the `Source` interface that clears `monitoredPolicies` and `latestIndex`, and call it alongside the `m.handlers` reset in `Run()`.
- In `processMessageAndUpdateHandlers()`, check whether a handler exists for the policy **before** the `if !updated { continue }` skip, and create one if it's missing even when `updated == false`.

### Workaround

Restarting the autoscaler agent fully recovers. Monitoring the `scaling.policy.total_num` telemetry gauge (or alerting on the `stopping policy handler due to context done` log line) helps detect the condition.

vishnuedavarath:
Recreate missing policy handlers from the current source snapshot even when a policy is marked unchanged. This fixes the restart path where manager state is cleared but source state is retained, leaving unchanged policies without handlers until they are modified.

This PR grew beyond the original handler-recreation fix and now also narrows recovery for retryable operational failures.

Additional changes in this update:

- retryable `EOF` / `connection refused` errors no longer drive a manager-wide restart path
- Nomad source now handles retryable watch failures locally by resetting its own watch state and rebuilding from a fresh snapshot
- handler runtime errors now stay local to the handler instead of feeding the manager control path
- transient handler creation failures are retried per policy, instead of requiring a manager restart or waiting for a later policy update
- duplicate pending retries for the same `(source, policyID)` are deduplicated

Fixes: #1334
Fixes: https://hashicorp.atlassian.net/browse/NMD-1654

### DIFF
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -87,7 +87,7 @@ type Handler struct {
 	limiter limiter
 
 	// errCh is used to surface errors while executing the policy
-	errChn chan<- error
+	errChn chan error
 	// ch is used to listen for policy updates.
 	updatesCh <-chan *sdk.ScalingPolicy
 
@@ -112,7 +112,7 @@ type Handler struct {
 
 type HandlerConfig struct {
 	UpdatesChan      chan *sdk.ScalingPolicy
-	ErrChan          chan<- error
+	ErrChan          chan error
 	Policy           *sdk.ScalingPolicy
 	Log              hclog.Logger
 	TargetController targetpkg.Controller
@@ -292,6 +292,11 @@ func (h *Handler) Run(ctx context.Context) {
 		case <-ctx.Done():
 			h.log.Error("stopping policy handler due to context done")
 			return
+
+		case err := <-h.errChn:
+			if err != nil {
+				h.log.Error("encountered a runtime error while executing the policy", "error", err)
+			}
 
 		case updatedPolicy := <-h.updatesCh:
 			// The policy can be nil if the channel is closed meaning the
--- a/policy/handler.go
+++ b/policy/handler.go
@@ -86,8 +86,6 @@ type Handler struct {
 
 	limiter limiter
 
-	// errCh is used to surface errors while executing the policy
-	errChn chan error
 	// ch is used to listen for policy updates.
 	updatesCh <-chan *sdk.ScalingPolicy
 
@@ -112,7 +110,6 @@ type Handler struct {
 
 type HandlerConfig struct {
 	UpdatesChan      chan *sdk.ScalingPolicy
-	ErrChan          chan error
 	Policy           *sdk.ScalingPolicy
 	Log              hclog.Logger
 	TargetController targetpkg.Controller
@@ -134,7 +131,6 @@ func NewPolicyHandler(config HandlerConfig) (*Handler, error) {
 		targetController:    config.TargetController,
 		updatesCh:           config.UpdatesChan,
 		policy:              config.Policy,
-		errChn:              config.ErrChan,
 		limiter:             config.Limiter,
 		stateLock:           sync.RWMutex{},
 		state:               StateIdle,
@@ -293,11 +289,6 @@ func (h *Handler) Run(ctx context.Context) {
 			h.log.Error("stopping policy handler due to context done")
 			return
 
-		case err := <-h.errChn:
-			if err != nil {
-				h.log.Error("encountered a runtime error while executing the policy", "error", err)
-			}
-
 		case updatedPolicy := <-h.updatesCh:
 			// The policy can be nil if the channel is closed meaning the
 			// handler is being removed, let the context cancelation take
@@ -307,7 +298,9 @@ func (h *Handler) Run(ctx context.Context) {
 			}
 
 			h.applyMutators(updatedPolicy)
-			h.updateHandler(updatedPolicy)
+			if err := h.updateHandler(updatedPolicy); err != nil {
+				h.log.Error("encountered a runtime error while executing the policy", "error", err)
+			}
 
 		case <-h.ticker.C:
 			if h.compiledPolicySchedule != nil && !h.compiledPolicySchedule.activeAt(nowFunc()) {
@@ -317,14 +310,15 @@ func (h *Handler) Run(ctx context.Context) {
 
 			currentStatus, err := h.runTargetStatus()
 			if err != nil {
-				h.errChn <- fmt.Errorf("handler: failed to get target status for target: %s, %w",
-					h.policy.Target.Name, err)
+				h.log.Error("encountered a runtime error while executing the policy",
+					"error", fmt.Errorf("handler: failed to get target status for target: %s, %w",
+						h.policy.Target.Name, err))
 				continue
 			}
 
 			// A nil status indicates the target doesn't exist, log and return.
 			if currentStatus == nil {
-				h.errChn <- errTargetNotFound
+				h.log.Error("encountered a runtime error while executing the policy", "error", errTargetNotFound)
 				continue
 			}
 
@@ -342,8 +336,9 @@ func (h *Handler) Run(ctx context.Context) {
 					h.log.Debug("stopping policy handler, context done during evaluation")
 					return
 				}
-				h.errChn <- fmt.Errorf("handler: unable to execute policy ID: %s, %w",
-					h.policy.ID, err)
+				h.log.Error("encountered a runtime error while executing the policy",
+					"error", fmt.Errorf("handler: unable to execute policy ID: %s, %w",
+						h.policy.ID, err))
 				continue
 			}
 
@@ -448,7 +443,7 @@ func (h *Handler) waitAndScale(ctx context.Context) error {
 
 // updateHandler updates the handler's internal state based on the changes in
 // the policy being monitored.
-func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
+func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) error {
 	h.log.Trace("updating handler", "policy_id", updatedPolicy.ID)
 
 	h.policyLock.Lock()
@@ -456,8 +451,7 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
 	err := h.applyPolicyState(updatedPolicy)
 	h.policyLock.Unlock()
 	if err != nil {
-		h.errChn <- fmt.Errorf("unable to build and apply updated policy state for policy ID %s: %w", updatedPolicy.ID, err)
-		return
+		return fmt.Errorf("unable to build and apply updated policy state for policy ID %s: %w", updatedPolicy.ID, err)
 	}
 
 	if intervalChanged {
@@ -472,6 +466,7 @@ func (h *Handler) updateHandler(updatedPolicy *sdk.ScalingPolicy) {
 	}
 
 	h.log.Debug("check handlers updated", "count", len(h.checkRunners))
+	return nil
 }
 
 // applyMutators applies the mutators registered with the handler in order and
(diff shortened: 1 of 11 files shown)


policy/nomad/source.go
```diff
@@ -175,6 +175,12 @@ func (s *Source) ReloadIDsMonitor() {
 	s.reloadCh <- struct{}{}
 }
 
+// Reset satisfies the policy.Source interface.
+func (s *Source) Reset() {
+	s.monitoredPolicies = map[policy.PolicyID]modifyIndex{}
+	s.latestIndex = 1
```
gulducat: I thought it odd that there was no mutex protecting these, and indeed if you add `testSource.Reset()` in `policy/nomad/source_test.go` on line 569 (just add it to each subtest run in `TestMonitoringIDs`), the race detector gets tripped when running `go test -race`. Before only the `Source.MonitorIDs` goroutine touched these fields, but now they can be affected concurrently.

In other words, if there was any test code exercising this, CI tests should catch the data race.

policy/manager.go
```diff
@@ -143,6 +143,9 @@ func (m *Manager) Run(ctx context.Context, evalCh chan<- *sdk.ScalingEvaluation)
 		// If we reach this point it means an unrecoverable error happened.
 		// We should reset any internal state and re-run the policy manager.
 		m.log.Debug("re-starting policy manager")
+		for _, s := range m.policySources {
+			s.Reset()
```
gulducat: Ideally we could target only problematic sources to reset, but a) at this layer, we have no knowledge of what source the error came from, just that _a_ source produced an (unrecoverable) error, and b) there's only really 2 sources anyway, right? `file` and `nomad`

Does that assessment match your understanding?

I kind of wish the `nomad` source would instead catch its own unrecoverable errors and handle that internally, since it doesn't really make any sense for a `file` source. Currently this pretends that `"connection refused"` and `"EOF"` are general, cross-source errors, but that's not the case at all.

What do you think? Can we add unrecoverable-error handling somewhere in `policy/nomad/source.Source.MonitorIDs`, and reload policies there, rather than bubbling it all the way up to here and resetting every (both) policy source?

gulducat (commented):
I can see how this could resolve the issue, but it seems like doubling down on a boundary violation... Perhaps the `nomad` source should itself be responsible for "unrecoverable" errors in its domain, rather than the overall/general policy manager making that call.

What do you think?

policy/manager.go
vishnuedavarath: Yes, that matches my understanding. We can’t target reset to just the errored source here because the source isn’t known at this level. Also, the same error channel is currently used for both source-monitoring errors and handler runtime errors, so unrelated transient failures can end up on the same path.

Because of that, any error that reaches the manager and contains `EOF` or `connection refused` is currently treated as unrecoverable, even when those errors are often specific to the component that produced them. For example, a transient plugin/handler-side error containing one of those strings can currently trigger the manager restart path, even though that doesn’t mean the source or manager state is stale(The Datadog `unexpected EOF` incident from the Jira link in the description).

So I agree it would be cleaner for the Nomad source to handle its own failures inside `policy/nomad/source.Source.MonitorIDs`, similarly for handler/runtime transient errors to stay local to the handler. If we move recovery to those owning layers, we may be able to remove or significantly narrow the manager-wide reset path.

policy/source.go
```diff
@@ -48,6 +48,10 @@ type Source interface {
 	// so that config items can be reloaded gracefully without restarting the
 	// agent.
 	ReloadIDsMonitor()
+
+	// Reset clears any internal watch state so the next MonitorIDs session can
+	// rebuild the current policy snapshot from scratch.
+	Reset()
```
gulducat: We should no longer need this new method on the interface, nor any of the other implementations, if the `nomad` source handles its own unique condition internally.

policy/manager.go
```diff
@@ -394,19 +438,11 @@ func (m *Manager) periodicMetricsReporter(ctx context.Context, interval time.Dur
...
-// upon reconnecting, the state of the Nomad cluster could be different from
-// what's stored in the policy manager (e.g., a different cluster becomes
-// available in the same address).
-func isUnrecoverableError(err error) bool {
-	unrecoverableErrors := []string{"connection refused", "EOF"}
-	for _, e := range unrecoverableErrors {
+// IsRetryableError identifies transient connectivity failures that callers can
+// handle locally without restarting the whole manager.
+func IsRetryableError(err error) bool {
+	retryableErrors := []string{"connection refused", "EOF"}
```
gulducat, line 415: Let's move this inside the nomad source, where it actually applies. The file source would not ever return this kind of error.

Edit: I see, this is the inverse, because ostensibly the nomad source would not surface these? Or if it does, now we ignore them? I'm just finding all this to be pretty unintuitive.

policy/manager.go
```diff
@@ -306,6 +316,40 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
 	return nil
 }
 
+func (m *Manager) schedulePolicyRetry(ctx context.Context, source SourceName, policyID PolicyID) {
+	key := retryKey{source: source, policyID: policyID}
```
gulducat, line 291: I'm confused - I thought we'd agreed that the retry logic should be moved into the nomad source. I imagined that if anything, the manager code could be simplified as a result, rather than expanded as it is in this commit.

policy/handler.go
```diff
@@ -293,6 +293,11 @@ func (h *Handler) Run(ctx context.Context) {
 			h.log.Error("stopping policy handler due to context done")
 			return
 
+		case err := <-h.errChn:
+			if err != nil {
+				h.log.Error("encountered a runtime error while executing the policy", "error", err)
+			}
```
gulducat: In general I do like bubbling up errors, but here we're only logging it anyway (and not breaking from this loop or anything), so why not just log it at its source?

The extra channel plumbing seems unhelpful, and I'm also always wary of changing a channel from directional `<-` to not... It makes me wonder who the owner of the channel is meant to be, primarily when it comes to closing it, as a write on a closed channel will cause a panic, for example. Ideally channel vars that _can_ be one-directional, are.

gulducat (changes requested):
I must say the latest commit is perplexing. It doesn't seem to match your reply to my comment, where it seemed that we were in agreement about what changes should be made.

Given the current state of the PR, I almost want to suggest that you close it and start fresh, aiming directly at moving `nomad`-specific error/state handling into the `nomad` source and cleaning up the `manager` code insofar as its scope may be narrowed by the nomad source change.

If you want to salvage this PR, feel free, but at the moment it's looking pretty rough to me, and I worry about its current trajectory.

policy/manager.go
vishnuedavarath: The piece I kept in `manager` is a different retry path: retrying a specific policy when handler creation fails transiently in `GetLatestVersion`, `GetTargetController`, or `NewPolicyHandler`. That retry is per-policy handler lifecycle, not source monitor recovery.

policy/manager.go
```diff
@@ -394,19 +409,11 @@ func (m *Manager) periodicMetricsReporter(ctx context.Context, interval time.Dur
...
-// Example of an unrecoverable error: the Nomad server becomes unreachable;
-// upon reconnecting, the state of the Nomad cluster could be different from
-// what's stored in the policy manager (e.g., a different cluster becomes
-// available in the same address).
-func isUnrecoverableError(err error) bool {
-	unrecoverableErrors := []string{"connection refused", "EOF"}
-	for _, e := range unrecoverableErrors {
+// IsRetryableError identifies transient connectivity failures that callers can
+// handle locally without restarting the whole manager.
+func IsRetryableError(err error) bool {
```
gulducat, line 414: Why make this public if it's only used within the package?

gulducat (approved):
Thanks for taking the pivot on this, looks much cleaner now!

I have just the one nit-pick remaining, but not blocking.

policy/manager.go
vishnuedavarath: Oh! I just missed this after removing the calls from other package. I'll fix this in the upcoming PR.

PR #1337, opened 2026-08-31
fix: wait for all tasks to finish before terminating node after drain

vishnuedavarath:
#### Background
When Nomad Autoscaler performs a cluster scale-in, it drains the selected node and terminates the backing instance once drain is reported complete. When an allocation is already in a terminal client status (e.g. `failed`) before drain is initiated, the drainer does not consider it for migration since it does not need to be migrated. The alloc is skipped, drain completes immediately, and Autoscaler proceeds to terminate the instance, even if a `poststop` lifecycle task from that terminal alloc is still actively running on the node. The result is that the `poststop` task is permanently lost.

#### Fix
After drain completes, before proceeding with instance termination, poll the node's allocations using Nomad blocking queries until all tasks across all allocations have reached a `dead` state. This ensures any remaining work on the node, including `poststop` lifecycle tasks on already-terminal allocations has fully completed before the node is removed.

Jira Link: https://hashicorp.atlassian.net/browse/NMD-1675

### DIFF
--- a/policy/manager.go
+++ b/policy/manager.go
@@ -204,7 +204,7 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
 		if err != nil {
 			m.log.Error("encountered an error getting the latest version for policy",
 				"policyID", policyID, "error", err)
-			if IsRetryableError(err) {
+			if isRetryableError(err) {
 				m.schedulePolicyRetry(ctx, message.Source, policyID)
 			}
 			continue
@@ -253,7 +253,7 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
 		if err != nil {
 			m.log.Error("encountered an error getting the target for the policy handler",
 				"policyID", policyID, "error", err)
-			if IsRetryableError(err) {
+			if isRetryableError(err) {
 				m.schedulePolicyRetry(ctx, message.Source, policyID)
 			}
 			continue
@@ -273,7 +273,7 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
 		if err != nil {
 			m.log.Error("encountered an error starting the policy handler",
 				"policyID", policyID, "error", err)
-			if IsRetryableError(err) {
+			if isRetryableError(err) {
 				m.schedulePolicyRetry(ctx, message.Source, policyID)
 			}
 			continue
@@ -409,9 +409,9 @@ func (m *Manager) periodicMetricsReporter(ctx context.Context, interval time.Dur
 	}
 }
 
-// IsRetryableError identifies transient connectivity failures that callers can
+// isRetryableError identifies transient connectivity failures that callers can
 // handle locally without restarting the whole manager.
-func IsRetryableError(err error) bool {
+func isRetryableError(err error) bool {
 	retryableErrors := []string{"connection refused", "EOF"}
 	for _, e := range retryableErrors {
 		if strings.Contains(err.Error(), e) {
--- a/sdk/helper/scaleutils/node_drain.go
+++ b/sdk/helper/scaleutils/node_drain.go
@@ -172,7 +172,6 @@ func (c *ClusterScaleUtils) waitForAllTasksDead(ctx context.Context, nodeID stri
 		}
 
 		if !blocking.IndexHasChanged(meta.LastIndex, q.WaitIndex) {
-			q.WaitIndex = meta.LastIndex
 			continue
 		}
 
--- a/sdk/helper/scaleutils/node_drain.go
+++ b/sdk/helper/scaleutils/node_drain.go
@@ -1,4 +1,4 @@
-// Copyright IBM Corp. 2020, 2025
+// Copyright IBM Corp. 2020, 2026
 // SPDX-License-Identifier: MPL-2.0
 
 package scaleutils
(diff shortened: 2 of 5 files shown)


policy/manager.go
```diff
@@ -204,7 +204,7 @@ func (m *Manager) processMessageAndUpdateHandlers(ctx context.Context, message I
 		if err != nil {
 			m.log.Error("encountered an error getting the latest version for policy",
 				"policyID", policyID, "error", err)
-			if IsRetryableError(err) {
+			if isRetryableError(err) {
```
vishnuedavarath, line 207: [Recent PR suggestion](https://github.com/hashicorp/nomad-autoscaler/pull/1336#discussion_r3875469929)

Adding this change too to this PR.

sdk/helper/scaleutils/node_drain.go
```diff
@@ -146,9 +147,60 @@ func (c *ClusterScaleUtils) drainNode(ctx context.Context, nodeID string, spec *
...
+
+	for {
+		allocs, meta, err := c.client.Nodes().Allocations(nodeID, q.WithContext(ctx))
+		if err != nil {
+			return fmt.Errorf("failed to list allocations for node: %w", err)
+		}
+
+		if !blocking.IndexHasChanged(meta.LastIndex, q.WaitIndex) {
+			q.WaitIndex = meta.LastIndex
+			continue
```
gulducat, line 175: Just reading this at face value, why do we need to set it if it hasn't changed?

sdk/helper/scaleutils/node_drain.go
```diff
@@ -146,9 +147,60 @@ func (c *ClusterScaleUtils) drainNode(ctx context.Context, nodeID string, spec *
...
+		}
+
+		allDead := true
+	allocCheck:
+		for _, alloc := range allocs {
+			for taskName, taskState := range alloc.TaskStates {
+				if taskState.State != "dead" {
+					c.log.Info("waiting for task to finish before terminating node",
+						"node_id", nodeID,
```
gulducat, line 184: This could potentially be a bit noisy, but I do think it is useful `Info` for people to understand autoscaling delays without full debug logging. We may consider downgrading it to `Debug` if people make a fuss, but personally I like this. 👍

sdk/helper/scaleutils/node_drain_test.go
```diff
@@ -149,12 +167,137 @@ func Test_DrainNode(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			callCount := 0
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				w.Header().Set("Content-Type", "application/json")
+
+				idx := callCount
+				if idx >= len(tc.allocsSeq) {
+					idx = len(tc.allocsSeq) - 1
+				}
+				allocs := tc.allocsSeq[idx]
```
gulducat: Nit-picking on clarity here: I take the intention here to be "after going through all the responses in the sequence, keep returning the last one", which is what the real API would do after stuff stops changing, and avoid an index out of range.

I might like a little comment on there, and if you want to play code golf, personally I think this is a bit clearer:

```suggestion
				// clamp index between 0 and the last slice entry
				idx := max(0, min(callCount, len(tc.allocsSeq)-1))
				allocs := tc.allocsSeq[idx]
```

It also took me a moment to realize that in the "no allocations" case, `[][]*api.Allocation{{}}` is not empty! It has one entry, and _that entry_ is empty. That's true to the API behavior, but just a little confusing while I was trying to determine how we're avoiding an out of range panic here for that case.

sdk/helper/scaleutils/node_drain_test.go
```diff
@@ -149,12 +167,137 @@ func Test_DrainNode(t *testing.T) {
...
+			ctx := context.Background()
+			if tc.ctxTimeout > 0 {
+				var cancel context.CancelFunc
+				ctx, cancel = context.WithTimeout(ctx, tc.ctxTimeout)
+				defer cancel()
+			}
+
+			err := cu.waitForAllTasksDead(ctx, "node-1")
+			if tc.expectedError {
+				must.Error(t, err)
```
gulducat: I generally prefer to assert a specific error, or short of that, the string contents of the error, or short of that, a substring.

gulducat (dismissed):
Looks good! No blockers, but a couple tidbits for your consideration.

sdk/helper/scaleutils/node_drain.go
```diff
@@ -146,9 +147,59 @@ func (c *ClusterScaleUtils) drainNode(ctx context.Context, nodeID string, spec *
...
+// on the node has completed before the caller proceeds with node termination
+func (c *ClusterScaleUtils) waitForAllTasksDead(ctx context.Context, nodeID string) error {
+	q := &api.QueryOptions{
+		WaitIndex: 0,
+	}
+
+	for {
+		allocs, meta, err := c.client.Nodes().Allocations(nodeID, q.WithContext(ctx))
+		if err != nil {
+			return fmt.Errorf("failed to list allocations for node: %w", err)
```
gulducat, line 171: This gave me pause, because there's no retries, so I double-checked how `monitorNodeDrain` handles that, and it looks like it does not! `MonitorDrain` in the `nomad/api` package bails on the first error, and so does our autoscaler code.

It shouldn't surprise me that we don't have more consistent retry logic, but somehow it still does. At any rate, while it adds one more possible place for a transient error to bail break the whole process, this is in keeping with the status quo, so ok then!

gulducat (approved):
LGTM!

PR #1345, opened 2026-09-18
Update datadog API to v2

mzhang8132:
## Description

- Update Datadog API from v1 (deprecated) to v2. 
- Update Datadog plugin to use `QueryTimeseriesData` instead of `QueryMetrics` (recommended by the Datadog developers)
- Update Datadog plugin tests since `QueryTimeseriesData` is a `POST` method rather than a `GET` and queries based on millisecond intervals rather than seconds

Ref: [NMD-1416](https://hashicorp.atlassian.net/browse/NMD-1416)

[NMD-1416]: https://hashicorp.atlassian.net/browse/NMD-1416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


plugins/builtin/apm/datadog/plugin/plugin_test.go
```diff
@@ -183,11 +185,18 @@ func TestAPMPlugin_Query(t *testing.T) {
...
 
-				// Check query params.
-				qp := r.URL.Query()
-				require.Equal(t, "avg:nomad.client.allocated.memory", qp.Get("query"))
-				require.Equal(t, "1600000000", qp.Get("from"))
-				require.Equal(t, "1610000000", qp.Get("to"))
+				// Check POST body fields.
+				body, err := io.ReadAll(r.Body)
+				require.NoError(t, err)
+				var payload map[string]interface{}
```
tgross, line 191: `any` is the modern way of saying `interface{}`

```suggestion
				var payload map[string]any
```

plugins/builtin/apm/datadog/plugin/plugin.go
```diff
@@ -174,41 +187,44 @@ func (a *APMPlugin) QueryMultiple(q string, r sdk.TimeRange) ([]sdk.TimestampedM
...
 
 	var results []sdk.TimestampedMetrics
-	for _, s := range series {
-		pl, ok := s.GetPointlistOk()
-		if !ok {
-			continue
-		}
-
+	for _, values := range valuesMatrix {
 		var result sdk.TimestampedMetrics
```
tgross: If this is likely to be large, can we pre-allocate it to the size of `min(len(values), len(times))`? Likewise with `results` above (there I think you can only pre-allocate it to the size of `valuesMatrix`?)

tgross (commented):
I've left a couple optimization / style comments but otherwise this LGTM

plugins/builtin/apm/datadog/plugin/plugin_test.go
mzhang8132: Yes, `any` was introduced in Go 1.18 as the modern/more readable version of `interface{}`

plugins/builtin/apm/datadog/plugin/plugin.go
mzhang8132: Good catch, let me push that change!

tgross (approved):

